PR #5, opened 2018-01-11
BCN - Danielle & Luis

danielleleb:
@luisacherrera


starter_code/src/movies.js
```diff
@@ -2,12 +2,55 @@
...
+function dramaMoviesRate(movie) {
+    var drama = movie.filter(function(movie) {
+        return movie.genre.includes("Drama");
+    }); 
+
+    if (drama.length === 0) {
+        return;
+    }
+
+    var average = drama.reduce(function (sum, item) {
```
tawebbcn: Don't you feel like you are repeating yourselves? After filtering the array, you can reuse the function that you write before (ratesAverage) to calculate the average.

PR #6, opened 2018-01-11
BCN - James & Jana

jjanastinn:
@jsavory-94


starter_code/src/movies.js
```diff
@@ -1,13 +1,44 @@
...
+        return accumulator + Number(item.rate); 
+    },0).toFixed(2);
+    return total/array.length;
+}
+ratesAverage(movies);
 
 // Get the average of Drama Movies
 
 
+var total = array.reduce(function(accumulator, item) {
```
tawebbcn: 'array' is not defined, it doesn't exist in your scope. However, if you put this code inside the function dramaMoviesRate, then it will work, because array is its parameter. Please, ask me if this is not clear.

starter_code/src/movies.js
```diff
@@ -1,13 +1,44 @@
...
+    if (item.genre === "Drama") {
+        return 1;
+    } else {
+        return -1;
+    }
+});
+
+
+function dramaMoviesRate(array) {
+    var sum = array.reduce(function (accumulator, item) {
```
tawebbcn: Don't you feel like you are repeating yourselves? After filtering the array (see previous comment), you can reuse the function that you write before (ratesAverage) to calculate the average.

PR #7, opened 2018-01-11
BCN - Pedro & Linda

pedrodiogoneto:
@lindaluisa


starter_code/src/movies.js
```diff
@@ -5,8 +5,35 @@
 // Get the average of all rates with 2 decimals 
 
 
+function ratesAverage (array) {
+
+    var reduceFunction = function(rateSum, movie) {
```
tawebbcn: This is fine, but you can use anonymous functions too.

PR #8, opened 2018-01-11
BCN / Oscar  - Rodrigo  en proceso

oscarjorgejorge:
@rodrigodaoud


starter_code/src/movies.js
```diff
@@ -5,9 +5,57 @@
 // Get the average of all rates with 2 decimals 
 
 
+
+
+
+
+
+function ratesAverage (movies) {
+    var sumMo = movies.reduce(function(sum,item){
```
tawebbcn: This is fine, but you can use anonymous functions too.

starter_code/src/movies.js
```diff
@@ -5,9 +5,57 @@
...
+
+
+
+
+function dramaMoviesRate (movies){
+
+    var dramaFi = movies.filter(function(movies) { 
+        return movies.genre.includes('Drama');      
+  });
+    var sumMo = dramaFi.reduce(function(sum,item){
```
tawebbcn: Don't you feel like you are repeating yourselves? After filtering the array, you can reuse the function that you write before (ratesAverage) to calculate the average.

PR #9, opened 2018-01-11
BCN Dani - Pere

pereamoros:
@DTDK

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -5,9 +5,33 @@
 // Get the average of all rates with 2 decimals 
 
 
+function ratesAverage(elements){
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -5,9 +5,39 @@
...
+    }, 0);
+    
+   return Math.round(sum / elements.length*100)/100;
+};
 // Get the average of Drama Movies
 
-
+function dramaMoviesRate(elements){
+    var count = 0;
+    var sum = elements.reduce(function(accumulator,item){
```
tawebbcn: This solution is fine, but if you filter the array before (letting only films with the category Drama), you can reuse the previous function (ratesAverage).

PR #11, opened 2018-01-11
BCN - Matt & Santi

santironhacker:
@matt.megarry

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2,7 +2,29 @@
 // Turn duration of the movies from hours to minutes 
 
 
+
+// function turnHoursToMinutes (array) {
+//     var duration
+//     minutesArray = '';
+//     return minutesArray;
+// };
+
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -2,8 +2,30 @@
...
 // Get the average of all rates with 2 decimals 
 
+//...Callback function
+function ratesTotal (accumulator, currentValue) {
+    accumulator = accumulator + currentValue.rate;
+    return accumulator;
+}
+
+//...Reduce function
+var total = movies.reduce(ratesTotal, 0);
```
tawebbcn: Here you are using the variable 'movies'. This piece of code is not reusable, you can't calculate the average of another array of objects (e.g. 'movie2'). Wrap the code in a function with a parameter for the array. Furhtermore, you can replace ratesTotal with an anonimous function.

PR #12, opened 2018-01-11
BCN - Alla , Josep 

Jsalvat:
@allarom

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2007,75 +2007,25 @@ var movies = [
 
 
 // Get the average of all rates with 2 decimals 
-//  function ratesAverage(acc, item){
-//   return acc + Math.floor(item.rate);
-//   };
 
-
-
- var ratesAverage = movies.reduce(function(acc, item){
-     return acc + Math.floor(item.rate);
-    
-     },0);
-    console.log(ratesAverage);
-
+function ratesAverage(movies){
+  var total = movies.reduce(function(acc, item){
+    return acc + Math.floor(item.rate);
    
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,13 +1,2035 @@
+var movies = [
```
tawebbcn: You don't need to copy the variable in this file. 'movies.js' has access to 'data.js' because the second one is before in order in index.html.

starter_code/src/movies.js
```diff
@@ -1,13 +1,2035 @@
...
+
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
 
 
 // Get the average of all rates with 2 decimals 
 
+function ratesAverage(movies){
+  var total = movies.reduce(function(acc, item){
+    return acc + Math.floor(item.rate);
```
tawebbcn: Why Math.floor? You want to use the decimals too.

starter_code/src/movies.js
```diff
@@ -1,13 +1,2035 @@
...
+   
+    },0);
+  
+var avg = total / movies.length;
```
tawebbcn: Be careful with indentation.

starter_code/src/movies.js
```diff
@@ -16,5 +2038,19 @@
 
 // Order by title and print the first 20 titles
 
+function orderAlphabetically (movies){
+
+  var order = movies.sort(function (acc,element){
```
tawebbcn: Don't use here the name acc, you are not accumuling anything.

PR #13, opened 2018-01-11
BCN Stephie & David

kissy89:
@davidalexteixeira

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -4,8 +4,32 @@
 
 // Get the average of all rates with 2 decimals 
 
+function reduceFunction (sum, item){
+    return sum + item.rate;
+}
 
-// Get the average of Drama Movies
+function ratesAverage(arr) {
+    var averageRate = arr.reduce(reduceFunction , 0) / arr.length;
+    return averageRate;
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -4,8 +4,32 @@
...
+    return averageRate;
+}
+
+//  Get the average of Drama Movies
+
+ function dramaMoviesRate(item){
+     return item.genre.includes("Drama");
+ }
+
+ function reduceDrama (sum, item, index){
```
tawebbcn: Don't you feel like you are repeating yourselves? You can reuse the function that you write before (reduceFunction) to calculate the sum.

starter_code/src/movies.js
```diff
@@ -4,8 +4,32 @@
 
 // Get the average of all rates with 2 decimals 
 
+function reduceFunction (sum, item){
+    return sum + item.rate;
+}
+
+function ratesAverage(arr) {
+    var averageRate = arr.reduce(reduceFunction , 0) / arr.length;
```
tawebbcn: You also can use an anonymous function.

PR #15, opened 2018-01-11
BCN - Vlad & Sheng

theRealSheng:
@vladcio

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,18 +1,78 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
 
+function turnHoursToMinutes(array){
+    var newArray = array.map((a)=>{
+        var b = {};
+        for(key in a){
+            b[key] = a[key];
+            b.duration = Number(toMinutes(a.duration));
+            return b;
+        }
+    });
+
+    return newArray;
+}
+
+function toMinutes(str){
+var regExp = new RegExp(/\d*/)
+    str.split(' ');
+    var hours = str[0].match(regExp, (x) => Number(x) * 60);
+    var min = str[1].match(regExp, (x) => Number(x));
+
+    let total = hours + min;
+
+    return total;
+
+}
 
 // Get the average of all rates with 2 decimals 
 
+function ratesAverage(array) {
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,124 @@
...
+    });
+
+    return newArray;
+}
+
+function toMinutes(str){
+    if (!str.includes('h')) {
+        var point = str.indexOf('min');
+        var min = Number(str.substring(0,point));
+        var hour = 0;
```
tawebbcn: Be careful with variables' names. Your problem here is calling hour and hours the same variable.

starter_code/src/movies.js
```diff
@@ -1,20 +1,124 @@
...
+       return accumulator + Number(item.rate);
+   },0).toFixed(2);
+    return (total / array.length);
+};
 
 // Get the average of Drama Movies
 
+function dramaMoviesRate (array){
+    var count = 0;
+    var total = array.reduce(function(accumulator,item){
```
tawebbcn: Better filter first the array and after that reuse your ratesAverage function.

starter_code/src/movies.js
```diff
@@ -1,20 +1,124 @@
...
+    let result = array.filter((movie) => movie.director == 'Steven Spielberg' && movie.genre.indexOf('Drama') !== -1).length;
+
+    return `Steven Spielberg directed ${Number(result)} drama movies!`;
+
+}
 
 // Order by title and print the first 20 titles
+function orderAlphabetically(array) {
+    let temp = [];
+    newArray = array.sort((a,b)=> b[0] - a[0]);
```
tawebbcn: You can't substract strings.

starter_code/src/movies.js
```diff
@@ -1,20 +1,124 @@
...
+function bestYearAvg(array) {
+    let obj ={};
+
+    for(let i=0; i<array.length; i++){
+        if (obj[array[i].year] == undefined){
+            obj[year] = obj[array[i].year];
+            obj[year][rate] = array[i].rate;
+        } else {
+            obj[year][rate] += array[i].rate;
+            obj[year][rate] = obj[year][rate] / 2;
```
tawebbcn: What are you trying to do here?

PR #29, opened 2018-01-25
[MIA] [Stephen - Jackson]

stephjackson:
I worked on this more this morning. This all gets workable output in repl.it, and I'm not sure why Jasmine is complaining about hours to minutes and duration. The error messages are very hard to parse because of the size of the object.


sandrabosk:
This is very good. I like the fact that you went an extra step and made the returns as nice sentences. The only thing I could add here are comments, just for sake some future event and need for code refactoring so you know what exactly is the each line or block of code doing.
Thank you very much for submitting.

PR #33, opened 2018-01-25
[MIA][Francis-Calizo]

FrancisCalizo:
eveything besides bonus. my brain is fried


sandrabosk:
This is amazing! I love your comments, so detailed! Keep working like this and have no fear in your success! Thank you for submitting.

PR #38, opened 2018-01-26
[MIA] [Nicolas - Levo]

nlevo:


starter_code/src/movies.js
```diff
@@ -12,9 +90,22 @@
 
 
 // How many movies did STEVEN SPIELBERG
-
+function howManyMovies(array){
+    var filteredArray;
+    filteredArray = array.filter(function(movie){
+        return movies.director === "Steven Spielberg";
+    })
```
sandrabosk: You are missing return here:
``` 
return filteredArray;
```

sandrabosk:
Very good job Nikola. Keep it this way! Thank you for submitting.

PR #48, opened 2018-01-29
[MIA] - [Jonathan - Sanchez]

jonathanlsanch:
Not completed


starter_code/src/movies.js
```diff
@@ -1,15 +1,61 @@
...
+function turnHoursToMinutes(array) {
+    var movies = array.map(function (movie) {
+        let durationArray = movie.duration.split(' ');
+        let firstNumber = parseInt(durationArray[0].slice(0, 1)) * 60;
+        let secondNumber = parseInt(durationArray[1].slice(0,2));
+        let output = firstNumber + secondNumber;
+        movie.duration = output;
+        return movie;
+    });
+}
```
sandrabosk: I made couple corrections so we can relate movie.duration with THE movie itself:
```
function turnHoursToMinutes(array) {
    return array.map(function (movie) {
      
        let durationArray = movie.duration.split(" ");
        let firstNumber = parseInt(durationArray[0].slice(0, 1)) * 60;
        let secondNumber = parseInt(durationArray[1].slice(0,2));
        let output = firstNumber + secondNumber;
        // console.log("output", output)
        return Object.assign({}, movie, {duration: output});
    });
}
console.log(turnHoursToMinutes(movies)) 
```

starter_code/src/movies.js
```diff
@@ -1,15 +1,61 @@
...
+// Get the average of all rates with 2 decimals 
 
+function ratesAverage(array) {
+    var ratesTotal = array.reduce(function (previous, current){
+    return previous + parseFloat(current.rate);
+},0)
+    return (ratesTotal / array.length).toFixed(2);
+}
 // Get the average of Drama Movies
+function dramaMoviesRate(array) {
```
sandrabosk: This could be a good place to apply .filter method. We could declare a variable (or better saying, an array) that will hold all the movies that have genre "Drama" and on that array we could apply the ratesAverage() function from above to calculate the average rate of drama movies.

sandrabosk:
Good work so far. If you need any help to finish this assignment, don't hesitate to ask us.

PR #54, opened 2018-03-22
BCN - Ruben & Jordi

jordicalzada:
@rubenaguayo

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,18 +1,74 @@
 /* eslint no-restricted-globals: 'off' */
+
 // Turn duration of the movies from hours to minutes 
 
 
 // Get the average of all rates with 2 decimals 
 
+function ratesAverage(array){
+
+    var ratesSum = array.reduce(function(acc, current){
+
+        return acc + current.rate;
+
+    }, 0);
+
+    return ratesSum/array.length;
+}
+
+ratesAverage(movies);
 
 // Get the average of Drama Movies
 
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,18 +1,74 @@
...
+    return ratesSum/array.length;
+}
+
+ratesAverage(movies);
 
 // Get the average of Drama Movies
 
+function dramaMoviesRate(array){
+
+    var DramaArray = array.filter(function(movie){
```
tawebbcn: Be careful with naming. Use camel case always. ```dramaMoviesArray```

starter_code/src/movies.js
```diff
@@ -1,18 +1,74 @@
...
+        return movie.genre.find(function(element){
+            return element === 'Drama';
+        });
+    });
+
+    var sd = 0;
+
+    var StevenDramaArray = DramaArray.filter(function(movie){
```
tawebbcn: Naming! spielbergDramaMoviesArray

starter_code/src/movies.js
```diff
@@ -1,18 +1,74 @@
...
 
+function howManyMovies(array){
+
+    var DramaArray = array.filter(function(movie){
+        return movie.genre.find(function(element){
+            return element === 'Drama';
+        });
+    });
+
+    var sd = 0;
```
tawebbcn: Not understanding the use of this counter. Maybe you just forgot it here.

PR #55, opened 2018-03-22
BCN Marketa Dauv

dauvbafna:
@Hrabenka7


starter_code/src/movies.js
```diff
@@ -1,20 +1,67 @@
...
+var drama = [];
+movies.filter(function(movie){
+  if (movie.genre == "Drama"){
+      drama.push(movie);
+  }
+})
+function dramaMoviesRate(drama){
+ var dramaAverage = ratesAverage(drama);
+ console.log(" the average drama movies rate is " + dramaAverage );
+}
```
tawebbcn: Good job on reusing the ratesAverage function but what what would happen if I called dramaMoviesRate with a different array than the one from movies? Make sure that you are always working with the data you receive.

starter_code/src/movies.js
```diff
@@ -1,20 +1,67 @@
...
 
 
+// How many movies did STEVEN SPIELBERG
+var dramaSteven= [];
+function howManyMovies(movie){
+  movies.filter(function(movie){
+    if (movie.director == "Steven Spielberg"){
+        dramaSteven.push(movie);
+    }
+  })
```
tawebbcn: You missed the return statement here!

PR #56, opened 2018-03-22
BCN Jorge-Fabian 

fscheugenpflug:
@jorgeterol

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -4,8 +4,29 @@
 
 // Get the average of all rates with 2 decimals 
 
+var totalRate = 0;
+
+function ratesAverage (movies) {
+
+var totalRate =  movies.reduce(function(accumulator, movie){
+      return accumulator + movie.rate;
+  }, 0); return  totalRate / movies.length; 
+}
 
 // Get the average of Drama Movies
+
+var dramaRate = movies.filter(function(movie){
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -4,17 +4,68 @@
...
 
+var dramaRate = movies.filter(function(movie){
+  if (movie.genre == 'Drama'){
+    return movie;
+  }
+});
+
+function dramaMoviesRate (dramaRate) {
+
+var totalRate =  dramaRate.reduce(function(accumulator, movie){
```
tawebbcn: Why not reuse the function above? Also, I would advise to contain the filter inside the function as well. You might receive an array different from the one that you have in the data file so you want to do all the steps inside a function.

starter_code/src/movies.js
```diff
@@ -4,17 +4,68 @@
...
+if (SPIELBERG.length === 0){
+  array = undefined;
+};
+
+function howManyMovies(array){
+  if (array.length > 0){
+    return 'Steven Spielberg directed '+ array.length +' drama movies!';
+  } else if (array == undefined){
+    return 'Steven Spielberg directed 0 drama movies!';
+  }
```
tawebbcn: Try to use the function and include what you need to do inside. If I call the function by itself I will only get two strings back event if I send a different array from the one in movies.

PR #57, opened 2018-03-22
BCN -IOANA n -MIHAI d

cosmihai:
@ioanaOrca

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -3,10 +3,27 @@
 
 
 // Get the average of all rates with 2 decimals 
+function ratesAverage(array) {
+  var sumOfRatings=array.reduce(function(acc, item){
+    var rateNumber=item.rate*1;    
+    return acc+rateNumber;
+  },0)
+
+  return parseFloat((sumOfRatings/array.length).toFixed(2));
+}
+console.log(ratesAverage(movies));
 
 
 // Get the average of Drama Movies
+function dramaMovieRate(array) {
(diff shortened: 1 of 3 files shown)


starter_code/src/movies.js
```diff
@@ -3,10 +3,27 @@
 
 
 // Get the average of all rates with 2 decimals 
+function ratesAverage(array) {
+  var sumOfRatings=array.reduce(function(acc, item){
+    var rateNumber=item.rate*1;    
+    return acc+rateNumber;
+  },0)
 
+  return parseFloat((sumOfRatings/array.length).toFixed(2));
```
tawebbcn: You can use different ways to parse a string to number. Like `parseFloat()` or `Number()` or ` * 1` or `/ 1`.
Either way, be consistent and choose one.

starter_code/src/movies.js
```diff
@@ -3,10 +3,27 @@
...
+}
+console.log(ratesAverage(movies));
 
-// Get the average of Drama Movies
 
+// Get the average of Drama Movies
+function dramaMovieRate(array) {
+  var dramaMovie=array.filter(function(item){
+    return (item.genre.indexOf('Drama')===1);
```
tawebbcn: Care! 
This filter will return the elements that have the 'Drama' in the second position of `genre`. So will not get the elements that have genre like `['Drama']`.
One way to approach this, can be:
```
return (item.genre.indexOf('Drama') !== -1);
```
With this will take all the elements that have `Drama` on the array `genre`

PR #60, opened 2018-03-22
BCN-Sam-Miriam

MiriamAparicio:
@samanthaferreira

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,13 +1,29 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
+// Turn duration of the movies from hours to minutes
+
 
 
 // Get the average of all rates with 2 decimals 
+function ratesAverage(array) {
+  var totalRates = array.reduce(function(accumulator, item) {
+    return accumulator + item.rate;
+  }, 0);
+
+  return parseFloat(totalRates / array.length);
+}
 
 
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,13 +1,29 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
+// Turn duration of the movies from hours to minutes
+
 
 
 // Get the average of all rates with 2 decimals 
+function ratesAverage(array) {
+  var totalRates = array.reduce(function(accumulator, item) {
```
tawebbcn: :metal:

starter_code/src/movies.js
```diff
@@ -1,13 +1,29 @@
...
+
+  return parseFloat(totalRates / array.length);
+}
 
 
 // Get the average of Drama Movies
 
+function dramaMoviesRate(array) {
+  var dramaMovies = array.filter(function(movie) {
+    return movie.genre.includes("Drama");
```
tawebbcn: :+1:

starter_code/src/movies.js
```diff
@@ -1,13 +1,29 @@
...
+  }, 0);
+
+  var averageDrama = ratesAverage(dramaMovies);
 
+  return averageDrama;
```
tawebbcn: you can just ```return ratesAverage(dramaMovies)```, but it's okay :smile:

PR #62, opened 2018-03-22
PAR Patrycja & Clément

Kleemdub:
Almost done, we failed on the bonus

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,155 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
+function turnHoursToMinutes(array){
+    var returnArray = 
+        array.map(function(oneMovie){
 
+            var duration = oneMovie["duration"];
+            var hour = 0;
+            var minute = 0;
+            duration = duration.split(" ");
+
+            for(var i = 0; i < duration.length; i++){
+              if(i == 0 && duration[i].indexOf('h') != -1){
+                hour = parseInt(duration[0])*60;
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,155 @@
...
+    return byTitle20;
+    
+}
 
-
+  orderAlphabetically(movies);
+  
 // Best yearly rate average
+
+function bestYearAvg(array){
```
ta-web-paris: You can try with this code :) 

```js

function bestYearAvg(moviesArr) {
  var ratesYear = {};
  var moviesYear = {};
  var averageYear = {};
  if (moviesArr.length === 0) {
    return undefined;
  }
  moviesArr.forEach(function (e) {
    if (ratesYear[e.year]) {
      moviesYear[e.year] += 1;
      ratesYear[e.year] += parseFloat(e.rate);
      averageYear[e.year] = parseFloat((ratesYear[e.year] / moviesYear[e.year]).toFixed(2));
    } else {
      ratesYear[e.year] = parseFloat(e.rate);
      moviesYear[e.year] = 1;
      averageYear[e.year] = parseFloat(e.rate);
    }
  });
  var year = Object.keys(averageYear).reduce(function (a, b) {
    if (averageYear[a] === averageYear[b]) {
      if (b < a) {
        return b;
      }
      return a;
    } else if (averageYear[a] > averageYear[b]) {
      return a;
    }
    return b;
  });
  return 'The best year was ' + year + ' with an average rate of ' + averageYear[year];
}

```

ta-web-paris:
Good job, this exercise wasn't easy but you manage to (almost) finish it :) 👍 

Keep up the good work ! 🥇

PR #63, opened 2018-03-22
BCN - RAPHA & GEORGE

raphamontenegro:
Took us a lot of time


starter_code/src/movies.js
```diff
@@ -5,8 +5,36 @@
 // Get the average of all rates with 2 decimals 
 
 
+
+function ratesAverage(array) {
```
tawebbcn: :+1:

starter_code/src/movies.js
```diff
@@ -5,8 +5,36 @@
...
+  return totalRate / array.length;
+};
+
+
 // Get the average of Drama Movies
 
+function dramaMoviesRate(array) {
+  var dramaMovies = movies.filter(function(elem) {
+    if (elem.genre.includes("Drama")) {
+      return true;
```
tawebbcn: Instead of this, you can just put:
```return elem.genre.includes("Drama")```
this will return true or false

PR #64, opened 2018-03-22
PAR Brianna & Louis & Billy

Yaats:
Don't know what is going on with the average functions.
Also don't know what went wrong afterwards.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,18 +1,128 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
+// Turn duration of the movies from hours to minutes
+// "2h 33min" -> 153
+// "2h"       -> 120
+// "33min"    -> 33
+
+// "2h"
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,18 +1,128 @@
...
+    } else {
+      return 0;
+    };
+    return blah;
+});
+}
 
 // How many movies did STEVEN SPIELBERG
 
+var howManyMovies = movies.filter (function(movies){
```
ta-web-paris: You were missing a parenthesis, this is why your code wasn't working anymore ;)

ta-web-paris:
```js
var howManyMovies = movies.filter(function(movies) {
  return movies.director.includes("Steven Spielberg");
});
```

ta-web-paris:
This exercise wasn't easy, so if you wan't to go further, here is a correction : 

https://gist.github.com/ta-web-paris/cec74228db050e437e1e02e09bf529e7

Keep up the good work 👍

PR #65, opened 2018-03-22
PAR Sami & Luke

Sami-Az:
We had some issues while trying to filter elements from the array


ta-web-paris:
If you are struggling with filter method, you should take a look at the documentation, it will help you :)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter

Anyway this exercise wasn't easy, so if you want to go further, here is a correction : https://gist.github.com/ta-web-paris/cec74228db050e437e1e02e09bf529e7

PR #67, opened 2018-03-22
BCN - Jenny - Lennie

Fantasmy:
@lenniez

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -4,8 +4,36 @@
 
 // Get the average of all rates with 2 decimals 
 
+var totalRates = movies.reduce(function (sum, movie){
+    return sum + Number(movie.rate);
+}, 0);
+
+console.log(totalRates);
+
+function ratesAverage (movies) {
+    return (totalRates/movies.length).toFixed(2);
+}
+
+//console.log(ratesAverage(movies));
+
+
 
 // Get the average of Drama Movies
+
+var dramaMovies = movies.filter(function(movie){
+    return movie.genre.includes('Drama');
+
+}); 
+
+var totalDramaRates = dramaMovies.reduce(function (sum, movie){
+    return sum + Number(movie.rate);
+}, 0);
+
+console.log(totalDramaRates);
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -4,9 +4,37 @@
 
 // Get the average of all rates with 2 decimals 
 
+var totalRates = movies.reduce(function (sum, movie){
```
tawebbcn: You should put this inside the ratesAverage function or in another fuction.

tawebbcn: If we change the data, the variable can't be modified, if you put it inside a function you can always call the function.

starter_code/src/movies.js
```diff
@@ -4,9 +4,37 @@
...
+    return sum + Number(movie.rate);
+}, 0);
+
+console.log(totalRates);
+
+function ratesAverage (movies) {
+    return (totalRates/movies.length).toFixed(2);
```
tawebbcn: .toFixed() method returns a string. You have to convert it into a number.

starter_code/src/movies.js
```diff
@@ -4,9 +4,37 @@
...
+
+}); 
+
+var totalDramaRates = dramaMovies.reduce(function (sum, movie){
+    return sum + Number(movie.rate);
+}, 0);
+
+console.log(totalDramaRates);
+
+var dramaMoviesRate = (totalDramaRates/dramaMovies.length).toFixed(2);
```
tawebbcn: this can be done with the function in the previous iteration.

starter_code/src/movies.js
```diff
@@ -4,9 +4,37 @@
...
+
+function ratesAverage (movies) {
+    return (totalRates/movies.length).toFixed(2);
+}
+
+//console.log(ratesAverage(movies));
+
+
 
 // Get the average of Drama Movies
```
tawebbcn: All of this should be inside a function (otherwise it won't pass the tests)

PR #69, opened 2018-03-22
PAR Maggie & Pierre

PierreJab:


starter_code/src/movies.js
```diff
@@ -1,20 +1,106 @@
...
+    });
+    counter = 0;
+    for (var i = 0; i < length.dramaArray; i++){
+        if(dramaArray.rate === ''){
+            dramaArray.rate = 0;
+            counter += 1;
+        };
+    }
+
+    if (dramaArray === []){
```
ta-web-paris: Try to use this instead :  

```js

if (dramaArray.length === 0) {
    return undefined;
  }

```
It will do the trick

starter_code/src/movies.js
```diff
@@ -1,20 +1,106 @@
...
 
 // How many movies did STEVEN SPIELBERG
+function howManyMovies(array){
+    if (array.length === 0){
+        return undefined;
+    }
+    counter = 0;
+    for (var i = 0; i < array.length; i++){
+        if ((array[i].director === 'Steven Spielberg') 
+        /* && (movie.genre.indexOf("Drama") !== -1  *//* || movie.genre === "Drama") */){
```
ta-web-paris: Okay so your code is good but as you didn't add the "Drama movie" condition, the counter was to high and didn't match Jasmin expectation, you can try with this code :  


```js
function howManyMovies(moviesArr) {
  if (moviesArr.length === 0) {
    return undefined;
  }
  var directorMovies = moviesArr.filter(function (e) {
    return e.director === 'Steven Spielberg' && e.genre.indexOf('Drama') !== -1;
  });
  return 'Steven Spielberg directed ' + directorMovies.length + ' drama movies!';
}

```

ta-web-paris: actually you were really close :)

ta-web-paris:
Good job, this exercise wasn't easy!

Keep up the good work 👍

PR #70, opened 2018-03-22
PAR LEO P. & ELISE

elisegriset92:
why is the following not working for iteration 3?:

var drama = movies.filter(function(movie){
return movies.genre
});

this returns an array of objects and not the arrays under genre that we would have then been able to access via indexOf.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,11 +1,76 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
+/* eslint no-restricted-globals: 'off' */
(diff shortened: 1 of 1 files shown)


ta-web-paris:
as you are passing "movie" as a parameter, you should use your parameter in the function, as you are trying to access movies.genre and movies isn't define, it won't work.

In your code movies refer to the array, and movie refers to each element of this array. 

For example : 

```js

var words = ['spray', 'limit', 'elite', 'exuberant', 'destruction', 'present'];

const result = words.filter(word => word.length > 6);

console.log(result);
// expected output: Array ["exuberant", "destruction", "present"]


```

PR #103, opened 2018-04-06
MIA-FTWD-Eleyni+Rafael-lab-javascript-all-times-movies

erodrig1010:
Still trying to pass some tests.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -5,81 +5,65 @@ var movies = require("../src/data")
 
 // Turn duration of the movies from hours to minutes 
 function turnHoursToMinutes(array){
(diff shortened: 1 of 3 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,109 @@
...
+  }
+  result = sumAverage / filtered.length;
+  return Number(result.toFixed(2));
+}
+
 
 
 // Order by time duration, in growing order
+function orderByDuration(array) {
+  var minutes = turnHoursToMinutes(array);
```
sandrabosk: Good approach!

sandrabosk:
Good work! Thank you for submitting.

PR #114, opened 2018-04-09
MIA-FTWD-Ruben Vargas,Ryan Craig Martin-lab-javascript-all-times-movies

Ambition101:


starter_code/src/movies.js
```diff
@@ -1,20 +1,146 @@
...
+// function ratesAverage(arr) {}
 
 
 // Get the average of all rates with 2 decimals 
+function ratesAverage(movies) {
+  var theSum = movies.reduce(function (sum, movie) {
+    return sum + Number(movie.rate);
+  }, 0);
+  var rateAvg = theSum / movies.length;
+  return Math.floor(rateAvg * 100) / 100
```
MarcosPalaciosIII: for a rating you would want to use Math.round() not Math.floor() since you don't want to lower a rating that would be 4.7 to be a 4 rather than a 5 that it should be.

starter_code/src/movies.js
```diff
@@ -1,20 +1,146 @@
...
+// ==========================================
 
 // Order by title and print the first 20 titles
 
+function orderAlphabetically(array) {
+  var ordered = array.map(function (movie) {
+      return movie.title
+    })
+    .sort()
+    .slice(0, 4)
```
MarcosPalaciosIII: .slice(0, 20)

MarcosPalaciosIII:
good work on the lab. you were heading in the right direction. Keep up the good work

PR #142, opened 2018-05-24
BER-JENS

je0705:
Not done yet. Need Feedback and few hints please.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -31,19 +31,33 @@ turnHoursToMinutes(moviesArray);
 
 // Get the average of all rates with 2 decimals 
 
-function ratesAverage(moviesArray) {
-  
-  var numMov = moviesArray.length;
+  function ratesAverage(movies) {
+    
+    var numMov = movies.length;
 
-  var totalAvg = moviesArray.reduce(function (acc, elem) { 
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,12 +1,63 @@
...
+  function ratesAverage(movies) {
+    
+    var numMov = movies.length;
+
+    var output = [];
+    for (var i = 0; i < numMov; ++i) {
+      output.push(movies[i].rate);
+    }
+    var totalAvgString = output.reduce(function(acc, elt) {
+      return (acc + elt) / numMov;
```
ta-web-ber: You wrote /numMov at the wrong place. It should be:
```
    
    var numMov = movies.length;

    var output = [];
    for (var i = 0; i < numMov; ++i) {
      output.push(movies[i].rate);
    }
    var totalAvgString = output.reduce(function(acc, elt) {
      return (acc + elt);
    });
    totalAvg = parseFloat(totalAvgString)/ numMov;

    // var factor = Math.pow(10, 2);
    // totalAvg = Math.round(totalAvg * factor) / factor;
    // console.log(totalAvg);
    return totalAvg;
  }
  ratesAverage(movies);

```

PR #143, opened 2018-05-24
BER - John Fish - Movies Javascript - done

j31:
I would have liked to spend more time to refactor this code; I could tighten it up quite a bit.

To be open, I was a little frustrated in places by the quality of this spec and the testing scripts.  For example, it expected the average of two films: one 8 and one "not rated" to be 4, when this should pretty clearly be 8.  In another, the "duration" was fed in to the test script as an integer (with no units given).  In the actual data set, "duration" is a string in the format "2h 20min".  So it's not clear what to do with the integer 200 in this case, and it's not consistent with the data format (so would not have been allowed to be stored in this format).  Also, it would have been nice to have been shown how to read the data file from the movies.js.  These are small things, but the days are long, and I felt like I got 80% worked through in 20% of the time, and then I spend the rest of the time trying to understand the spec via the testing scripts.  Still, it's a good project, but it would benefit from more polish in the instructions and the testing scripts.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,2142 @@
+
 /* eslint no-restricted-globals: 'off' */
+
+
 // Turn duration of the movies from hours to minutes 
+
+function turnHoursToMinutes(movies){
+  return movies.map((movie, i, movies) => {
+    if (typeof movie.duration == "number") {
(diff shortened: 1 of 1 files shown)


ta-web-ber:
@mc100s

PR #145, opened 2018-05-24
BER - Silvio

silvio-galli:
Last function missing


silvio-galli:
Added last function; not very glad that I did not find the way to reuse the `ratesAverage()` function inside this one.

ta-web-ber:
@mc100s

PR #147, opened 2018-05-26
BER -Laure Dassy movies 

LaureDas:


starter_code/src/movies.js
```diff
@@ -1,20 +1,93 @@
...
+                return b.duration;
+            }
+        }
+   return a.duration - b.duration;
+ });
+ 
+ return moviesSort;
+ }
 
 // How many movies did STEVEN SPIELBERG
```
ta-web-ber: Director is not an array and genre it is. The solution would be:
```
  if (movies.length===0)
    return undefined;

    var dramaArr= movies.filter(function(elem){
 
    return elem.genre.indexOf('Drama') !==-1 && (elem.director === 'Steven Spielberg')
   
    });
    
return "Steven Spielberg directed " + dramaArr.length + " drama movies!";
}

PR #149, opened 2018-05-27
WPTBCN Alejandro Close

alejandroclose:
All iterations done. The bonus looks interesting but I guess for now I don't have time!!! 
Note to me: Should come back to this in the future :)


tawebbcn:
Muy bien Alejandro,
No pasa el Test de Jasmine pero funciona todo correctamente y tiene una lógica bastante correcta 👍!

PR #154, opened 2018-05-31
PAR Manelle & Jihad

ManelleG:
@eloji9

There are some codes that are incorrect and we didn't have time to delete them.


ta-web-paris:
Be careful you have to pay attention to Jasmine errors, even if some errors are not relevant as we changed the instructions, but you still have to try to make it green ;) 

Otherwise good job, and if you want to go further here a correction of this exercise : https://gist.github.com/ta-web-paris/cec74228db050e437e1e02e09bf529e7

PR #156, opened 2018-06-01
Par Meri and Marie

marieglr:
we struggled with one error from Jasmine saying the average function should work even if some drama movies do not have a rate, we could not get rid of this one... do you know how we could ? 
Thanks !


starter_code/src/movies.js
```diff
@@ -1,15 +1,2111 @@
...
+//we create a new array composed only of drama movies that have a rate property
+// var ratedDramaMovies = dramaMovies.filter(function(aDramaMovie){
+//   return aDramaMovie.rate;
+// });
+
+var ratedDramaMovies = [];
+var nnn=0
+
+dramaMovies.forEach(function (dramaMovie){
+  if (!dramaMovie.rate){
```
ta-web-paris: not sure what you try to did with this if statement, but when you remove it, it will actually put some content in your array, otherwise your array will stay empty ;)

ta-web-paris: Ok I understand what you've try to do, but you should do it like this : 

```js 

dramaMovies = [{ name: "hello", rate: 8 }, { name: "prout", rate: "" }];
dramaMovies.forEach(function(dramaMovie) {
  if (dramaMovie.rate == "") {
    dramaMovie.rate = 0;
    ratedDramaMovies.push(dramaMovie);
  } else ratedDramaMovies.push(dramaMovie);
});

```

We have the excpected output (4) but for some reason it's not working on jasmine.. I'm gonna take a look at it

ta-web-paris:
It's strange I manage to write a code that output what is expected by jasmine but the error is still there, I'm gonna take a look, otherwise good job! 

If you want to have a correction of the exercise you can check this link : https://gist.github.com/ta-web-paris/cec74228db050e437e1e02e09bf529e7

keep up the good work and if you have any question regarding the concept we've seen, let me know!  👍

PR #181, opened 2018-06-14
BCN Daniela-Kujawa and Montse-Badia 

danielakujawa:
@MontseBadia


starter_code/src/movies.js
```diff
@@ -1,20 +1,95 @@
...
+}
+
+
 
 // Order by title and print the first 20 titles
 
+function orderAlphabetically(movies){
+
+    var orderedMovies = movies.sort(function(item1, item2){
+    return item1.title - item2.title;
```
tawebbcn:
```
'string' - 'string'
<--NaN
```

starter_code/src/movies.js
```diff
@@ -1,20 +1,95 @@
...
+  if(orderedTitles.length > 20){
+    
+    return orderedTitles.splice(0, 20);
+ 
+   } 
+
+   
+  return orderedTitles;
+
+}
```
tawebbcn: Here you have the solution of this iteration:
```
  moviesArray.sort(function (a, b) {
    return a.title < b.title ? -1 : 1;
  });
  var top20Movies = [];
  var limit = 20;
  if (moviesArray.length < 20) {
    limit = moviesArray.length;
  }
  for (var i = 0; i < limit; i++) {
    top20Movies.push(moviesArray[i].title);
  }
  return top20Movies;
}
```

PR #194, opened 2018-06-15
MIA Giannini, Junior, Christine Beck STILL IN PROGRESS:

GianniniCharles:
STILL IN PROGRESS: Will push again when complete

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,81 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
-
-
+////****juniors */
+function turnHoursToMinutes(moviesArray) {
+    var moviesWithCorrectTimeFormat = [];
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,81 @@
...
+},0);
+var ratesAvg = ratesSum/ratesArrayFloat.length;
+var ratesAvgRounded = Math.round(ratesAvg*100)/100;
+console.log(ratesAvgRounded);
+return ratesAvgRounded;
+}
+ratesAverage(movies)
+function dramaMoviesRate(array){
+ var dramaMovies =   array.filter(function(array){
+return array.genre === 'Drama';
```
sandrabosk: Since we are getting array from array.genre we can't use === to check if something is in the array. We should do:

sandrabosk:
Good job so far. Please add the rest.

PR #196, opened 2018-06-15
MIA Jackson Mades

Jackloco:
ONly 10 failures :D

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,53 +1,64 @@
 /* eslint no-restricted-globals: 'off' */
+
 // Turn duration of the movies from hours to minutes
-function turnHoursToMinutes(movies){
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,167 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
 
+// Turn duration of the movies from hours to minutes
+function turnHoursToMinutes(movies) {
+    return movies.map(function (curMovie) {
```
sandrabosk: Overly complicated. I like that you took very thorough approach but it might have xyz different combinations so this is not very efficient. Take a look at the solution Nick gave to you.

sandrabosk:
Very good work. Keep it like this. Thank you for submitting.

PR #198, opened 2018-06-17
MIA Giannini, Junior, Christine Beck

GianniniCharles:
Completed functions howManyMovies, orderAlphabetically. Requesting assistance on bestYearAvg lines 148 to end.

the test is asking for correct answer to a multiple elements array, but to organize the ratings into an average, I presume it'd make sense to use an object to represent the averages from each year instead of arrays. Which approach would you recommend? Question also asked to TA Manny


MarcosPalaciosIII:
good work on the lab. For the best year portion you can ask me later on and I'll explain it a little better

PR #226, opened 2018-08-09
PAR John & Emmanuelle

emmanuellejaldorau:
@johndimitroff

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,12 +1,48 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
 
+function ratesAverage(array) {
 
-// Get the average of all rates with 2 decimals 
-
+    // Get the average of all rates with 2 decimals 
+    var accumulatedRates = 
+    array.reduce(function (acc, oneMovie) {
+        return acc + Number(oneMovie.rate);
+    }, 0);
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,12 +1,48 @@
...
+                else {
+                return acc + Number(oneMovie.rate);}
+            }, 0);
+        
+        var averageRateNotRounded = accumulatedRates / dramaArray.length;
+        var averageRateRounded = Math.round(averageRateNotRounded *100)/100;
+        
+        return averageRateRounded;
+      }
+    }
```
marieglr: Good job, this works ! 
a few comments : 
- we now have access to a method "includes", so we dont really need anymore this little trick with "indexOf > 0" to check if something exists.
- Also,  to simplify your code and keep it dry, you can re-use the rateAvarage function above after you have filtered the array 
- to get rid of the last error on jasmine, you can wrap the code inside the function (or only the return) inside an if statement :   
if (dramaArray.length > 0) {//code here}

marieglr:
The logic is good and thank you for naming your variables in a very clear way ! 
Keep up like that :)
Here is a complete solution for the exercise: https://github.com/ironhack-paris-webdev-august-2018/lab-all-times-movies2/blob/master/starter_code/src/movies.js

PR #227, opened 2018-08-09
PAR Julie and Abi

abisummers:
@youlie03

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,33 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
+// Turn duration of the movies from hours to minutes
+function turnHoursToMinutes() {}
+// Get the average of all rates with 2 decimals
 
+function ratesAverage(myArray) {
+  var sum = myArray.reduce(function(acc, curr) {
+    return acc + parseFloat(curr.rate);
+  }, 0);
+  //return sum;
 
-// Get the average of all rates with 2 decimals 
-
+  var average = sum / myArray.length;
+  return average;
+}
 
 // Get the average of Drama Movies
-
+function dramaMoviesRate(myArray) {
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,33 @@
...
+  var sum = myArray.reduce(function(acc, curr) {
+    return acc + parseFloat(curr.rate);
+  }, 0);
+  //return sum;
 
-// Get the average of all rates with 2 decimals 
-
+  var average = sum / myArray.length;
+  return average;
+}
```
ta-web-paris: great, clear and efficient!

starter_code/src/movies.js
```diff
@@ -1,20 +1,33 @@
...
 
 // Get the average of Drama Movies
-
+function dramaMoviesRate(myArray) {
+  var genderFilter = myArray.filter(function(movieGenre) {
+    if (myArray.genre === "Drama") {
+      return genderFilter;
+    }
+  });
```
ta-web-paris: small error here : this is how the filter method works : 
- it makes a loop on the array it is passed on, 
- for each element of the array, it checks if the condition you are returning is true
- if it is true, it keeps this element inside the new array it creates. 

```
  var genderFilter = myArray.filter(function(movie) {
    return movie.genre.includes("Drama");
  });
```
and then you can pass the ratesAverage function you created above on the genderFilter var to get the average rate of drama movies : 

```
function dramaMoviesRate(myArray) {
  var genderFilter = myArray.filter(function(movie) {
    return movie.genre.includes("Drama");
  });
   return ratesAverage(myDramaArr);
}

```

(you can add conditions to this code to handle all of the special cases Jasmine tests for, like if myArray is empty)

ta-web-paris:
Keep praticing loop methods, they are super useful :)
a complete solution for this exercise can be found here : 
https://github.com/ironhack-paris-webdev-august-2018/lab-all-times-movies2/blob/master/starter_code/src/movies.js

PR #228, opened 2018-08-09
PAR Manon & Pauline

paulettepaillette:
@msln8


starter_code/src/movies.js
```diff
@@ -4,9 +4,40 @@
...
+
+  // get the average
+  function dramaMoviesRate (myArray) {
+  var myDramaArray =  dramaMoviesThing(myArray);
+  var myDramaAverage = 
+    myDramaArray.reduce(function(acc, oneRate){
+      return acc + parseFloat(oneRate.rate);
+    }, 0);
+    return (myDramaAverage / myDramaArray.length);
+  }
```
ta-web-paris: Yes, good job !
since the second part of that function is exactly the same as the ratesAverage function, you chan just call it on your drama array : 
```
 function dramaMoviesRate (myArray) {
  var myDramaArray =  dramaMoviesThing(myArray);
    return ratesAverage(myDramaArray);
  }
```
To get rid of the error you still have on Jasmine, you can add this condition : 
```
 function dramaMoviesRate (myArray) {
  var myDramaArray =  dramaMoviesThing(myArray);
  if (myDramaArr.length > 0) {
    return ratesAverage(myDramaArray);
  }
  }
```
(The very last error on Jasmin doesn't make sense to me, but you can add this condition in the ratesAverage function to get rid of it : 
    if (curr.rate === "") {
      curr.rate = 0;
    }
)

ta-web-paris:
The logics of reduce and filter methods seems fully understood, well done ! 
all the solutions to this lab can be found here : https://github.com/ironhack-paris-webdev-august-2018/lab-all-times-movies2/blob/master/starter_code/src/movies.js

PR #230, opened 2018-08-09
Par Yvens et Constance

Yvens16:
@Cr27910


ta-web-paris:
The logic on this function is good ! You understood the syntax for reduce and the way a function works :+1: .
All the solutions for this lab can be found here : https://github.com/ironhack-paris-webdev-august-2018/lab-all-times-movies2/blob/master/starter_code/src/movies.js
We can also do a reinforcement lesson on this lab to go through all the iterations step by step together

PR #231, opened 2018-08-09
PAR Mohamed @SMoZam & Luca

lucamerzi:
NOTE: the "movies" variable is not available in index.html, I have the impression that the file "data.js" is not correctly linked, since it is in the "src" folder.
-- Luca
@SMoZam


starter_code/src/movies.js
```diff
@@ -1,20 +1,117 @@
...
 
-// Get the average of Drama Movies
 
+// ==========================================================
 
+// Get the average of Drama Movies
+function dramaMoviesRate(arrayRate){
+  let filteredArray = arrayRate.filter(function(el){
+    return (el.genre.indexOf("Drama") !== -1)
+  });
```
ta-web-paris: this works perfect, but there is also a new method called "includes" that makes unnecessary to use this little trick, you can do the same thing by saying : 
```
return (el.genre.includes("Drama"))
```

starter_code/src/movies.js
```diff
@@ -1,20 +1,117 @@
...
+  let result = orderedTitle.sort(function(a,b){
+  return (a.title < b.title) ? -1 : 1 ; 
+  })
+
+  if (result.length >= 20){
+    return result.slice(0, 20);
+  } else {
+    return result;}
+}
+
```
ta-web-paris: Jasmine asks for an array of titles, so you can start the function with a map : 
Also, remember that the sort methods works behing the scenes by comparing the characters unicodes, and that upper case and lower case letters have diferent unicodes, so this can lead to incorrect alphabetical sorting.
```
function orderAlphabetically(orderedTitle){
  //Get an array with only the movie titles
  var movieTitles = orderedTitle.map(function(oneMovie) {
    return oneMovie.title;
  });

  //Sort the movie titles in alphabetical order
  movieTitles.sort(function(titleA, titleB) {
    titleA.toUpperCase(); // ignore upper and lowercase
    titleB.toUpperCase(); // ignore upper and lowercase
    if (titleA < titleB) {
      return -1;
    }
    if (titleA > titleB) {
      return 1;
    }
  });

  //keep only the first 20 ones
  return movieTitles.slice(0, 20);
}
```

starter_code/src/movies.js
```diff
@@ -1,20 +1,117 @@
...
+    return undefined
+  }
+
+  let filteredArray = arr.filter(function(el){
+    return (el.genre.indexOf("Drama") !== -1)
+  });
+
+  let spielberg = filteredArray.filter(function(el){
+    return el.director === "Steven Spielberg"
+  })
```
ta-web-paris: Again, this works just fine, but to reduce your code, you can combine these two filters using the && operator : 
```
let filteredArray = arr.filter(function(el){
    return ((el.genre.indexOf("Drama") !== -1) && (el.director.includes( "Steven Spielberg")))
  });

```

ta-web-paris:
Great job !!
To get rid of all the small errors Jasmine is tracking, and to see the bonus, you can check this code : 
https://github.com/ironhack-paris-webdev-august-2018/lab-all-times-movies2/blob/master/starter_code/src/movies.js

PR #236, opened 2018-08-16
MIA FTWD MikeWithers

mwithers313:


starter_code/src/movies.js
```diff
@@ -1,20 +1,209 @@
...
+    }
+
+
+
+
+function dramaMoviesRate(movies){
+
+
+var includesDrama = movies.filter((eachNum)=>{
+     return eachNum.genre == ("Drama");
```
sandrabosk: You're safer if you use `.includes()` in this case.

sandrabosk:
Good work so far. Thank you for submitting.

PR #237, opened 2018-08-16
MIA Douglas DeJong FTWD 

DougDeJong:
Still 11 errors - getting there.


MarcosPalaciosIII:
seems like your code should be working but for reason it is not registering your code on specrunner. 

overall good job on the all times movies. 

keep up the good work.

PR #239, opened 2018-08-17
MIA FTWD Nick Crispino

crispinonicky:


MarcosPalaciosIII:
nice try with the all times movies. You have to check in the drama rating function for possible empty ratings and still calculate avg on the number of drama movies as well as have an undefined rating if there are no drama movies. 

Keep up the nice work.

PR #289, opened 2018-09-29
[PAR PT Rita Mernissi] 

mernri:
first iteration works but doesn't pass the tests

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,6 +1,19 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
 
+function turnHoursToMinutes(duration) {
+    var splitDuration = duration.split("h ");
+    var hours = 60 * parseInt(splitDuration[0]);
+    var splitMinutes = parseInt(splitDuration[1].slice(0,2));
+    var totalInMinutes = hours + splitMinutes;
+   return String(totalInMinutes);
+}
+  
+var newDuration = movies.map(function(object){
+    object.duration = turnHoursToMinutes(object.duration);
+    return object;
+  })
+
 
 // Get the average of all rates with 2 decimals 
 


starter_code/src/movies.js
```diff
@@ -1,6 +1,19 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
 
+function turnHoursToMinutes(duration) {
```
ZwoRmi: This function should return an array, you need to return the result of the `movies.map` and not a duration

starter_code/src/movies.js
```diff
@@ -1,6 +1,19 @@
...
+    var splitDuration = duration.split("h ");
+    var hours = 60 * parseInt(splitDuration[0]);
+    var splitMinutes = parseInt(splitDuration[1].slice(0,2));
+    var totalInMinutes = hours + splitMinutes;
+   return String(totalInMinutes);
```
ZwoRmi: My mistake, it should be a Number, not a String, so you can remove the `String` function

starter_code/src/movies.js
```diff
@@ -1,6 +1,19 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
 
+function turnHoursToMinutes(duration) {
+    var splitDuration = duration.split("h ");
```
ZwoRmi: In the data.js file, you can see that sometimes the duration is expressed only in minutes, like "22min" or only in hours, like "2h". Your code will break for those test cases

starter_code/src/movies.js
```diff
@@ -1,6 +1,19 @@
...
+    var hours = 60 * parseInt(splitDuration[0]);
+    var splitMinutes = parseInt(splitDuration[1].slice(0,2));
+    var totalInMinutes = hours + splitMinutes;
+   return String(totalInMinutes);
+}
+  
+var newDuration = movies.map(function(object){
+    object.duration = turnHoursToMinutes(object.duration);
+    return object;
```
ZwoRmi: You have to return a new object, using `Object.assign()`

PR #290, opened 2018-09-30
PAR PT - Julie Chane Ching

JulieCC:
Hi @ta-web-paris , 
j'ai presque tout fait sauf orderByDuration qui ne fonctionne pas :(


starter_code/src/movies.js
```diff
@@ -1,20 +1,102 @@
...
+
+  if (dramaMovies.length == 0) {
+    return undefined;
+  } else {
+    var result = ratesAverage(dramaMovies);
+    return Math.round(result * 100) / 100;
+  }
+}
 
 // Order by time duration, in growing order
```
ta-webpt-par: You need to keep in mind that the code runs and you already transformed all the durations in minutes and that what is asked here is to return "movieArray" in a growing order.
So first, you sort the durations keeping in mind the alphabetical order.
you return these durations then the movieArray.
This is a case of a function inside a function. 
Keep in mind the final objective but start from the smallest function.

    moviesArray.sort(function (a, b) {
        if (a.title > b.title) {
          return 1;
        }
      }
    });
    return moviesArray;
  }

PR #291, opened 2018-10-02
[PAR-PT] Elodie

EloK6:
WIP
@ta-webpt-par

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,61 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
-
+function turnHoursToMinutes (duration) {
+    var duration2=duration.split(" ");
+    var durationTotal=0;
+    
+    if (duration2.length===1) {
+    durationTotal+=parseInt(duration2[0]);
+    return durationTotal;
+    } else {
+    var durationHour=parseInt(duration2[0]);
+    var durationMin=parseInt(duration2[1]);
+    var durationTotal=(durationHour*60)+durationMin;
+    }
+    return durationTotal;
+  }
 
 // Get the average of all rates with 2 decimals 
-
+// Don't understand the tips "Maybe you want to "reduce" the data to a single value"
+function ratesAverage (array) {
+    var calculAverage = 0;
+    var average = array.map(function (movie) {
+        calculAverage +=parseFloat(movie.rate);
+    });
+    return calculAverage / array.length;
+}
 
 // Get the average of Drama Movies
-
+function dramaMoviesRate (array) {
+    var dramaMovies = array.filter(function(movie) {
+        return movie.genre.includes("Drama");
+    });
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,61 @@
...
+    if (duration2.length===1) {
+    durationTotal+=parseInt(duration2[0]);
+    return durationTotal;
+    } else {
+    var durationHour=parseInt(duration2[0]);
+    var durationMin=parseInt(duration2[1]);
+    var durationTotal=(durationHour*60)+durationMin;
+    }
+    return durationTotal;
+  }
```
ta-webpt-par: The idea here is to be able to return the whole array of movies with the duration transformed in minutes for all of them. To do so, you need to be able to access the hours and the minutes without "h" and "min" and do the mathematical operations on all the cases (movie with h and min, movie with just h and movie with just min).

starter_code/src/movies.js
```diff
@@ -1,20 +1,61 @@
...
 // Get the average of all rates with 2 decimals 
-
+// Don't understand the tips "Maybe you want to "reduce" the data to a single value"
+function ratesAverage (array) {
+    var calculAverage = 0;
+    var average = array.map(function (movie) {
+        calculAverage +=parseFloat(movie.rate);
+    });
+    return calculAverage / array.length;
+}
```
ta-webpt-par: Very good

starter_code/src/movies.js
```diff
@@ -1,20 +1,61 @@
...
 // Order by time duration, in growing order
-
+function orderByDuration (array) {
+    var order = array.sort(function (a, b) {
+        if (a.duration !== b.duration) {
+            return a.duration - b.duration;
+        }
+    });
+    return order;
+}
```
ta-webpt-par: You need to be able to order the movies alphabetically by their title if they have the same duration.

starter_code/src/movies.js
```diff
@@ -1,20 +1,61 @@
...
-
+function howManyMovies (array) {
+   var count=0;
+    var StevenSpielberg = array.filter(function(movie) {
+        if(movie.director===" Steven Spielberg" && movie.genre.includes("Drama")) {
+            count += 1;
+        }
+    });
+    return count;
+}
```
ta-webpt-par: It's better to divide the whole code in a way that you first select the drama movies then check the director to be Spielberg

starter_code/src/movies.js
```diff
@@ -1,20 +1,61 @@
...
 
 // Get the average of Drama Movies
-
+function dramaMoviesRate (array) {
+    var dramaMovies = array.filter(function(movie) {
+        return movie.genre.includes("Drama");
+    });
+    return ratesAverage(dramaMovies);
+}
```
ta-webpt-par: You need to add the possibility to return the average even if one of the movies does not have rate by adding an:
if (dramaMovies.length == 0) {
     return undefined;
}

PR #292, opened 2018-10-02
Juliana JEAN-MARIE / Sylvia MAVUBA - done

JulianaJM:


starter_code/src/movies.js
```diff
@@ -1,20 +1,138 @@
...
 
-// Order by title and print the first 20 titles
+  if(movies.length === 0)  return undefined;
+
+  if(spDramaMoviesLength === 1) return 'Steven Spielberg directed 1 drama movies!';
 
+  if(spDramaMoviesLength === 4) return 'Steven Spielberg directed 4 drama movies!';
+
+  
+  if(spDramaMoviesLength === 0) return 'Steven Spielberg directed 0 drama movies!';
```
ta-webpt-par: That's not really what we wanted you to do. We expected this instead:
```js
  if (movies.length === 0) return undefined;
  return 'Steven Spielberg directed '+ spDramaMoviesLength + ' drama movies!';
```

starter_code/src/movies.js
```diff
@@ -1,20 +1,138 @@
...
+      duration = (parseInt(hours) * 60) + parseInt(minutes);
+    } else {
+      if(durationArray[0].indexOf('h') !== -1) {
+        //calc
+        hours = durationArray[0].replace("h","");
+        duration = parseInt(hours) * 60;
+
+      } else if(durationArray[0].indexOf('min') !== -1) {
+        duration = parseInt(durationArray[0].replace("min", ""));
+      }
```
ta-webpt-par: Your code could be shorter using a regex to parse this string

PR #293, opened 2018-10-11
IRONHACK -  Change of iteration 1 to make it simpler

mc100s:
I have created a new simpler iteration 1. The old iteration 1 is now a bonus.

**Why this change?**
Because students used to spend more than an hour on the first iteration. Instead, I've created a simpler `map` exercise they can do in 5 minutes.

**Solution** 
I have written some tests and check if everything works correctly. You can find below an example of a solution.

```javascript
function getSummary(movies) {
  return movies.map(function(m){
    return m.title + ' by ' + m.director + ' (' + m.year + ')'
  })
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,17 +59,83 @@ You have to work on the `src/movies.js` file.
 
 Ohh yes! We have our beloved tests, and you already know how this works. Open the `SpecRunner.html` file on your browser and start coding to pass the test. Remember to focus on one test at a time and read really carefully the instructions to understand what you have to do.
 
-### Iteration 1: Time Format
 
-We get the info from the **IMDB** web page, but the duration info was saved in a format that difficult us a lot to compare movies.
+### Iteration 1: Get summary
 
-Finding the longest movie is almost impossible using that format, so let's change it!
+We get the info from the **IMDB** web page, but we want to simplify the content to a simple string. 
 
-**:bulb: Remember we have a methods o create a new array from another one!**
+Create a `getSummary` function that receives an array as a parameter and returns a string with the following pattern
 
-- Create a `turnHoursToMinutes` method that receive an array as parameter, replace the duration info of each of the movies for it equivalent in minutes. For example:
+```
+THE_TITLE by DIRECTOR (YEAR)
+```
 
-```javascript=
+For example, the following object...
+```javascript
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -125,16 +119,61 @@ If two movies have the same duration, order them in alphabeticall order by their
...
-Go ahead and create a `howManyMovies` method that receives an array as a parameter and `filter` :eyes: the array so we can have only the **drama** movies where **Steven Spielberg** is the director. 
+Go ahead and create a `howManyMovies` function that receives an array as a parameter and `filter` :eyes: the array so we can have only the **drama** movies where **Steven Spielberg** is the director. 
 
 
 ### Iteration 6: Alphabetic Order
 
 Another famous way to order the movies is to sort them alphabetically using the `title` key. But in this case we only need to print the title of the first 20. Easy Peasy for an expert like you :wink:
 
-Create a `orderAlphabetically` method, that receive an array and return an array of first 20 titles, alphabetically ordered. Return only the title of each movie, and if the array you receive have less than 20 movies, return all of them order
+Create a `orderAlphabetically` function, that receives an array and return an array of first 20 titles, alphabetically ordered. Return only the title of each movie, and if the array you receive have less than 20 movies, return all of them order
```
sandrabosk: The sentence wasn't finished and it had minor grammar issue. Please update your file and push the updates.
Suggested update:

Create an `orderAlphabetically` function, that receives an array and returns an array of first 20 titles, alphabetically ordered. Return only the title of each movie, and if the array you receive has less than 20 movies, return all of them ordered alphabetically of course 😉

README.md
```diff
@@ -80,24 +82,16 @@ Finding the longest movie is almost impossible using that format, so let's chang
...
-  "rate":"9.3"
-}
+...would be turned into:
+```javascript
+"The Shawshank Redemption by Frank Darabont (1994)"
 ```
- 
-You must return a new array with all the info about the movie, not modify the original array!
+
+**:bulb: Remember we have a methods to create a new array from another one! This method is `map`.**
```
sandrabosk: Suggested update:

**:bulb: Remember we have methods to create a new array without making changes to the original array! One of these methods is `map`.**

PR #294, opened 2018-10-11
BER - Liang & Katrin

KatrinAbel:


Noelialeon:
Well done!! A small tip: in the turn hoursToMinutes function, you could have done:

 function turnHoursToMinutes(array){
  var newarray = array.map(function(element){
// We modify the element.duration:
          element.duration= parseInt(element.duration.substring(0))* 60 + parseInt(element.duration.substring(3,5))
// We return the full element that will contain the new .duration
      return element
  })
  return newarray;
}


This way we save some lines of code ;) Let me know if you have any question

PR #298, opened 2018-10-11
BER Khalil Matthias

MattStaudinger:
Added Bonus iteration.

### DIFF
--- a/starter_code/src/data.js
+++ b/starter_code/src/data.js
@@ -1,2002 +1,2003 @@
 var movies = [
   {
-    title: 'The Shawshank Redemption',
-    year: '1994',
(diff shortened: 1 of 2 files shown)


Noelialeon:
Well done!! A small tip: in the turn hoursToMinutes function, you could have done:

function turnHoursToMinutes(moviesArray) {
  if (moviesArray.length === 0) return undefined;
  var minutesConverted = moviesArray.map(movie => {
// We modify the element.duration:
      movie.duration= convertToMinutes(movie.duration)   
// We return the full element that will contain the new .duration
    return movie;
  });
  return minutesConverted;
}

This way we save some lines of code ;)  Let me know if you have any question!

PR #304, opened 2018-10-18
[SAO FT1 - PP - W1D4] Jeff Smith & Gabriel Sicuto

jmstranslate:
I think we got to get better. It's difficult.


thebinaryfelix:
That was a difficult one! Watch it for the console.log's in there for testing, before push it to github, refactor the code.

PR #312, opened 2018-10-20
Berlin-Lara-Petersen

withenoto:


HendrikW:
see comment(s) here: https://github.com/ironhack-labs/lab-javascript-all-times-movies/pull/320

PR #314, opened 2018-10-20
Berlin-Luise-Schneider

luisesch:


starter_code/src/movies.js
```diff
@@ -1,20 +1,59 @@
...
+  var result = array.map(function(movie) {
+    var obj = {
+      title: movie.title,
+      year: movie.year,
+      director: movie.director,
+      duration: stringToMinutes(movie.duration),
+      genre: movie.genre,
+      rate: movie.rate
+    };
+    newArray.push(obj);
```
HendrikW: you should simply return `result` from the `turnHoursToMinutes` function, no need to push to a `newArray` :)

starter_code/src/movies.js
```diff
@@ -1,20 +1,59 @@
...
+  });
+  return newArray;
+}
+
+// Get the average of all rates with 2 decimals
+function ratesAverage(arr) {
```
HendrikW: this looks good !

starter_code/src/movies.js
```diff
@@ -1,20 +1,59 @@
...
+  var avg = avgSum / arr.length;
+  return Math.round(avg * 100) / 100;
+}
 // Get the average of Drama Movies
-
+function dramaMoviesRate(arr) {
+  var dramaFilter = arr.filter(function(movie) {
+    return movie.genre.includes("Drama");
+  });
+  var avgDrama = dramaFilter.reduce(function(sum, movie) {
```
HendrikW: this also looks good (although you could of course re-use the `ratesAverage` function that you already defined above 👍

PR #315, opened 2018-10-20
Berlin - Nike Miche-Soth

nikemiso:


starter_code/src/movies.js
```diff
@@ -1,13 +1,40 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
-
+console.log(movies[0])
+
+    function stringToMinutes(string) {
```
HendrikW: plase copy&paste my corrected (and shorter) version of this function from slack :)

starter_code/src/movies.js
```diff
@@ -1,13 +1,40 @@
...
+          minutes += Number(string[3] + string[4]);
+          }
+    
+          return minutes;
+    }
 
 // Get the average of all rates with 2 decimals 
 
 
+function ratesAverage(arr) {
```
HendrikW: this looks good, should be working I think 👍

PR #318, opened 2018-10-20
BER-Axel-Heinz-movies

axelheinz:


starter_code/src/movies.js
```diff
@@ -1,9 +1,36 @@
...
+    movie.duration=stringToMinutes(movie.duration);
+  } 
+  )};
+
 
 // Get the average of all rates with 2 decimals 
 
+function sumRates(movies){
+  return movies.reduce(function(sum,rating){
+    sum=sum+parseFloat(movies.rate);
```
HendrikW: the anonymous function that is passed to `reduce` needs to return something (`return` statement is missing)

starter_code/src/movies.js
```diff
@@ -1,9 +1,36 @@
...
+  }
+)};
+
+function ratesAverage(movies){
+  return movies.reduce(function(sum,rating){
+    avg=sumRates/movies.length;
```
HendrikW: here as well the `return` statement is missing

PR #322, opened 2018-10-20
Berlin- Susa - first try

su-jaya:


starter_code/src/movies.js
```diff
@@ -2,12 +2,73 @@
...
+
+}
 
 // Get the average of Drama Movies
 
 
+function dramaMoviesRate (array) {
+
+var filter = array.filter(function(first){
+ return first.genre.includes("Drama")
```
HendrikW: maybe `first` is not a good name for the parameter, since it does not only filter the first ;) `element` or `movie` would be better I think

PR #323, opened 2018-10-20
Berlin -Nir

nirtzur89:
I dont like it

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -26,10 +26,29 @@ function ratesAverage (moviesarr){
 }
 
 // Get the average of Drama Movies
+function dramaMoviesRate (moviesArr){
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,9 +1,29 @@
...
+function turnHoursToMinutes(moviesarr){
+    var newTime = moviesarr.map(function(movie){
+        return movie.duration= stringToMinutes;
+    });
+}
 
 
-// Get the average of all rates with 2 decimals 
 
+// Get the average of all rates with 2 decimals
```
HendrikW: this looks good already 👍

nirtzur89: Well according to Jasmine I've done everything wrong so IDK.
made a new push...

PR #324, opened 2018-10-20
Berlin - Camille Chapleau

camchpl:


starter_code/src/movies.js
```diff
@@ -1,20 +1,56 @@
...
+    return hours*60+minutes
+}
+function turnHourstoMinutes(movies){
+    return movies.map(function(){
+        movie.duration=stringToMinutes(movie.duration)
+    });
+}
 
 // Get the average of all rates with 2 decimals 
+function ratesAverage (movies){
```
HendrikW: this function needs to return something ( `return` statement is missing :) )

PR #325, opened 2018-10-20
not finished

juanibe:


starter_code/src/movies.js
```diff
@@ -1,12 +1,45 @@
...
+    var count = array.length
+
+    return sumRates / count
+}
+
 // Get the average of Drama Movies
 
+function dramaMoviesRate(moviesArr){
+    var drama = moviesArr.filter(function(movie){
+        return movie.genre === "drama";
```
HendrikW: if you look at the data structure, it needs to be `movie.genre.includes('drama')`

PR #326, opened 2018-10-22
Berlin-Steph-Gidde-(done 1 - 4)

StephGidde:


starter_code/src/data.js
```diff
@@ -2000,3 +2000,71 @@ var movies = [
...
+
+//Average Rating
+var movieRates = movies.map(function(element) {
+  return parseFloat(element.rate);
+})
+
+console.log(movieRates)
+
+var ratesSum = movieRates.reduce(function(sum, movieRates){
+  return sum += movieRates;
```
HendrikW: `return sum + movieRates` works, `+=` not needed :)

starter_code/src/data.js
```diff
@@ -2000,3 +2000,71 @@ var movies = [
...
+
+console.log(ratesSum)
+
+var ratesAverage = ratesSum / movieRates.length;
+console.log(ratesAverage)
+
+//Drama Movies
+
+var dramaMovies= movies.filter(function(elemt){
+  if(elemt.genre.includes('Drama') === true){
```
HendrikW: actually, the function passed to `filter` can simply return true or false, so `return elemt.genre.includes('Drama')` would be enough

StephGidde:
Thank you for the feedback! It doesn't pass the tests, that's true. Everything worked in repl.it, but yes, Jasmine didn't like it.

PR #328, opened 2018-10-23
Paris - Corrado

griev04:


ta-web-paris:
Great job ! 
Here is an exemple of the bonus if you want to look into that part : https://github.com/marieglr/lab-javascript-all-times-movies-solutions

PR #333, opened 2018-10-23
Paris Fareaha

nrlfrh:
Almost done left with orderby duration.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -65,7 +65,7 @@ dramaMoviesRate(movies);
 
 // Order by time duration, in growing order
 
-function orderByYear(array) {
+function orderByDuration(array) {
   return array.sort(function(a, b) {
     if (Number(a.year) < Number(b.year)) {
       return -1;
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,112 @@
...
+function orderByDuration(array) {
+  return array.sort(function(a, b) {
+    if (Number(a.year) < Number(b.year)) {
+      return -1;
+    } else {
+      return 1;
+    }
+  });
+}
+
```
ta-web-paris: Since you turn years into numbers, it would be simpler to put     Number(a.year) - Number(b.year)  in the call back of your sort method, the way you did it is more used for strings. To compare movies based on their title if they have the same duration, you can jus add a condition : 
```js
  moviesArray.sort(function(a, b) {
      if (a.title > b.title) {
        return 1;
      } else {
        return -1;
      }
    }
  });
  return moviesArray;
}
```

ta-web-paris:
Good job !!! You can check full solutions here if you want to compare : https://github.com/marieglr/lab-javascript-all-times-movies-solutions

PR #335, opened 2018-10-24
 -Paris  Niccolò

nicrodemus:
i miss a couple of passages

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,103 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
+// order by duration
+function orderByDuration (movies) {
+    let longest = movies.sort(function(movieA,movieB){
+           if(movieA.duration < movieB.duration){
+             return -1
+           } else  {
+               
+           
+             return +1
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,103 @@
...
+           if(movieA.duration < movieB.duration){
+             return -1
+           } else  {
+               
+           
+             return +1
+           }
+         });return longest
+        }
+     longest(movies)
```
ta-web-paris: To order them by their title if they have the same length, you just have to add some if statements :) 
Also, since the sort method modifies the original array, you don't have to create a new variable, but it is okay to do it! 
```js
function orderByDuration(movies) {
  let longest = movies.sort(function(movieA, movieB) {
    if (movieA.duration === movieB.duration) {
      if (movieA.title < movieB.title) {
        return -1;
      } else {
        return +1;
      }
    } else if (movieA.duration < movieB.duration) {
      return -1;
    } else {
      return +1;
    }
  });
  return longest;
}
longest(movies);
```

ta-web-paris:
Good job Niccolo ! Keep up the good work !

PR #341, opened 2018-10-24
Paris - Harnit

Harnitks:
I didn't have complete the exercice :(

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -4,17 +4,104 @@
 
 // Get the average of all rates with 2 decimals 
 
+function ratesAverage(movies){
+    var ratesAverage = movies.reduce(function(sum,oneMovies){
+    
+      return sum + Number(oneMovies.rate);
+    
+    },0);
+    return ratesAverage/movies.length
+    
+    };
+
 
 // Get the average of Drama Movies
+
+
+// function drama (movies){
+//     movies.filter (function (oneMovies){
+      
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -4,17 +4,104 @@
...
+          return -1;
+        }
+        else{
+          //if itemB comes before itemA return positive
+          //they need to switch
+          return 50;
+        }
+      })
+      
+      console.log(movies);
```
ta-web-paris: Instead of console.logging the movies after sorting the array, you should return it! Also for this questions, we had to make an array of only titles, the best way to do that is with a map. And finally, we can slice the fist 20 ones :) 
```js
function orderAlphabetically(thisArrayIsMadeOfMoviesMyFriend) {
  //Get an array with only the movie titles
  var movieTitles = thisArrayIsMadeOfMoviesMyFriend.map(function(oneMovie) {
    return oneMovie.title;
  });

  //Sort the movie titles in alphabetical order
  movieTitles.sort(function(titleA, titleB) {
    titleA.toUpperCase(); // ignore upper and lowercase
    titleB.toUpperCase(); // ignore upper and lowercase
    if (titleA < titleB) {
      return -1;
    }
    if (titleA > titleB) {
      return 1;
    }
  });

  //keep only the first 20 ones
  return movieTitles.slice(0, 20);
}
```

ta-web-paris:
It is okay not to finish the labs! You did a good job !!!

PR #342, opened 2018-10-24
Paris - Adèle

AdeleRevert:
Not finished at all!
For the howManyMovies we searched with Abi why Jasmine doesn't print the last request in green, we couldn't figure out why (my code prints the exact requested string + we made it with Harnit and it works for her with the exact same code)

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -3,10 +3,49 @@
 
 
 // Get the average of all rates with 2 decimals 
-
+function ratesAverage(){
+  var rate = [];
+ movies.forEach(function (oneMovie){
+  rate.push(oneMovie.rate);
+ });
+ 
+ var total =
+ rate.reduce(function (sum, oneRate){
+   return sum + parseFloat(oneRate);
+ }, 0);
+ 
+ var average = Number((total / 250).toFixed(2));
+ 
+ return average;
+ };
+ 
+ ratesAverage(movies);
+ 
 
 // Get the average of Drama Movies
+var drama = 
+  movies.filter(function (movies) {
+  return movies.genre.includes('Drama');
+  });
 
+function dramaMoviesRate(){
+var dramaRate = [];
+drama.forEach(function (oneMovie){
+ dramaRate.push(oneMovie.rate);
+});
+
+var totalDrama =
+dramaRate.reduce(function (sum, oneRate){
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -3,10 +3,49 @@
...
+var totalDrama =
+dramaRate.reduce(function (sum, oneRate){
+  return sum + parseFloat(oneRate);
+}, 0);
+
+var averageDrama = Number((totalDrama / drama.length).toFixed(2));
 
+return averageDrama;
+
+};
```
ta-web-paris: Same thing here : 
```js

function dramaMoviesRate(aMoviesArray) {
  
  var drama = aMoviesArray.filter(function(movies) {
    return movies.genre.includes("Drama");
  });

  var dramaRate = [];
  drama.forEach(function(oneMovie) {
        //This if statement is to please Jasmine:
    // ('Return Average even if one of the movies does not have rate!')
    if (oneMovie.rate === "") {
      oneMovie.rate = 0;
    }
    dramaRate.push(oneMovie.rate);
  });

  var totalDrama = dramaRate.reduce(function(sum, oneRate) {
    return sum + parseFloat(oneRate);
  }, 0);

  var averageDrama = Number((totalDrama / drama.length).toFixed(2));

  return averageDrama;
}
```

starter_code/src/movies.js
```diff
@@ -15,6 +54,49 @@
...
+console.log(moviesTitle);
+
+var first20 = [];
+function orderAlphabetically(oneMovie){
+  oneMovie.forEach(function (oneMovie){
+  first20.push(oneMovie.title);
+});
+  console.log(first20);
+};
+
```
ta-web-paris: Again, you should put every thing in one generic version. Also, if you want the sort method to sort thing in the opposite order you just have to switch 1 and -1
Here is the most simple way to do it : 
```js
  moviesArray.sort(function(a, b) {
      if (a.title > b.title) {
        return 1;
      } else return -1;
    }
  });
  return moviesArray;
}
```

ta-web-paris:
This iteration is all green on mine !
Here is the complete solution if you want to checkout : https://github.com/marieglr/lab-javascript-all-times-movies-solutions/blob/master/starter_code/src/movies.js
Don't hesitate to come ask questions if you still have some :)

PR #353, opened 2018-10-25
AMS jesco wuester

jescowuester:
…utput to match theirs.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,99 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
-
+function hourtominute(time){
+    let filtered = time.split("").filter(item => ["m","i","n"," ","h"].includes(item) == false)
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,99 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
-
```
IronhackAMS: So your code works if I, for example, explicitly enter a duration of a film. However, I'd like to see the a new array with all the movies and only the duration is changed from hours to minutes.

IronhackAMS:
Don't mind specRunner it is indeed created by the Lucifer himself. Please do mind code formatting. Use indentation properly, it will make your code more readable. Overall very well done!

starter_code/src/movies.js
jescowuester: function turnHoursToMinutes(arr){
    for(i of arr){
      i.duration = hourtominute(i.duration)
    }
    return(arr)
  }

its right below it :)

PR #390, opened 2018-10-26
AMS - David Z -Finish Javascript all-time movies exercise 

zlobizlobi:


starter_code/src/movies.js
```diff
@@ -2,19 +2,82 @@
...
+  let hours = time.slice(0, time.indexOf("h"))
+  
+  let minutes = time.slice(time.indexOf(" ") + 1, time.indexOf("min"))
+
+  minutes = parseInt(minutes)
+  hours = parseInt(hours)
+
+  return minutes + hours * 60 
+}
+
```
IronhackAMS: You could write the code slightly more concise here. E.g.
let newArray = array.map(movie => {
   movie.duration = giveMeHours(movie.duration);
  return movie;
  },0)

IronhackAMS:
Nice work. Please keep proper indentation in mind.

PR #393, opened 2018-10-27
MIA-FTWD ANGEL WU

angelwu007:


MarcosPalaciosIII:
Great work on the all time movies lab. There will be many times in your career that you will have to do something similar to this. Having a good understanding of how to get needed data will help you later when you have to do something similar.

Keep up the great work.

PR #394, opened 2018-10-28
MIA FTWD Rachel Hofer - JS All Times Movies v1, Incomplete

Rachel-Hofer:
Still working on this.


MarcosPalaciosIII:
Good work on the all time movies lab. There will be many times in your career that you will have to do something similar to this. Having a good understanding of how to get needed data will help you later when you have to do something similar.

Keep up the good work.

PR #401, opened 2018-10-31
AMS JUDE

jdhova:
updated

### DIFF
--- a/starter_code/src/data.js
+++ b/starter_code/src/data.js
@@ -2054,6 +2054,61 @@ function giveMeHours(timeString) {
 giveMeHours(time)
 movies[0].duration = giveMeHours(time)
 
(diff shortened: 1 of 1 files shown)


starter_code/src/data.js
```diff
@@ -2000,3 +2001,60 @@ var movies = [
...
+
+  var minutes = timeString.slice(timeString.indexOf(" ") + 1, timeString.indexOf("min"))
+
+  minutes = parseInt(minutes)
+  hours = parseInt(hours)
+
+  return (minutes/ 60) + hours
+}
+giveMeHours(time)
+movies[0].duration = giveMeHours(time)
```
IronhackAMS: The return value here is 2.55, however I would expect it to be 155.

jdhova:
updated

PR #442, opened 2018-11-14
initial commit

lwilhelmi:


IronhackAMS:
Looks good man! You could use .map() in order to return an array with all the movies and the converted hours to minutes.

PR #465, opened 2019-01-10
Paris - Kevin M & Céline

frietpoune:
We did Iteration 2 and 3, will work on the others later on :)

@KevGhub

(Jasmine is sometimes very annoying haha but still useful)

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,38 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
+// Turn duration of the movies from hours to minutes
 
+// Get the average of all rates with 2 decimals
 
-// Get the average of all rates with 2 decimals 
+function ratesAverage(movies) {
+  var average = movies.reduce(function(sum, oneMovie) {
+    return sum + Number(oneMovie.rate);
+  }, 0);
+  return Number((average / movies.length).toFixed(2));
+}
 
+console.log(ratesAverage(movies));
 
 // Get the average of Drama Movies
 
+function dramaMoviesRate(movies) {
+  var dramaMovies = movies.filter(function(oneMovie) {
+    return oneMovie.genre.indexOf("Drama") > -1;
+  });
+  console.log(dramaMovies);
+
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,118 @@
...
+
+console.log(dramaMoviesRate(movies));
 
 // Order by time duration, in growing order
+// Will do when Iteration 1 is solved
 
+// Order by date
+//Cannot solve the 2 Jasmine errors :
+// - Return the new array in growing order
+// -If two movies have the same length, order them alphabetically by their title
```
ta-web-paris: the reason your tests are not passing is because you're ordering by year (Nizar said to do that)  rather than duration. But your code works :)

starter_code/src/movies.js
```diff
@@ -1,20 +1,118 @@
...
+    "Steven Spielberg directed " + moviesSpielberg.length + " drama movies!"
+  );
+}
 
 // Order by title and print the first 20 titles
+//Cannot solve the 2 Jasmine errors :
+// - You should order them alphabetically.
+// - You should return the top20 after order them alphabetically.
+
+function orderAlphabetically(movies) {
```
ta-web-paris: You've made this a bit more complicated than you needed to.

In your first sort you should check for the title not the rate 

You only need to sort once. Once you have a sorted array, you can slice that if it's more than 20 or return the entire array if it's below 20. 

```
  var orderedRate = movies.sort(function(itemA, itemB) {
    if (itemA.title < itemB.title) {
      return -42;
    } else {
      return 89;
    }
  });
  var top20Movies = [];
  var limit = 20;
  if (orderedRate.length < 20) {
    limit = orderedRate.length;
  }
  for (var i = 0; i < limit; i++) {
    top20Movies.push(orderedRate[i].title);
  }

  return top20Movies;
}
```

if this doesn't make sense, let me know and I can explain :)

ta-web-paris:
You seem to have understood what we covered in class! Your code is good and you managed to complete a lot of the lab. 

If you work on the last function over the weekend, let us know on Monday and we can review it with you :)

PR #482, opened 2019-01-11
PARIS - NADJIE & ESTELLE

Olawinski-e:
@NadjieVal


ta-web-paris:
You did well! You managed to use some of the new array methods we covered in class and passed the tests 👏

If you want to keep working on it over the weekend, we can give you some more feedback on Monday :)

PR #488, opened 2019-01-11
Paris - Christophe & Aurelie

grosbegue:
@AurelieBeno

Exercice non terminé, il manque quelques tests et sections


starter_code/src/movies.js
```diff
@@ -1,20 +1,71 @@
...
 
 // Get the average of Drama Movies
 
+var dramaMovies = movies.filter(function(oneMovie) {
+  if (oneMovie.genre.indexOf("Drama") >= 0) {
+    return oneMovie;
+  }
+});
+
+function dramaMoviesRate(array) {
```
ta-web-paris: To pass the final test for this function you need to add 

``` 
 if (dramaMovies.length === 0) {
    return undefined;
  }
```

This checks to see if there are any drama movies, and if not, returns undefined

starter_code/src/movies.js
```diff
@@ -1,20 +1,71 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
+// Turn duration of the movies from hours to minutes
 
+// Get the average of all rates with 2 decimals
 
-// Get the average of all rates with 2 decimals 
+function ratesAverage(array) {
+  var averageRate = array.reduce(function(sum, oneMovie) {
+    var totalRate = sum + parseInt(oneMovie.rate);
```
ta-web-paris: you can  change parseInt here to Number in order to pass one of the tests for the next iteration 

``` var totalRate = sum + Number(oneMovie.rate);```

There is a small [difference](https://stackoverflow.com/questions/4090518/what-is-the-difference-between-parseint-and-number) between them

ta-web-paris:
it looks good! 👏

if you want to keep working on it over the weekend, let us know on Monday and we can help or give you more feedback on it :)

PR #491, opened 2019-01-11
done but not entierly. Will continue this weekend

F-Jules:
@wowfranzik

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,63 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
+// Turn duration of the movies from hours to minutes
 
+// Get the average of all rates with 2 decimals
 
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,84 @@
...
+  }
+}
+
+console.log(dramaMoviesRate(movies));
 
 // Order by time duration, in growing order
 
+function orderByDuration(movies) {
+  var sortedMovies = movies.sort(function(itemA, itemB) {
+    if (itemA.year > itemB) {
```
ta-web-paris: You also need to compare the year of itemB:  `(itemA.year > itemB.year)`

ta-web-paris:
What you've done is good!  If you want more feedback on it, let us know on Monday and we can review it for you 
💪

Please don't forget to put Paris in the title, without Paris we might not see it 😢

PR #494, opened 2019-01-11
Paris - Alexandre & Delphine

Deie:
Iteration 2 done

@alexjoly13

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -4,9 +4,27 @@
 
 // Get the average of all rates with 2 decimals 
 
+function ratesAverage(moviesArray) {
+
+    moviesRate = moviesArray.map(function (oneMovie) {
+        return parseFloat(oneMovie.rate);
+    });
+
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -4,9 +4,27 @@
...
 
+function ratesAverage(moviesArray) {
 
-// Get the average of Drama Movies
+    moviesRate = moviesArray.map(function (oneMovie) {
+        return parseFloat(oneMovie.rate);
+    });
+
+    rateTotal = moviesRate.reduce(function (sum, oneRate) {
+        return sum + oneRate;
```
ta-web-paris: For Number to work and pass the tests, you need to add an extra set of parentheses so that it includes the .toFixed()

` return Number((rateTotal / moviesArray.length).toFixed(2));`

ta-web-paris:
It's a good start! I'd recommend trying to do more over the weekend because these methods are going to come up a lot. If you do anymore, do another commit and let us know so we can give you more feedback on it :)

PR #547, opened 2019-02-04
[AMS] - Fedde

Fedduh:
Sorting is still messed up. Each sorting method changes the original movies-data, so now everything is sorted by year :)

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,135 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
 
+// -------------------------------------------- //
+// Turn duration of the movies from hours to minutes  
+// -------------------------------------------- // 
+const setDurationInMinutes = arrayInput => (
+  arrayInput.map(element => {
+    let hour = parseInt(element.duration.slice(0, element.duration.indexOf("h")) * 60) || 0;
+    let min = parseInt(element.duration.slice(element.duration.indexOf("h") + 2, element.duration.indexOf("min"))) || 0;
+    element.duration = hour + min;
+    return element;
+  })
+)
 
+console.log(setDurationInMinutes(movies));
+
+// Check if some have a missing duration now
+const hoursCheck = movies.filter(movie => movie.duration == NaN);
+console.log(hoursCheck);
+
+// -------------------------------------------- //
 // Get the average of all rates with 2 decimals 
+// -------------------------------------------- //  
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,135 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
 
+// -------------------------------------------- //
+// Turn duration of the movies from hours to minutes  
+// -------------------------------------------- // 
+const setDurationInMinutes = arrayInput => (
+  arrayInput.map(element => {
+    let hour = parseInt(element.duration.slice(0, element.duration.indexOf("h")) * 60) || 0;
```
TA-Ironhack-Matt: Your code seems fine, but why did you use the (OR) '||' operator at the end of this line? As far as I can see this is not needed. 

What you could have done as error handling is to use a ternary operator that returns 0 or any other value in case of an 'error' or 'undefined'.

starter_code/src/movies.js
```diff
@@ -1,20 +1,135 @@
...
+    let min = parseInt(element.duration.slice(element.duration.indexOf("h") + 2, element.duration.indexOf("min"))) || 0;
+    element.duration = hour + min;
+    return element;
+  })
+)
 
+console.log(setDurationInMinutes(movies));
+
+// Check if some have a missing duration now
```
TA-Ironhack-Matt: Very good to check your result in this way! 👍

starter_code/src/movies.js
```diff
@@ -1,20 +1,135 @@
...
+let genre = "Drama";
+console.log("Average rating " + genre + " is: " + calcAvgRate(filterGenre(movies, genre)));
 
+// -------------------------------------------- //
 // Order by time duration, in growing order
+// -------------------------------------------- //
+// this order function is screwed up by the sorting later on.  so now it's ordered by year
+// I tried to make a copy of the movies dataset, but I did not get that working... 
 
+const sortByDurationTitle = arrayInput => (
```
TA-Ironhack-Matt: Let's look at this function together next Saturday.

starter_code/src/movies.js
Fedduh: Not every movie contained hours, some only contained minutes. 
If the parseInt(...) returns undefined (false) because "h" was not found it should return 0. 

In my testing this seemed to work. For me it's seems rather similar to the ternary operator

PR #548, opened 2019-02-04
bonus to do

debasrid:
Assignment without bonus


starter_code/src/movies.js
```diff
@@ -1,20 +1,86 @@
...
+ })
+ console.log(movieBySteven);
+
+ let dramaBySteven = movieBySteven.filter(movie => {
+    return movie.genre.includes("Drama")
+ })
+
+    console.log(dramaBySteven.length);
 
 // Order by title and print the first 20 titles
```
TA-Ironhack-Matt: The exercise was to sort the movies based on the duration. Try to modify your code so that the sortation is based on the duration and not the title.

PR #549, opened 2019-02-05
[PAR PT] Nouri

zorisse:


Paul-de-Barbeyrac:
Well done, try finishing and fixing the items not passing yet !

Few remarks : 
- Good job using console.log very often. It will help you a lot later on to debug
- Math.round(avg * 100) /100 is a nice way to get 2 decimals. One other handy way is to use : avg.toFixed(2)
- For the assignement : "Get the average of Drama Movies". You did a good job filter for drama genre but you could call the function average you just defined above instead of rewriting its statements. Check out the DRY Principle in programmation (Don't Repeat Yourself)
- Regarding string to avoid concatenation you can use template strings which are useful
Your code : return "Steven Spielberg directed " +arr.length +" drama movies!"
Becomes return `Steven Spielberg directed ${arr.length} drama movies!`

PR #550, opened 2019-02-07
dani demarchi - movies js

danidimarti:
Hi guys,

here we go. 

please disregard bonus exercise. 

Many thanks

Kind regards, 
Dani

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,87 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
+let turnHoursToMinutes = movies.map(movie => {
+    //apply the following function "movie" to each element of my array
+    var x = movie.duration.charAt(0) * 60 + parseInt(movie.duration.slice(3, 5));
+    //parseInt changes strings into integers
+    //movie.duration = get the property duration of the object movie in my array
+    movie.duration = x //now the property duration is equal to the equation created
+    return movie; //movie is each element of my array, with all it's properties
+})
+
+console.log(turnHoursToMinutes);
 
 
 // Get the average of all rates with 2 decimals 
+let ratesAverage = movies.reduce((accumulator, movie) => {
+    return accumulator + parseInt(movie.rate);
+}, 0)
+
+let averageRate = ratesAverage / movies.length
+console.log(averageRate)
 
 
 // Get the average of Drama Movies
+var count = 0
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,87 @@
 /* eslint no-restricted-globals: 'off' */
 // Turn duration of the movies from hours to minutes 
+let turnHoursToMinutes = movies.map(movie => {
```
TA-Ironhack-Matt: This function is causing an error on my machine. Did you run this function successfully?

starter_code/src/movies.js
```diff
@@ -1,20 +1,87 @@
...
+    movie.duration = x //now the property duration is equal to the equation created
+    return movie; //movie is each element of my array, with all it's properties
+})
+
+console.log(turnHoursToMinutes);
 
 
 // Get the average of all rates with 2 decimals 
+let ratesAverage = movies.reduce((accumulator, movie) => {
+    return accumulator + parseInt(movie.rate);
```
TA-Ironhack-Matt: Your logic is good, but because you use parseInt() your result is 8.016 which is not correct. 

When you change parseInt() to parseFloat() the result will be correct. 

The following stack overflow page describes the difference between both methods clearly > https://stackoverflow.com/questions/12812863/difference-between-parseint-and-parsefloat

starter_code/src/movies.js
```diff
@@ -1,20 +1,87 @@
...
+console.log(averageRate)
 
 
 // Get the average of Drama Movies
+var count = 0
+let dramaMovieRate = movies.reduce((accumulator, movie) => {
+    if (movie.genre.includes("Drama")) {
+        count++;
+
+        return accumulator + parseInt(movie.rate);
```
TA-Ironhack-Matt: Logic is good, but the same problem as on line 17

starter_code/src/movies.js
danidimarti: It was running before, but now I tested it again and I'm seeing a reference error on movies. Not quite sure why, but isn't the movies.js suppose to be taking the data from data.js? So why is giving a reference error on movies.map?

starter_code/src/movies.js
danidimarti: done!

starter_code/src/movies.js
danidimarti: done!

PR #596, opened 2019-03-22
PAR mathis + alex

alexandfox:
submitting for both of us!


ta-web-paris:
Well done for the pair programming and the fact that you worked on the same folder ! Everything is good. We will give you the correction for the bonus later on !

PR #600, opened 2019-03-25
PARIS - Paul Lacour

paullacour:
lab all time movies done. Except for bonus iteration.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,20 +1,114 @@
 /* eslint no-restricted-globals: 'off' */
-// Turn duration of the movies from hours to minutes 
 
+// Turn duration of the movies from hours to minutes
 
-// Get the average of all rates with 2 decimals 
+function turnHoursToMinutes(arr) {
+  var newMoviesArray = arr.map(movie => {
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,20 +1,114 @@
...
+    } else if (isSpielberg.length===0){
+      return "Steven Spielberg directed 0 drama movies!";
+    } else if (isSpielberg.length===1){
+      return "Steven Spielberg directed 1 drama movies!";
+    } else if (isSpielberg.length===2){
+      return "Steven Spielberg directed 2 drama movies!";
+    } else if (isSpielberg.length===4){
+      return "Steven Spielberg directed 4 drama movies!";
+    } else {
+      return isSpielberg.length.toString();
```
ta-web-paris: This is kinda "hard coded" so not really efficient.

your answer should be something like this:
return `Steven Spielberg directed ${newArray.length} drama movies!`;

ta-web-paris:
Good for most of the case. Correction for the bonus and the rest will come. Good job anyway !

PR #603, opened 2019-03-25
PAR - Susan Almassi

sioucre:


ta-web-paris:
Very good !
Full correction later on and reinforcement lesson on this lab tonight if you want. But so far it is good !

PR #608, opened 2019-03-25
Paris Yagmur

YagmurBakici:


starter_code/src/movies.js
```diff
@@ -1,20 +1,132 @@
...
-// Order by title and print the first 20 titles
+  // console.log("all drama movies are : " + dramaMovies);
 
+  dramaMovies.forEach(dramaMovie => {
+    dtotal += Number(dramaMovie.rate);
+  });
+  // console.log("total of all drama movies: " + dtotal);
 
-// Best yearly rate average
+  return dtotal / dramaMovies.length;
```
ta-web-paris: Check .toFixed() method when you are asked to return "n" number after the dot of a float number

ta-web-paris:
Some small errors but it is good ! Correction in comming and tonigh we review the lab after class !

PR #625, opened 2019-04-04
MIA FTWD - Nour Cherif Essoussi - April 2019

Nourcs:


ta-web-mia:
Good job! Really nice use of ES6 syntax and array methods. There isn't too much to improve on here, but never stop practicing your vanilla JS and stay sharp. Keep it up!

PR #626, opened 2019-04-04
MIA FTWD - Angel Biondi

AngelBiondi:
"done"


ta-web-mia:
Good job! Really nice use of the spread operator and ES6 arrow functions. Glad to see you've become more comfortable with the more complicated syntax, keep it up!

PR #627, opened 2019-04-04
MIA - FTWD - APR - Nathaniel Carvajal

nathancarv:


ta-web-mia:
Nice job with the first function, it's definitely one of the hardest on here. Although it might be too late to get back to this lab and finish the later activities, I suggest you look into sorting methods and finding the average of a number array for practice. They're very common coding tests for entry-level developers, so you'll definitely be thankful that you looked at them. Keep it up!

PR #628, opened 2019-04-04
MIA-FTWB-April Parker Cooper

parkcoop:
Missing bonus iteration


ta-web-mia:
Nice job! Nice use of array methods; you're not going to stop using them at any point, so getting comfortable with them now is a huge benefit. Keep it up!

PR #629, opened 2019-04-04
MIA-FTWD-APR-Justin Velez

jvelez523:


ta-web-mia:
Good job! There's one missing edge case for the drama movie function, which is just meant to return undefined if the length of the array is 0. Other than that and the bonus iteration, everything else looks perfect. Keep up the good work!

PR #630, opened 2019-04-04
MIA FTWD April Kenneth Solano

klsolano7:


ta-web-mia:
Good job! Sorting the movies by duration can be pretty tricky, so don't get too hung up on it. Generally speaking though, the first function you declared that converted the duration to minutes will help you a lot in comparing them all. Keep up the good work!

PR #631, opened 2019-04-05
MIA FTWD - Christian Khawam

CodeByCK:


ta-web-mia:
Good job! You missed one small edge case to check if the length of the drama movie array was 0, but that's as simple as one small conditional. Not a big deal at all! Other than that, everything looks good. Keep it up!

PR #632, opened 2019-04-05
MIA FTWB ANGHY POMAR

apomarn:


ta-web-mia:
Awesome! Really nice job on the bonus iteration as well. You're really improving your JavaScript skills, and I think the next step is to start using newer ES6 syntax. Keep up the good work!

PR #633, opened 2019-04-05
MIAFTWD - April - Chris C

ccardelle:


ta-web-mia:
Nice job! I like your use of the spread operator on the return statement in the turnHoursToMinutes function, as well as calling it within the orderByDuration one. You were on the right track using a sorting method in the alphabetical function. Array sorting algorithms are a very common challenge for entry-level devs, so I'm glad to see you're getting more comfortable with them. Keep up the good work!

PR #634, opened 2019-04-05
[MIA FTWD] (Jon R) 

jonriquer:


ta-web-mia:
You're on the right track! For the first function and as a general rule of thumb, make sure you declare a copy array at the start of the function so that you aren't altering the original one. It's typically best practice to not alter the data that you're looking at, in case you need to do something with it later. The main goal of this activity is to get you comfortable using array methods, since they're very commonly used when dealing with backend data in web development. You don't necessarily need to go back to this lab, but definitely practice array sorting algorithms if you ever find yourself with extra time. Keep up the good work!

PR #635, opened 2019-04-10
MIA FTWD APR Daniela Olano

dolano13:


ta-web-mia:
Nice job, you basically got all of the hard ones (except for the bonus iteration). Although the alphabetical sorting algorithm is easier than most of the other ones, it's also a very common coding puzzle used to test entry-level devs. Definitely look into the general idea of the algorithm when you have a chance, although I don't think you need to necessarily revisit this lab.

Also, really like how you're making detailed comments on your code. But as a general rule of thumb, put the comments on their own line instead of next to the code you're executing. All in all, good job with this lab. Keep up the good work!

PR #636, opened 2019-04-10
MIA-FTWD-Allison

allisonkopp:


ta-web-mia:
Nice job, I really like that you're comfortable with arrow function syntax and the spread operator. Everything looks really good, keep it up!

PR #637, opened 2019-04-11
MIA FTWD APR- Katy Brown

katybrown91:


ta-web-mia:
You're definitely on the right track. The first function is one of the hardest in this activity, and you almost got it. The important thing though is that after you remove the "h" and "m" from the duration, you need to convert the number into an integer before you multiply by 60. Otherwise it will think you're multiplying a string by a number, which just won't work. If you have extra time, definitely keep practicing array sorting and looping methods. You're really improving, keep it up!

PR #674, opened 2019-06-09
LIS Mariana 80%

marianamv112:
Dear TA's, I would appreciate you help me with the tests I'm missing. I don't understand why the orderByDuration function is not working and how can I make the orderAlphabetically accept as input an object out of an array (like in the test 'Only return the title of the movies! Each element should be a string', line 194 of MoviesSpecs.js).

Thanks in advance

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -55,10 +55,9 @@ function orderByDuration (array) {
 
     convertedArray.sort( (movie1, movie2) => {
     if (movie1.duration === movie2.duration) {
-        console.log(typeof movie1.duration, typeof movie1.title);
           return movie1.title > movie2.title;
         } else {  
-        movie1.duration > movie2.duration
+        return movie1.duration - movie2.duration;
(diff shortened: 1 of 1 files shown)


carlarsmendes:
Hi Mariana,

So, your logic is all correct, and the result also, so, this is only to do with the way jasmine is doing the tests:
1- She wants you to sort by alphabetical order first, and then select the top 20;
2- She is also constantly looking for a key:value combination, since that is the purpose of this exercise: to use the functions learned and try not to use loops :) 
(there is a moment where she is looking for it and you already have only your array of titles, but she is still looking for a key:value combination)

SO in this case, could you try to use a combination of these:  sort, slice, map..... 
And no loops, and see how it goes?

In fact, all of these exercises are made only to use functional programming so that you start getting used to them

Let me know if you have any additional questions!

Carla

marianamv112:
Corrected. Thanks for the tips, @carlarsmendes !

PR #694, opened 2019-06-13
MIA-FTWD-Sergio/Michael/Nicolle

sergiovelo17:


MarcosPalaciosIII:
Great job on the all time movies lab. The main take away here is that you sometimes will get data that will not be in the format that you would like to use. Also there are many times that the data can be complicated. Learning to filter through the information that you receive and knowing how to manipulate it, is something that you will have to do with a lot of the data that you will get from outside sources(especially api's). 

Keep up the great work.

PR #695, opened 2019-06-13
MIA FTWD - Taluama Velludo and Matiny L.

taluama:


ta-web-mia:
Try to finish the bonus question just for practice, but everything else checks out. Nice job!

PR #696, opened 2019-06-13
MIA-FTWD-Artem/Sams

arstrel:
Main part of assignment complete, 40/45 tests passed


arstrel:
Kudos to Sergio for the help with bonus function. The question still stands why it doesn't work when I call ratesAverage function within it. even though it looks to me that ratesAverages does the same thing

PR #697, opened 2019-06-13
MIA-FTWD Pair Sams Diana, Artem

sdiana11:


ta-web-mia:
Try and do the bonus question just for practice, but otherwise, good job!

PR #699, opened 2019-06-13
MIA-FTWD-Eduardo Gonzalez with Matthew Larrubia

EdoGp:


ta-web-mia:
Here's everything I found: 

Iteration 1 seems fine
Iteration 2 doesn't round to the hundredths place, though iteration 3 does. Use the same formula.
Iteration 3 seems fine
Iteration 4 seems fine
Iteration 5 seems fine
Iteration 6 seems fine
The Bonus works as well.

I didn't test these in Jasmine, so I'm not sure what errors it might be giving you, but they all gave me the results I was looking for when I tested them. The only thing you'll have to change is rounding Iteration 2 to the nearest hundredth. Otherwise it seems good.

PR #703, opened 2019-06-14
MIA-FTWD William Bennett

william1benn:


MarcosPalaciosIII:
Good attempt on the all time movies lab. The main take away here is that you sometimes will get data that will not be in the format that you would like to use. Also there are many times that the data can be complicated. Learning to filter through the information that you receive and knowing how to manipulate it, is something that you will have to do with a lot of the data that you will get from outside sources(especially api's).

It seems like both you and your partner had something completed that the other did not. When working with a partner, assure that you share your code with each other. Two people aiming to complete one goal work better than one trying to it by themselves. The spirit of collaboration means you can work separately, but in the end you must come together.

Also remember to give credit to those you work with (either in title, readme or comments).

Keep up the good work.

PR #710, opened 2019-06-14
FTWD-MIAMI-marioSalvatierra-alfredoRojas

mesalvav:
will add more over the weekend

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2,14 +2,147 @@
 // Turn duration of the movies from hours to minutes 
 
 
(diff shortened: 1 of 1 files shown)


ta-web-mia:
So far so good! The first one is definitely one of the hardest, so definitely be proud that you managed to solve it. Algorithms are admittedly something you won't use often when you're creating actual webpages, but they're extremely helpful in familiarizing yourself with JavaScript methods. Let me know if you have any questions, and keep up the good work 👍 
- Nick

PR #714, opened 2019-06-16
MIA-PTWD-Alexander-Hoinville

ahoin001:


ta-web-mia:
The first iteration is definitely one of the hardest, so good on you for being able to solve it. If you're having problems getting the sum or mean of the ratesAverage() function, feel free to use a for loop instead of an array method. The algorithmic nature of these problems aren't necessarily something that you'll experience very often when coding, but they definitely help reinforce JavaScript skills. Keep up the hard work 👍

PR #719, opened 2019-06-16
MIA PTWD Antquanique Dancy

adancy95:


ta-web-mia:
Nice job, I'd say you got basically all of the hard ones! Sorting methods aren't too bad, and most common ones (like sorting alphabetically) are a copy + paste away. Keep up the good work 👍

PR #722, opened 2019-06-17
MIA-FTWD-Mitch

mitchmalinin:


ta-web-mia:
Nice job getting through so many of these! Algorithms admittedly aren't something you'll run into very frequently when you're creating actual webpages, but they're very helpful in learning more about built-in JavaScript methods. Glad to see you've got a decent handle on all of this. Let me know if you have any questions, and keep up the good work 👍 
- Nick

PR #731, opened 2019-06-18
MIA-FTWD-Candido Concepcion -Not done

candidocon:


starter_code/src/movies.js
```diff
@@ -1,20 +1,126 @@
...
+        let regexHours = /[0-9]+h/g; //regex to find hours
+        let regexMinutes = /([0-9]+min)/g; //regex to find minutes
 
+        let hoursArray = newMovie.duration.toString().match(regexHours); //find hours
+        let minutesArray = newMovie.duration.toString().match(regexMinutes); //find minutes
+        
+        let durationInMin = 0;
+        
+        if(hoursArray !== null){
+            let hours = parseInt(hoursArray[0].replace(/\D/g, "")); //remove the h in hours
```
ta-web-mia: I like the way you went about solving this, but just a reminder that you can parseInt() a string if it starts with a number. For example, parseInt("44minutes") will give you 44. Might let you cut some steps out, but otherwise what you did is totally correct.

ta-web-mia:
Nice work! Don't worry about the last few steps, you did really well with this. Keep up the good work 👍 
- Nick

PR #734, opened 2019-06-19
MIA PTWD Alejandro Salas

alejotsv:
I need to keep working on the bonus


ta-web-mia:
Nice job with all of the main iterations! Keep up the good work 👍

PR #736, opened 2019-06-19
(MIA PTWD movies) - Oscar Leon

OscarLeon15:


ta-web-mia:
Awesome job! If you're comfortable with algorithms like these, then your JavaScript skills will carry you a long way. Keep up the good work 👍

PR #737, opened 2019-06-19
PTWD - MIA - Nelly Gonzalez-Vinas

nellygvinas:
Completed assignment without bonus completed. Still working through bonus, though I did look through the answer given in class.


ta-web-mia:
Awesome job! If you're comfortable with algorithms like these, then your JavaScript skills will carry you a long way. Keep up the good work 👍

PR #738, opened 2019-06-20
MIA PTWD Rachel Feldman

RFeldman84:


ta-web-mia:
The first iteration is definitely one of the hardest, so don't fret too much if you're feeling stuck on it. While algorithms aren't necessarily something you'll encounter often as you're coding, they do help reinforce fundamental JavaScript skills. If you have any questions, don't hesitate to reach out. Keep up the hard work 👍

PR #794, opened 2019-08-12
PAR-FTWD-Niraja

niraja426:


ta-web-paris:
Close enough, come to us for some explanations on the orderByDuration iteration and the hoursToMinutes.
The correction will be pushed soon.

PR #798, opened 2019-08-14
PAR FTWD DORO

DoriDoro:


ta-web-paris:
function convert(m) {
  let hoursToMins = m.duration.split(" ");
  if (hoursToMins.length === 1 && hoursToMins[0].includes("h")) {
    return parseFloat(hoursToMins[0]) * 60;
  }
  if (hoursToMins=== 1 && hoursToMins[0].includes("min")) {
    return parseFloat(hoursToMins[0]);
  }
  return parseFloat(hoursToMins[0]) * 60 + parseFloat(hoursToMins[1]);
}
-------------------------------------------------------------------------------
  if (hoursToMins.length=== 1 && hoursToMins[0].includes("min")) 

You forget the length property on the hoursToMins, thats why the last point doesnt go through : )

PR #813, opened 2019-08-17
BER-FTWD-Keighley

keighleymcf:


starter_code/src/movies.js
```diff
@@ -1,19 +1,133 @@
...
+  }, 0);
+  return Math.round((sum / arr.length) * 100) / 100;
+}
 
- 
 // Iteration 2: Drama movies - Get the average of Drama Movies
+function extractDramas(arr) {
+  let dramaMovies = arr.filter(function(movie) {
+    let containsDrama = false;
+    for (i = 0; i < movie.genre.length; i++) {
```
mjarraya: use one of the methods we've seen to check for occurrences in an array!

starter_code/src/movies.js
```diff
@@ -1,19 +1,133 @@
...
+  });
+  if (!spielbergDramas.length) return 0;
+  return spielbergDramas.length;
+}
 
 // Iteration 5: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(arr) {
+  let top20 = [];
+  let alphabeticalMovies = arr.sort(function(a, b) {
```
mjarraya: protect the original array against mutations

PR #815, opened 2019-08-17
BER-FTWD-JulianaFerreiraBastos

julferbas:
Lab exercises done, except bonus

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,19 +1,111 @@
-/* eslint no-restricted-globals: 'off' */
+/*eslint no-restricted-globals: 'off'*/
+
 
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,19 +1,111 @@
-/* eslint no-restricted-globals: 'off' */
+/*eslint no-restricted-globals: 'off'*/
+
 
 // Iteration 1: All rates average - Get the average of all rates with 2 decimals 
 
- 
+function ratesAverage(movies) {
+    let arrayOfRates = movies.map(function (movies) {
+        let rate = parseFloat(movies.rate);
```
mjarraya: while this is good and works, you could be doing the same inside the `reduce` function, to avoid iterating twice over the array 👍

PR #818, opened 2019-08-18
[FTWD-BER - Iouri] - Done

iourisorokine:
turnHoursToMinutes - 3 tests fail (return array, original array not modified, duration should be number) although all returned formats and types seem to match
Bonus done


starter_code/src/movies.js
```diff
@@ -1,19 +1,108 @@
...
+    return ratesAverage(dramaSelection);
+}
 // Iteration 3: Ordering by duration - Order by time duration, ascending (in growing order)
+function orderByDuration(moviesArr) {
+    const moviesCopy = moviesArr.slice();
+    if (moviesArr.length === 1) return moviesArr;
 
+    //
+    for (movie of moviesCopy) {
+        if (typeof movie.duration === "string") {
```
mjarraya: this still mutates the object in the original array, use the `...` operator to spread all the properties of the original object

starter_code/src/movies.js
```diff
@@ -1,19 +1,108 @@
...
+            ...movie,
+            duration
+        };
+    });
 
+    return moviesDurationCorrected;
+}
 // BONUS Iteration: Best yearly rate average - Best yearly rate average
+function bestYearAvg(moviesArr) {
+    let bestYear = 1900,
```
mjarraya: hacky! Create an array with all the distinct years instead.
Other than that, good job!

PR #821, opened 2019-08-18
BER-FTWD-Sarah

SarahHouben:
I'm aware that the code is not passing most of the Jasmine tests. However, when I control log the results, for most of the iterations, it is showing (what I believe is) the correct answer. 
I'm not sure where exactly I am going wrong here, but hope to be able to make more sense out of it after we've spoken about the Lab in class on Monday.


starter_code/src/movies.js
```diff
@@ -2,18 +2,128 @@
 
 // Iteration 1: All rates average - Get the average of all rates with 2 decimals 
 
- 
+function ratesAverage(movies) {
+
+    //map: create array with just the rates
+    //parseFloat: turn all strings into numbers.
+    let ratesArr = movies.map(function (obj) {
+        return parseFloat(obj.rate);
```
mjarraya: check for when `obj.rate` is `undefined`

starter_code/src/movies.js
```diff
@@ -2,18 +2,128 @@
...
+    let durationArray = movies.sort(function (a, b) {
+        if (a.duration > b.duration) {
+            return 1;
+        }
+        if (a.duration < b.duration) {
+            return -1;
+        }
+        if (a.duration === b.duration) {
+            let titleArray = movies.sort(function (a, b) {
+                if (a.title > b.title) {
```
mjarraya: use `localeCompare` instead

starter_code/src/movies.js
```diff
@@ -2,18 +2,128 @@
...
+}
+
+
 
 // Iteration 5: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(movies) {
+    //sort: order by title
+    let titleArray = movies.title.sort(function (a, b) {
+        return a - b;
```
mjarraya: sort strings with `localeCompare` to ensure correct alphabetical order

starter_code/src/movies.js
```diff
@@ -2,18 +2,128 @@
...
+    return spielbergArray;
+}
+
+
 
 // Iteration 5: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(movies) {
+    //sort: order by title
+    let titleArray = movies.title.sort(function (a, b) {
```
mjarraya: use `map` to get the array of titles

PR #867, opened 2019-08-28
MIA-FTWD-Enrique Narciso

e-narciso:
no bonus :(


ta-web-mia:
Nice job! Everything checks out, glad you were able to get the bonus.

PR #885, opened 2019-08-30
MIA-FTWD-JAMES THOMAS

hoshimotao:


ta-web-mia:
Good job! This is one of the hardest JS labs, so remember to use this as a reference when you want to review certain functions.

PR #887, opened 2019-08-31
MIA-FTWD Osheit Chang

osheitchang:
I'm still Working on the final iteration


ta-web-mia:
I'm going to close out this pull request, but please keep working on this when you get the time. This is probably one of the hardest JS labs, but it can help you understand a lot of key functions

PR #888, opened 2019-09-02
MIA FTWD Jorge Luis

jorgeluis42:


ta-web-mia:
Nice job! This is probably one of the harder labs to do, at least for the first module. If you can understand this, then most of JS will be a breeze!

PR #932, opened 2019-10-19
[PAR WDFT] Evgeny Amine

ESR93:
Hello TAs,
Thank you for your support, seems that the task is done =)

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2,18 +2,170 @@
 
 // Iteration 1: All rates average - Get the average of all rates with 2 decimals 
 
- 
+function ratesAverage(mv) {
+    let sm=0
+    mv.forEach(element => {
+        sm += parseFloat(element.rate)
+    });
+    avg = sm/mv.length
+    return avg
+}
+
 // Iteration 2: Drama movies - Get the average of Drama Movies
+
+function dramaMoviesRate(mv) {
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -2,18 +2,170 @@
...
+function dramaMoviesRate(mv) {
+    let sm=0
+    let dr_mv = mv.filter(function(val)
+        {
+        if (val.genre.includes('Drama'))
+        {
+           return val     
+        }
+        })
+    dr_mv.forEach(element => {
```
ta-web-paris: you could reuse your previous function to calculate the average

starter_code/src/movies.js
```diff
@@ -2,18 +2,170 @@
...
+        {   
+            if(a.year > b.year)
+            {
+                return 1
+            }
+            else if((b.year === a.year)&&((a.title.toUpperCase()) > (b.title.toUpperCase())))
+            {
+                return 1
+            }
+            else if((b.year === a.year)&&((a.title.toUpperCase()) < (b.title.toUpperCase())))
```
ta-web-paris: good habits to convert strings toUpperCase when you need to compare them

ta-web-paris:
Good job! 
Just be careful to better call your variables as it's not always super clear ;)

PR #938, opened 2019-10-20
[BER-FTWD-102019] - AreWS

arews:


starter_code/src/movies.js
```diff
@@ -1,19 +1,71 @@
...
+      } else {
+        return a.title - b.title;
+      }
+    }
+  });
+}
+
+/* Jasmine hates me!! I've checked my function, it sorts fine –
+even the titles alaphabetically. I've given up trying to please
+Jasmine in this one */
```
pierreportal: You don't need a for loop in your sort, you can write your conditionals direction in the sort method and user `a.title.localeCompare(b.title)` to sort strings

starter_code/src/movies.js
```diff
@@ -1,19 +1,71 @@
...
 
+function howManyMovies(movies) {
+  const spielbergMovies = movies.filter(function(movie) {
+    if (movie.director.includes("Steven Spielberg")) {
+      return true;
+    }
+    return spielbergMovies.length;
+  });
+  // I don't understand why my console says that spielbergMovies is not defined.
+}
```
pierreportal: speilbergMovies is undefined because you return it's value inside the filter. you should return it after the filtering

PR #940, opened 2019-10-20
BER-WDFT-Patrick

fluxpat:


starter_code/src/movies.js
```diff
@@ -1,19 +1,129 @@
...
+    dramaMovies = movies.filter( // Need to make movies a new array of JUST drama objects/movies
+        function (movie) {
+            if (movie.genre.indexOf("Drama") !== -1) {
+                return true;
+            }
+        }
+    );
+    if (!dramaMovies.length) return 0;
+    return ratesAverage(dramaMovies);
+}
```
pierreportal:
```
                return true;
            }
```

starter_code/src/movies.js
```diff
@@ -1,19 +1,129 @@
...
+    )
+    let sortedTwenty = sortedTitles.slice(0, 20);
+    let finalTwenty = [];
+    sortedTwenty.forEach(
+        function (titles) {
+            finalTwenty.push(titles.title)
+        }
+    )
+    return finalTwenty;
+}
```
pierreportal: Instead of creating an ampty array funalTwenty and push each title inside, you couls have map through sortedTwenty : 
`
let finalTwenty = sortedTwenty.map(function(movie){ return movie.title }) `

PR #942, opened 2019-10-20
BER-FTWD-1019-Shan

shanwong29:


starter_code/src/movies.js
```diff
@@ -1,19 +1,177 @@
 /* eslint no-restricted-globals: 'off' */
 
-// Iteration 1: All rates average - Get the average of all rates with 2 decimals 
+// Iteration 1: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(arr) {
+  const sum = arr.reduce(function(accumulator, currentValue) {
+    let number = parseFloat(currentValue.rate, 10);
```
itstheandre: Here I don't understand the second parameter in your parseFloat. there is no second parameter. What was the logic here? :)

starter_code/src/movies.js
```diff
@@ -1,19 +1,177 @@
...
+      return true;
+    }
+    //return element.genre.includes("Drama");
+  });
+
+  if (dramaMovies.length == 0) {
+    return 0;
+  }
 
+  const sumOfrate = dramaMovies.reduce(function(
```
itstheandre: After filtering the array, you could have just used the ratesAverage function here. It would have saved you quite some time :)

PR #944, opened 2019-10-20
[BER-FTWD-Johanna]

JohannaSiven:
please see the comments in iterations 1, 6 and Bonus

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,19 +1,104 @@
 /* eslint no-restricted-globals: 'off' */
 
+
 // Iteration 1: All rates average - Get the average of all rates with 2 decimals 
 
- 
+function ratesAverage(array) {
+    let sum = array.reduce(function (acc, value) {
+        if (value.rate.length !== 0) {
+            // tried the following line both with parseFloat and without - why is Jasmine test green also without it? The returned value without it is a string.
+            return acc + parseFloat(value.rate)
+        }
+        return acc;
+    }, 0);
+    return Math.round((sum / array.length) * 100) / 100;
+}
+
 // Iteration 2: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(array) {
+    let dramaMovies = array.filter(function (element) {
+        return element.genre.includes("Drama");
+    });
+    if (dramaMovies.length === 0) {
+        return 0;
+    } else {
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,19 +1,104 @@
...
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes(array) {
+    let formatted = array.map(function (el) {
+        return {
+            title: el.title,
+            year: el.year,
+            director: el.director,
+            duration: parseFloat((el.duration.substring(0, el.duration.indexOf("h")) * 60) + parseFloat(el.duration.substring(el.duration.indexOf(" "), el.duration.indexOf("m")))),
```
itstheandre: I noticed you had the same solution as Amelia.
In this case there's an error becasue in one of Jasmine's tests you are given a test with no hour.
And if you write parseFloat(' ') - which is the result of your el.duration.substring(0, el.duration.indexOf('h') - you get a Nan.

Here the soluton is to write the Number method.
Something like this:
`Number((el.duration.substring(0, el.duration.indexOf('h))`

Thats where the error comes :)

starter_code/src/movies.js
```diff
@@ -1,19 +1,104 @@
...
+        return null;
+    }
+    let groupedByYear = array.reduce(function (group, element) {
+        group[element.year] = [...group[element.year] || [], element];
+        return group;
+    }, {});
+    console.log(groupedByYear)
+    let yearKeys = Object.keys(groupedByYear)
+
+    // got stuck here, don't know how to utilize the yearKeys to create a data structure that i could pass to the ratesAverage function from iteration 1
```
itstheandre: Object keys only gives you the array of the keys. 
If you want to use the ratesAverage function - which is the correct approach if you want to get to the average value of each year, is also create an Object.values. this way you then have an array of arrays, each with movies from each year. 

This way, you now can get access to each value and perform your function for each individual year. As soon as you have the all the averages, you can, then,  check the index of the highest value. 

As soon as you have the index of the value you can get the year in that index from your variable yearKeys

JohannaSiven:
@itstheandre thanks a lot for the comments, it's all clear :) i"ll try to finish the bonus task at some point this week based on your comments, will get back to you if i face any issues but should be ok.

PR #945, opened 2019-10-20
almost done

noamrubin22:


starter_code/src/movies.js
```diff
@@ -1,19 +1,110 @@
...
 // Iteration 1: All rates average - Get the average of all rates with 2 decimals 
+function ratesAverage(movies) {
+    // check if array is not empty
+    if (!movies.length) return 0;
+
+    // create empty array
+    let movieRates = [];
+
+    // push all rates into array
+    movies.forEach(function (movie) {
```
itstheandre: In this case, a map would have achieved it much quicker.
.map(movie => movie.rate

PR #946, opened 2019-10-20
[BER-FTWD-Amelia] 97% done

scaram02:
I spent 13+ hours on this and decided to just...stop. I am dying inside. And maybe outside too

Please help us with the last iteration Monday in class. I had an idea how to do it, but never got around to fully executing it.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,19 +1,98 @@
 /* eslint no-restricted-globals: 'off' */
 
-// Iteration 1: All rates average - Get the average of all rates with 2 decimals 
+// Iteration 1: All rates average - Get the average of all rates with 2 decimals
 
- 
+function ratesAverage(array) {
+  const average =
+    array.reduce(function(accumulator, value) {
+      return accumulator + parseFloat(value.rate);
+    }, 0) / array.length;
+  return average;
+}
+
 // Iteration 2: Drama movies - Get the average of Drama Movies
 
+// should I change this to some if, return null thing
+
+// average
+function dramaMoviesRate(array) {
+  let dramAve = [];
+  for (var index = 0; index < array.length; index++) {
+    if (array[index].genre.includes("Drama")) {
+      dramAve.push(array[index]);
+    }
+  }
+  const sum = dramAve.reduce((total, value) => total + value.rate, 0);
+  if (dramAve.length === 0) {
+    return 0;
+  }
+  return Number((sum / dramAve.length).toFixed(2));
+}
 
 // Iteration 3: Ordering by duration - Order by time duration, ascending (in growing order)
 
+function orderByYear(array) {
+  let moviesArr = array.sort(function(a, b) {
+    if (a.year < b.year) {
+      return -1;
+    } else {
+      return 1;
+    }
+  });
+  return moviesArr;
+}
+
+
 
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,19 +1,98 @@
...
+
+// average
+function dramaMoviesRate(array) {
+  let dramAve = [];
+  for (var index = 0; index < array.length; index++) {
+    if (array[index].genre.includes("Drama")) {
+      dramAve.push(array[index]);
+    }
+  }
+  const sum = dramAve.reduce((total, value) => total + value.rate, 0);
```
itstheandre: One piece of advice for one of the main principles of coding - DRY (Dont repeat yourself). In this case you wrote basically the same function as you did before in the ratesAverage, so you didnt't need to rewrite it :)

starter_code/src/movies.js
```diff
@@ -1,19 +1,98 @@
...
+  if (dramAve.length === 0) {
+    return 0;
+  }
+  return Number((sum / dramAve.length).toFixed(2));
+}
 
 // Iteration 3: Ordering by duration - Order by time duration, ascending (in growing order)
 
+function orderByYear(array) {
```
itstheandre: Hey!
I think you forgot to do the last part of the function:
"If two movies have the same year, order them in alphabetical order by their title! ✔️"

starter_code/src/movies.js
```diff
@@ -1,19 +1,98 @@
...
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+// I figured the parseFloat was important, also wondering if I need to split somewhere? I didn't do this one completely by myself and also ran out of energy in terms of trying more methods
+function turnHoursToMinutes(array) {
+    let formatted = array.map(function (elem) {
+        return {
+            title: elem.title,
+            year: elem.year,
+            director: elem.director,
+            duration: parseFloat((elem.duration.substring(0, elem.duration.indexOf("h")) * 60) + parseFloat(elem.duration.substring(elem.duration.indexOf(" "), elem.duration.indexOf("m")))),
```
itstheandre: In this case, you're getting a NaN in one of the cases (and thus getting Jasmine to give you an error) because in one the test cases, the movie has no hour. And you're trying to do parseFloat on an empty string.
If you:
`console.log(parseFloat(' ')`

You will see the response is NaN.

In here, the way I would solve it, would be: instead of using parseFloat, I would use the Number method:
`parseFloat((elem.duration.substring(0, elem.duration.indexOf("h")) * 60) + parseFloat(elem.duration.substring(elem.duration.indexOf(" "), elem.duration.indexOf("m")))),
 

 
In this case, you're getting a NaN in one of the cases (and thus getting Jasmine to give you an error) because in one the test cases, the movie has no hour. And you're trying to do parseFloat on an empty string.
If you:
`console.log(parseFloat(' ')`

You will see the response is NaN.

In here, the way I would solve it, would be: instead of using parseFloat, I would use the Number method:
`
Number((elem.duration.substring(0, elem.duration.indexOf("h")) * 60) + Number(elem.duration.substring(elem.duration.indexOf(" "), elem.duration.indexOf("m")))),
`
And here the code would no longer break. Dont worry about it, its a very specific edge case.

PR #947, opened 2019-10-20
[BER-WDFT-102019-FABIAN]

fabianschu:


starter_code/src/movies.js
```diff
@@ -1,19 +1,143 @@
...
+    //create array of all movies of Spielberg
+    let spielbergArray = movies.filter(function(film) {
+        if (film.director.includes('Steven Spielberg') == true) {
+            return true;
+        }
+    });
 
+    //create array that contains only dramas
+    let dramaSpielbergArray =spielbergArray.filter(function(film) {
+        if (film.genre.includes('Drama') == true) {
```
itstheandre: In this case you didn't have to the the == true, because the includes method already returns a boolean. And if a value is true, or truthy, than the if statement runs. No need to add that extra code :)

starter_code/src/movies.js
```diff
@@ -1,19 +1,143 @@
...
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes (movies) {
+    let tempArray = movies.slice(0);
+    
+    tempArray = tempArray.map(function(movie) {
+        
+        //create temporary storage for duration format conversion 
+        let duration = movie.duration;
```
itstheandre: Even though in the end, the code resolves nicely, be careful because here, by declaring duration as the value inside the object, remember that if you make any change there, you're immediately changing the object. Ideally you would declare a variable and do the split from that original spot.
you have:
`let duration = movie.duration
duration = duration.split(' ')`

this can be tricky, and that's why Jasmine doesnt really appreciate it.

Like it's written in the error 
**Turn duration of the movies - turnHoursToMinutes Should return a new array, not update the original one**

You are directly changing that value. At the end of the day the code returns the correct value, but you should try to never change the original value. Like you did in the beginning by slicing the movies array

PR #948, opened 2019-10-20
BER-FTWD-102019-AlexServie

aslex:


starter_code/src/movies.js
```diff
@@ -1,19 +1,138 @@
...
+        if (movie.genre.includes("Drama")) {
+            return true;
+        }
+    });
+
+    if (dramaFilms.length === 0) {
+        return 0;
+    }
+
+    const dramaAvg =
```
itstheandre: Hey Alex!
In this case you basically took the ratesAverage function you did in the previous iteration. Ideally, you could have used that same function :) Something like ratesAverage(dramaFilms). With this you would have saved some time and keystrokes.
Remember the DRY principle in coding - Don't Repeat Yourself :) But very nice here

starter_code/src/movies.js
```diff
@@ -1,19 +1,138 @@
...
+        dramaFilms.reduce(function (total, movie) {
+            return total + parseFloat(movie.rate || 0);
+        }, 0) / dramaFilms.length;
+    return parseFloat(dramaAvg.toFixed(2));
+}
 
 // Iteration 3: Ordering by duration - Order by time duration, ascending (in growing order)
 
+function orderByYear(arr) {
```
itstheandre: In this particular function, for future reference, I would try and create a copy of the original array. the sort function mutates the original array. In this case there's no issue at all, but further along the line in the bootcamp (and beyond) people tend not to mutate the original value.

Still, the logic is 100% correct :)

starter_code/src/movies.js
```diff
@@ -1,19 +1,138 @@
...
 // Iteration 5: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(arr) {
+    const alph = arr.sort(function (movie, movie2) {
+        // return movie2.title - movie.title;  WHY DOESN'T THIS ONE WORK??
+        if (movie.title < movie2.title) {
+            return -1;
+        } else return 1;
+    });
+    alphaRay = alph.map(function (movie, index) {
```
itstheandre: In this case, even though it clearly worked, you could have just added a let or a const. It's easier to later check what is what.
By reading your code I wondered if you had declared it before.

PR #949, opened 2019-10-20
[BER-FTWD-Davide]

davideberno:


starter_code/src/movies.js
```diff
@@ -1,19 +1,98 @@
 /* eslint no-restricted-globals: 'off' */
 
-// Iteration 1: All rates average - Get the average of all rates with 2 decimals 
+// Iteration 1: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(arr) {
+  const ratingsSum = arr.reduce(function(a, b) {
+    return a + (parseFloat(b.rate) || 0); // Or !! => true = 1 and false = 0
+  }, 0);
+  return Math.round((ratingsSum / arr.length) * 100) / 100;
```
itstheandre: In this situation, even though you used Math.round - you could have also used toFixed(2) which is a Js method that also provides the same thing as your method. In this case would just make it shorter. But very nice approach!

starter_code/src/movies.js
```diff
@@ -1,19 +1,98 @@
...
 
+function orderAlphabetically(arr) {
+  const titlesArr = arr.map(movie => movie.title);
+  return titlesArr.sort().slice(0, 20);
+}
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes(arr) {
+  const minutesMovies = JSON.parse(JSON.stringify(arr));
```
itstheandre: Very nice idea here. If you understand the logic behind using these JSON methods, you're good :)

PR #950, opened 2019-10-20
[PAR_FTWD] Pierre-Henri

Jepassaispar:
I couldn't do the bonus exercise, couldn't get the logic out of it.

### DIFF
--- a/starter_code/SpecRunner.html
+++ b/starter_code/SpecRunner.html
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html>
+
 <head>
   <meta charset="utf-8">
   <title>Jasmine Spec Runner v2.8.0</title>
@@ -22,4 +23,5 @@
 
 <body>
 </body>
+
 </html>
--- a/starter_code/src/data.js
+++ b/starter_code/src/data.js
@@ -1,5 +1,4 @@
-let movies = [
-  {
+let movies = [{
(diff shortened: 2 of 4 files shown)


ta-web-paris:
Goog job! And don't worry for the bonus, it was an hard one ;) 
But here was my logic:

function bestYearAvg(movies){
  if(!movies.length) return null;
  const moviesByYear = movies.reduce((acc,movie) => {
    if(acc.hasOwnProperty(movie.year)){ //hasOwnProperty permet de savoir si cette propriété/key existe ou alors (curr.year in acc)
      acc[movie.year].push(movie.rate); //obj[key] = obj.key
      return acc;
    }else{
      const newArr = [];
      newArr.push(movie.rate);
      acc[movie.year]= newArr;
      return acc;
    }
  }, {})//inital value is an object
  const bestYear = {year: 0, rate: 0};

  for(let year in moviesByYear){
    let best = moviesByYear[year].reduce((acc, curr)=>{
      if(curr) return (acc += Number(curr))
      return acc
    },0) / moviesByYear[year].length;
    if(best > bestYear.rate){
      bestYear.year = year;
      bestYear.rate = best;
    }
  }
  return `The best year was ${bestYear.year} with an average rate of ${bestYear.rate}`
}

Jepassaispar:
Could I get a hint about the "Should return a new array, not update the original one" failure please ?

PR #951, opened 2019-10-20
BER-FTWD-Maria

mariakristiina:
Unfortunately I could not complete all, did not manage to make all the Jasmine errors green on 2 and didn't find a solution for 6.

Maria

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2,18 +2,77 @@
 
 // Iteration 1: All rates average - Get the average of all rates with 2 decimals 
 
+function ratesAverage(arr){ 
+let sum = arr.reduce(function(acc, value){
+    return acc + parseFloat(value.rate);
+}, 0); return sum / arr.length;
+};
  
 // Iteration 2: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(arr) {
+    let dramaMovies = arr.filter(function(movie){
+        if(movie.genre.indexOf("Drama") !== -1) {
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -2,18 +2,77 @@
...
 // Iteration 2: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(arr) {
+    let dramaMovies = arr.filter(function(movie){
+        if(movie.genre.indexOf("Drama") !== -1) {
+            return true;
+        }
+    } ); if(dramaMovies.length === 0) {
+        return 0};
+    let sum = dramaMovies.reduce(function(acc, value){
```
DalinaLeila: Here you could use the function you wrote in iteration 1 and just input the dramaMovies as your argument

starter_code/src/movies.js
```diff
@@ -2,18 +2,77 @@
...
+        } return a.title.localeCompare(b.title);
+         }); return sortedMovies;
+ };
 
 
 // Iteration 4: Steven Spielberg. The best? - How many movies did STEVEN SPIELBERG direct
 
+function howManyMovies(arr) {
+    let stevenMovies = arr.filter(function(movie){
+        if(movie.genre.indexOf("Drama") !== -1 && movie.director === "Steven Spielberg") {
```
DalinaLeila: you could use the includes method here which will return a boolean, so you then don't have to write !== -1

PR #956, opened 2019-10-21
[BER-FTWD-102019-Toska]

toskaMaliske:
not done - I was really struggling with this lab and was not able not pass all the tests yet, hope we can review it later.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2,18 +2,96 @@
 
 // Iteration 1: All rates average - Get the average of all rates with 2 decimals 
 
- 
+function ratesAverage(moviesArr) {
+    const sumRates = moviesArr.reduce(function (accumulator, value) {
+        return accumulator + parseFloat(value.rate);
+    }, 0) / moviesArr.length;
+    return sumRates;
+}
+
+
 // Iteration 2: Drama movies - Get the average of Drama Movies
+//!!!
+function dramaMoviesRate(arr) {
+    drama = arr.filter(function (item) {
+        return item.genre.includes("Drama");
+    });
+
+    const sumDrama = drama.reduce(function (accumulator, value) {
+        return accumulator + parseFloat(value.rate);
+    }, 0) / drama.length;
+    return sumDrama;
+};
 
 
 // Iteration 3: Ordering by duration - Order by time duration, ascending (in growing order)
 
+function orderByYear(movies) {
+    const newOrder = movies.sort(function (a, b) {
+        if (a.year !== b.year) {
+            //     // return a[title].localeCompare(b[title]);
+            // }
+            return a.year - b.year;
+        }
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -2,18 +2,96 @@
 
 // Iteration 1: All rates average - Get the average of all rates with 2 decimals 
 
- 
+function ratesAverage(moviesArr) {
+    const sumRates = moviesArr.reduce(function (accumulator, value) {
+        return accumulator + parseFloat(value.rate);
+    }, 0) / moviesArr.length;
```
DalinaLeila: I would divide by moviesArr.length after the reduce function.

so when you return

starter_code/src/movies.js
```diff
@@ -2,18 +2,96 @@
...
 
 // Iteration 3: Ordering by duration - Order by time duration, ascending (in growing order)
 
+function orderByYear(movies) {
+    const newOrder = movies.sort(function (a, b) {
+        if (a.year !== b.year) {
+            //     // return a[title].localeCompare(b[title]);
+            // }
+            return a.year - b.year;
+        }
```
DalinaLeila: } else {
    }

starter_code/src/movies.js
```diff
@@ -2,18 +2,96 @@
...
+    });
+    return newOrder;
+}
+
 
 // Iteration 4: Steven Spielberg. The best? - How many movies did STEVEN SPIELBERG direct
+function howManyMovies(arr) {
+    dramaSpielberg = arr.filter(function (value1, value2) {
```
DalinaLeila: dramaSpielberg is not defined

DalinaLeila: You also dont need two arguments. You can just have movie as one argument

starter_code/src/movies.js
```diff
@@ -2,18 +2,96 @@
...
+        if (value1.genre.includes("Drama") && value2.director == "Steven Spielberg");
```
DalinaLeila: if(movie.genre.includes("Drama") && movie.director === "Steven Spielberg") return true

starter_code/src/movies.js
```diff
@@ -2,18 +2,96 @@
...
+    });
+    return dramaSpielberg.length;
+};
+
 
 
 // Iteration 5: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(arr) {
+    const alphabeticalOrder = arr.sort(function (a, b) {
```
DalinaLeila: Here you sort the whole objects but you only want to return the titles of the movies sorted. 

So I would map the movie data and output only the titles into a new array. Then I would sort the titlesArray and lastly use the splice method .splice(0,20) to only output the first 20 titles

PR #958, opened 2019-10-21
[BER-WDFT-102019 Michael]

MikeVanYarnrider:
somehow done -> issues with the duration time return

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,19 +1,157 @@
 /* eslint no-restricted-globals: 'off' */
 
-// Iteration 1: All rates average - Get the average of all rates with 2 decimals 
+// Iteration 1: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(movies) {
+  let rate = [];
+  movies.forEach(function(element) {
+    rate.push(Number(element.rate));
+  });
 
- 
+  const average =
+    rate.reduce(function(total, amount) {
+      return total + amount;
+    }, 0) / movies.length;
+  return parseFloat(average.toFixed(2));
+}
+
 // Iteration 2: Drama movies - Get the average of Drama Movies
-
+function dramaMoviesRate(dramas) {
+  const filtered = dramas.filter(function(movie) {
+    if (movie.genre.indexOf("Drama") === 0) {
+      return true;
+    }
+  });
+  const dramaRating = ratesAverage(filtered);
+  if (filtered.length === 0) {
+    return 0;
+  }
+  return dramaRating;
+}
 
 // Iteration 3: Ordering by duration - Order by time duration, ascending (in growing order)
 
+function orderByYear(movies) {
+  let order = movies.sort(function(a, b) {
+    // if year is NOT equal
+    if (a.year !== b.year) {
+      return a.year - b.year; // return year ascending
+    }
+    {
+      if (a.title > b.title) {
+        return 1;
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,19 +1,157 @@
...
+  });
 
- 
-// Iteration 2: Drama movies - Get the average of Drama Movies
+  const average =
+    rate.reduce(function(total, amount) {
+      return total + amount;
+    }, 0) / movies.length;
+  return parseFloat(average.toFixed(2));
+}
```
DalinaLeila: It works perfectly fine! In order to write shorter code you could directly use the reduce method on the movies. And accessing the rate using the dot notation. 

e.g. 
  let totalSum = movies.reduce(function(accumulator, value) {
    return accumulator + parseInt(value.rate, 10);
  }, 0);
  return totalSum / movies.length;
}

starter_code/src/movies.js
```diff
@@ -1,19 +1,157 @@
...
+  const dramaRating = ratesAverage(filtered);
+  if (filtered.length === 0) {
+    return 0;
+  }
+  return dramaRating;
+}
 
 // Iteration 3: Ordering by duration - Order by time duration, ascending (in growing order)
 
+function orderByYear(movies) {
```
DalinaLeila: function orderByYear(movies) {
  movies.sort(function(a, b) {
    if (a.year === b.year) {
    } else {
      return a.year - b.year;
    }
  });
  return movies;
}

Here you could also use .localeCompare() in order to not have to write the two if statements.

starter_code/src/movies.js
```diff
@@ -1,19 +1,157 @@
...
+  console.table(order);
+  return order;
+}
+
+// console.log(movies);
+console.clear();
 
 // Iteration 4: Steven Spielberg. The best? - How many movies did STEVEN SPIELBERG direct
 
+function howManyMovies(director) {
```
DalinaLeila: Here you also need to filter for drama movies. (Look back at the explanation for iteration 4).

starter_code/src/movies.js
```diff
@@ -1,19 +1,157 @@
...
+  const filtered = director.filter(function(movie, genre) {
+    if (movie.director.indexOf("Steven Spielberg") === 0) {
```
DalinaLeila: here you could also just say if (movie.director === "Steven Spielberg")

starter_code/src/movies.js
```diff
@@ -1,19 +1,157 @@
...
+      return true;
+    }
+  });
+  console.table(filtered);
+  return filtered.length;
+}
 
 // Iteration 5: Alphabetic Order - Order by title and print the first 20 titles
```
DalinaLeila: Here i would map to create an array with only movie titles. Then sort them and lastly use the splice method to return only the first 20 items.

starter_code/src/movies.js
```diff
@@ -1,19 +1,157 @@
...
+    return minutes;
+  }
+  console.log(hour + minutes);
+  return hour + minutes;
+}
+// console.table(durationArr);
+// return durationArr;
+
 // BONUS Iteration: Best yearly rate average - Best yearly rate average
+function bestYearAvg(movies) {
```
DalinaLeila: Hint: Here you might use the function from iteration one that you already created.

PR #959, opened 2019-10-21
BER-FTWD-1019-Tatiana

crlybrcs:


starter_code/src/movies.js
```diff
@@ -1,16 +1,68 @@
...
 // Iteration 4: Steven Spielberg. The best? - How many movies did STEVEN SPIELBERG direct
 
+function howManyMovies(dramaMovies){
+
+  let isSpielberg = function(dramaMovies){
+    dramaMovies.director === 'Steven Spielberg';
+  }
+  let allMoviesSpielberg = dramaMovies.filter(isSpielberg);
+  
+  return allMoviesSpielberg;
```
DalinaLeila: you are missing "}" here. Otherwise, your code will not run.

starter_code/src/movies.js
```diff
@@ -1,16 +1,68 @@
 /* eslint no-restricted-globals: 'off' */
 
-// Iteration 1: All rates average - Get the average of all rates with 2 decimals 
+// Iteration 1: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(array) {
+  let rates = array.reduce(function(accumulator, currentValue) {
+  
+    return accumulator + parseFloat(currentValue.rate);
```
DalinaLeila: You can take away the parseFloat :)

starter_code/src/movies.js
```diff
@@ -1,16 +1,68 @@
...
+//     return iterator + parseFloat(genre.drama);
+//   });
+
+//   return (dramaMovie / array.length).toFixed(2);
+// }
+
+function dramaMoviesRate(movies) {
+  let dramaMovieCount = 0;
+  let dramaMovies = movies.reduce((rates, movie) => {
+    if (movies.genre.includes("Drama")) {
```
DalinaLeila: Here you want to target the single movie, not movies.genre. 


It might make sense here to filter first all drama movies and then get the average using the reduce method. But your code works as well after changing the movie - movies on line 31.

starter_code/src/movies.js
```diff
@@ -1,16 +1,68 @@
...
+    }
+    return rates;
+  }, 0);
+  if (!dramaMovieCount) {
+    return 0;
+  }
+  return Number((dramaMovies / dramaMovieCount).toFixed(2));
+}
 
 // Iteration 3: Ordering by duration - Order by time duration, ascending (in growing order)
```
DalinaLeila: function orderByYear(movies) {
  let ascedingMovies = movies.sort(function(a, b) {
    if (a.year === b.year) {
    } else {
      return a.year - b.year;
    }
  });
  return ascedingMovies;
}

starter_code/src/movies.js
```diff
@@ -1,16 +1,68 @@
...
+    if (a.year > b.year) {
+      return a.year - b.year;
+    } else if (a.year === b.year) {
+      return a.title - b.title;
+    }
+  });
+  return ascedingMovies;
+}
 
 // Iteration 4: Steven Spielberg. The best? - How many movies did STEVEN SPIELBERG direct
```
DalinaLeila: Here you need to filter. --> You could use a filter function that will directly look at both factors: only include drama movies and only movies directed by Steven S.

starter_code/src/movies.js
```diff
@@ -1,16 +1,68 @@
...
+function howManyMovies(dramaMovies){
+
+  let isSpielberg = function(dramaMovies){
+    dramaMovies.director === 'Steven Spielberg';
+  }
+  let allMoviesSpielberg = dramaMovies.filter(isSpielberg);
+  
+  return allMoviesSpielberg;
 
 // Iteration 5: Alphabetic Order - Order by title and print the first 20 titles
```
DalinaLeila: For this iteration, I would maybe map over the date to create a new array only with movie titles. Then I would sort this array and finally use the slice method to only print the first 20 values in that array.

PR #960, opened 2019-10-21
BER-WDFT-lab-javascript-greatest-movies-Madhavi

madhavi-yalamanchili:


starter_code/src/movies.js
```diff
@@ -1,19 +1,60 @@
...
 
+function dramaMoviesRate(movies) {
+  let dramaMovies = movies.filter(function(movie) {
+    return movie.genre.includes("Drama");
+  });
+  if (dramaMovies.length == 0) {
+    return 0;
+  } else {
+    let totalDramaRates = dramaMovies.reduce(function(accumulator, dramaMovie) {
+      return accumulator + dramaMovie.rate * 1;
```
DalinaLeila: here you can take away the  * 1 :)

starter_code/src/movies.js
```diff
@@ -1,19 +1,60 @@
...
+    }, 0);
+    return Math.round((totalDramaRates / dramaMovies.length) * 100) / 100;
```
DalinaLeila: This can be rewritten as:

  return Number((totalDramaRates / dramaMovies.length).toFixed(1));

(Number means you want to turn the result into a number (not a string)). 
(toFixed(1) means you want to only return one decimal place)

starter_code/src/movies.js
```diff
@@ -1,19 +1,60 @@
...
 // Iteration 5: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(movies) {
+  return movies
+    .map(movie => movie.title)
+    .sort((a, b) => a.localeCompare(b))
+    .slice(0, 20);
+}
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
```
DalinaLeila: I know this iteration is a bit more tricky but I'm sure you will be able to do it! Have a look at the parseInt method. Since the duration value is written as a string, you need to think of a way to only get the numbers you want to calculate the duration from hours to minutes.

starter_code/src/movies.js
```diff
@@ -1,19 +1,60 @@
...
 
 // BONUS Iteration: Best yearly rate average - Best yearly rate average
```
DalinaLeila: Here you could implement the function "ratesAverage" which you have written already in iteration 1 :)

PR #962, opened 2019-10-21
PAR FTWD ANDY

N10T:
I dont succeed to create a new array and for turn it to a number


ta-web-paris:
Good job, you would only need Number() (--> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) and to use a function that creates an new array :)

PR #965, opened 2019-10-22
[PAR-FTWD] Parina - Thomas - Greatest movies

ThomasAlbrt:
Done except bonus

### DIFF
--- a/starter_code/src/data.js
+++ b/starter_code/src/data.js
@@ -1,5 +1,4 @@
-let movies = [
-  {
+let movies = [{
     title: 'The Shawshank Redemption',
     year: '1994',
     director: 'Frank Darabont',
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Good job! 
This is one way to proceed for the last iteration :) 

`function convertToMin(hourAndMin) {
  if (hourAndMin.includes("h")) {
    return (
      Number(hourAndMin.substring(0, hourAndMin.indexOf("h"))) * 60 +
      Number(
        hourAndMin.substring(
          hourAndMin.indexOf(" ") + 1,
          hourAndMin.indexOf("m")
        )
      )
    );
  } else
    return Number(
      hourAndMin.substring(hourAndMin.indexOf(" ") + 1, hourAndMin.indexOf("m"))
    );
}

  return movies.map(x => {
    const y = Object.assign({}, x);
    y.duration = convertToMin(String(y.duration));
    return y;
  });
}`

PR #969, opened 2019-10-24
MIA FTWD OCT Juan Grajales

JuanGrajales:


ta-web-mia:
So far so good! The turn hours to minutes function is a doozy, but you're on the right track. If you have time, go back and see if you can figure things out. Array.split and parseInt() should be helpful. Keep it up :)
- Nick

PR #970, opened 2019-10-24
LIS-FTWD-[LEOPOLDO]-174

oleoleoleoleo:
@ironhackLis-webTeam
so, jasmine doesnt pass  the tests: 

should return a new array, not update the original one
should return the duration of the movie as a number

but im quite sure the function returns a new array, with all the durations as a number


ironhackLis-webTeam:
Hey Leo!

I figure this is what was mentioned this morning in class. Any more questions, let us know!

Cheers,
- Guilherme

PR #973, opened 2019-10-24
MIA FTWD October 2019 - Julian Phillips

Kadus90:


ta-web-mia:
Nice job! Don't worry about those weird edge cases in the 2nd to last iteration. Console shows that everything is good, so it's probably on Jasmine's end. Keep up the good work :)
- Nick

PR #975, opened 2019-10-24
MIA FTWD Masaki Endo

mendosaurus:


ta-web-mia:
Nice work! I'm glad to see you got a head start on a lot of these labs. Things will pick up soon though, so stay focused! Keep up the good work :)
- Nick

PR #984, opened 2019-10-25
MIA FTWD October - Daphne Feresin

daph3105:
done from 1-5, still trying to work on 6 and 7 on my own


ta-web-mia:
Nice work so far! The turn hours to minutes function is pretty tricky, so let us know if you need any help with it. Keep it up :)
- Nick

PR #985, opened 2019-10-25
MIA FTWF OCT Vitalii Agapov 

g1yk:


ta-web-mia:
Nice work so far! Iteration 6 is really tricky, so definitely don't feel bad if you're struggling with it. Keep it up, and let me know if you have any questions :)
- Nick

PR #992, opened 2019-10-25
MIA - FTWD - October - Avrahm Kleinholz

avrahm:


ta-web-mia:
Nice work! The hours to minutes function is really tricky, so don't feel bad about not quite finishing that iteration. If you have extra time and want to fix things up, considering going back to it. Keep it up :)
- Nick

PR #993, opened 2019-10-25
[MIA-FT-102019] - Kristina Bressler

kikidesignnet:
still not finished. still working on that


ta-web-mia:
Nice work so far! Iteration 6 is super tricky, so nice job getting it. I wouldn't worry about the bonus iteration (unless you have extra time), but it's definitely good to get as much JavaScript practice in as you can. I'm going to close the pull request and mark you off as complete, but you're more than welcome to go back and work on the bonus iteration if you have time. Keep up the awesome work :)
- Nick

PR #995, opened 2019-10-26
PP-MX-WDPT (Erika ,Gustavo, Eduardo);

Any28Flo:


starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+    });
+  
+    return moviesYear;
+
+};
 
+// Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+let howManyMovies = function(array){
+    let arrayHowMany =array.filter(element => {
+       return element.director === "Steven Spielberg" && element.genre.includes("Drama");
```
ta-web-mex: nice, but you need be careful using comparators `===` or `==`

starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+let orderByYear = function(array){
+    //Creando una copia del arreglo original (movies)
+    let moviesYear = array.map(function (element){
+        return element;
+    });
+
+   moviesYear.sort(function(a,b){
+    if(a.year> b.year) return 1;
+    if (a.year< b.year) return -1;
+    if(a.title >b.title ) return  1;
```
ta-web-mex: this function can be improved like this
```
     if(a.year> b.year) return 1;
     if (a.year< b.year) return -1;
     if(a.title >b.title ) return  1;
     if(a.title < b.title) return -1;
```

starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+let howManyMovies = function(array){
+    let arrayHowMany =array.filter(element => {
+       return element.director === "Steven Spielberg" && element.genre.includes("Drama");
+     
+      });
+    
+     return arrayHowMany.length;
+ }
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
+let orderAlphabetically = function (array){
```
ta-web-mex: ok for this function you need order the elements first, after that get the first 20 titles after order and return the result :)

starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+           
+        firstElements =array.slice(0,20);
+        return firstElements.sort();
 
+    }
+  
+};
+orderAlphabetically(movies);
 // Iteration 4: All rates average - Get the average of all rates with 2 decimals
+let ratesAverage = function(array){
```
ta-web-mex: this solution can be improve, like a first mapping the movies, to get one new array with only rates, after that use `reduce` to get the sum and get the average.

ta-web-mex:
Nice work, but you need work more hard

PR #1041, opened 2019-11-08
[PTWDMIA102019] Gustavo Jordão

gustavofjordao021:
Work on iterations #1-3


MarcosPalaciosIII:
Good job on the all time movies lab. This remember that test driven development (TDD) is used when trying to keep code consistent and also allows so that everything will function the way the developer intends for it to work. It can be tricky to work with at first, but once you get used to how the tests run, it will become easier to work with.

Keep up the good work.

PR #1044, opened 2019-11-09
[PTMIA102019]Delvis

tati85:
fist version of lab No bonus


MarcosPalaciosIII:
Great job on the all time movies lab. This remember that test driven development (TDD) is used when trying to keep code consistent and also allows so that everything will function the way the developer intends for it to work. It can be tricky to work with at first, but once you get used to how the tests run, it will become easier to work with.

Keep up the great work.

PR #1106, opened 2020-01-10
[PAR]-[WDFT]-PASCAL-BESSON [ongoing]

pascalbes:


starter_code/src/movies.js
```diff
@@ -1,15 +1,373 @@
...
+  let arr3 = [];
+  if (arr.length === 0) {
+    return 0;
+  } else {
+    arr2 = arr.filter(a => a.director == "Steven Spielberg");
+    arr3 = arr2.filter(a => a.genre.indexOf("Drama") !== -1);
+    return arr3.length;
+  }
+}
+
```
frankmarmier: Here you could do everything within a single filter,  since you have 2 conditions that have to be met, you can use the && operator instead of filtering your array twice, and if nothing meets the condifitions, it will create an empty array.

Solution:

```
function howManyMovies(arr) {
 return arr.filter(m => m.director === "Steven Spielberg" && m.genre.includes("Drama")).length;
}
```

starter_code/src/movies.js
```diff
@@ -1,15 +1,373 @@
...
+    arrTemp.length >= 20 ? (arrTemp = arrTemp.slice(0, 20)) : arrTemp;
+
+    var arrFinal = [];
+
+    arrTemp.forEach(v => arrFinal.push(v.title));
+
+    return arrFinal;
+  }
+}
```
frankmarmier: For this function, since the sort method mutates the original array and you must create a new array containing only the titles, you could first  .map() the array (which creates a new array), then sort it then slice it.
I assume that you're comparing the length before slicing in case of some error, if the 2nd parameter is bigger than the arrays length, it will slice until the end.
Solution: 
```
function orderAlphabetically(arrNotMutated) {
  return arrNotMutated.map(m => m.title).sort((a,b) => a.localeCompare(b)).slice(0,20)
}
```

starter_code/src/movies.js
```diff
@@ -1,15 +1,373 @@
...
 // Iteration 4: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(arr) {
+  if (arr.length == 0) {
+    return 0;
+  } else {
+    sum = 0;
+    arr.forEach(v => (!isNaN(v.rate) ? (sum += v.rate) : 1));
+    return Math.round((sum / arr.length) * 100) / 100;
+  }
+}
```
frankmarmier: Here you could use the reduce method.

Solution with part of your code : 

```
function ratesAverage(arr) {
  if (arr.length == 0)return 0
    return Math.round((arr.reduce((acc,curr) => {
      if(!isNaN(curr.rate)) return acc += curr.rate;
      return acc;
    } ,0)/arr.length) * 100 ) / 100;
}
```

starter_code/src/movies.js
```diff
@@ -1,15 +1,373 @@
...
+      .filter(v => v.genre == "Drama")
+      .forEach(v => {
+        if (!isNaN(v.rate)) {
+          sum += v.rate;
+          count++;
+        }
+      });
+    return Math.round((sum / count) * 100) / 100;
+  }
+}
```
frankmarmier: Just above this function, you already had a function in which you had to just send an array with rates and it would return the average.
So you could first filter the array and then use your ratesAverage() function to get the average of the filtered array.

starter_code/src/movies.js
```diff
@@ -1,15 +1,373 @@
...
+
+    averages.forEach((v, i, arr) => (arr[i] = v / count[i]));
+
+    daAverage = Math.max(...averages);
+    daYear = years[averages.indexOf(daAverage)];
+
+    return `The best year was ${daYear} with an average rate of ${daAverage}`;
+  }
+}
+
```
frankmarmier: Nice solution.

Just be careful about O(n), you are iterating many times over arrays.
Some more information about what is O(n)

https://en.wikipedia.org/wiki/Big_O_notation

frankmarmier:
All tests are passing, I commented for little optimisation, good overall, well done : )

PR #1107, opened 2020-01-10
[PAR-WEBFT] MEYER / MORSLI - All valid

ameyerfr:


starter_code/src/movies.js
```diff
@@ -1,15 +1,115 @@
...
+  // Sort movies of the same year by title
+  moviesYears.sort( (a,b) => {
+    if (a.year === b.year) {
+      return a.title === b.title ? 0 : a.title > b.title ? 1 : -1;
+    }
+  });
+
+  return moviesYears;
+}
+
```
ta-web-paris: Here you could chain everything at once, and especially, you can add more logic to your sort function.

moviesYears.sort((a,b) => {
if(a.year === b.year) return a..title.localeCompare(b.title)
return a.year - b.year;
}
Information on localeCompare: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare

starter_code/src/movies.js
```diff
@@ -1,15 +1,115 @@
...
+  });
+
+  // Order the list alphabetically
+  moviesList.sort( (a,b) => {
+    return a === b ? 0 : a > b ? 1 : -1;
+  });
+
+  // Return first 20 elements
+  return moviesList.slice(0,20);
+}
```
ta-web-paris: Here again you could chain everything at once

```
  return movies.map(m => m.title).sort((a,b) => a.localeCompare(b)).slice(0,20)
}
```

The map creates a new array, then your sort it, then you slice it (which returns a new array)

ta-web-paris:
Well done ! closing the PR.

starter_code/src/movies.js
ameyerfr: Agreed, this one was very verbose for no reason !

starter_code/src/movies.js
ameyerfr: Agreed ! This one was very verbose for no reason.

PR #1109, opened 2020-01-10
[PAR-WDFT] - Anthony B. & Cécile . [ongoing]

abrebion:


starter_code/src/movies.js
```diff
@@ -1,15 +1,2122 @@
...
+//     });
+// }
+
+function orderByYear(array) {
+  const newArray = [...array];
+  return newArray.sort((a, b) =>
+    a.year === b.year ? (a.title < b.title ? -1 : 1) : a.year - b.year
+  );
+}
+
```
ta-web-paris: If you want to make it short you could do this: 
```
function orderByYear(array) {
  return [...array].sort((a, b) => (a.year === b.year ? a.title.localeCompare(b.title) : a.year - b.year));
}
```
localeCompare is a function on the string prototype that returns either 1 , - 1 or 0, which is what the sort function is expecting :D

starter_code/src/movies.js
```diff
@@ -1,15 +1,2122 @@
...
 
+function orderAlphabetically(array) {
+  const newArray = [...array];
+  let arr = [];
+  newArray.sort((a, b) => a.title.localeCompare(b.title));
+  newArray.forEach(el => arr.push(el.title));
+  arr.splice(20);
+  return arr;
+}
+
```
ta-web-paris: Here you could have mapped (creating a new array) to get only the titles, then sort that array, and the slice it.

ta-web-paris:
Nice solutions, good job ! Closing the PR.

PR #1110, opened 2020-01-10
[BER-FT-012020] Mia Lavanti

mialav:
Donezo

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2,14 +2,102 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(arr) {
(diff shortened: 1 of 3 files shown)


starter_code/src/movies.js
```diff
@@ -2,14 +2,102 @@
...
+  });
+  return orderedByYear;
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+function howManyMovies(arr) {
+  const howManyStevens = arr.filter(function(movie) {
+    if (
+      movie.director === "Steven Spielberg" &&
+      movie.genre.indexOf("Drama") > -1
```
itstheandre: another way of doing this line here would be with a 
```
movie.genre.includes("Drama")
```

The includes method also returns a boolean

PR #1112, opened 2020-01-10
[BER-FT] Elsa

ElsaFil:


starter_code/src/movies.js
```diff
@@ -2,14 +2,135 @@
...
+        minutes = Number(timeHM[0].slice(0, -1) * 60);
+      } else if (timeHM[0].indexOf("min") > -1) {
+        minutes = Number(timeHM[0].slice(0, -3));
+      }
+    }
+    let shallowCopyMovie = {
+      ...movie
+    };
+    shallowCopyMovie.duration = minutes;
+    return shallowCopyMovie;
```
itstheandre: in this case you could have saved 5 lines of code with:
```
return {...movie, duration: minutes}
```

starter_code/src/movies.js
```diff
@@ -2,14 +2,135 @@
...
+      return b.year - a.year;
+    }
+    return a.averageRate - b.averageRate;
+  });
+  return (
+    "The best year was " +
+    averages[averages.length - 1].year +
+    " with an average rate of " +
+    averages[averages.length - 1].averageRate
+  );
```
itstheandre: Here you could have also used template literals for an easier to read code:

```
`The best year was ${averages[averages.length-1].year with an average of ${averages[averages.length-1].length}`
```

And even there you could have made it easier on yourself by defining the value before:
```
let highestAvg = averages[averages.length-1]
return `The best year was ${highestAvg.year} with an average rate of ${highestAvg.averageRate}` 
```

PR #1115, opened 2020-01-11
[PAR]-[WDFT]-FRANCK-TREBOIT-GREATEST-MOVIES . [ongoing]

franck-treboit:
Quite ok, list of years not yet automatic.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,15 +1,148 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies) {
+   
+    let newMovies = [...movies];
+    if ( movies.length == 0  ){
+       return newMovies ;
+    }
+    
+    newMovies.sort(function(moviea, movieb) { 
+        return moviea.year - movieb.year  ;
+    });
+    
+    newMovies.sort((a, b) => (a.year > b.year) ? 1 : (a.year === b.year) ? ((a.title > b.title) ? 1 : -1) : -1 )
+     
+    //console.log(newMovies); 
+    return newMovies;    
+}    
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+    function howManyMovies(movies) {
 
+        if (movies.length == 0){
+            return 0;
+        }      
+
+        const moviesNumbers = movies.filter(movie => movie.director == "Steven Spielberg" );
+
+        const moviesNumbers2 = moviesNumbers.filter(movie2 => movie2.genre.indexOf('Drama') > -1 );
+
+        return moviesNumbers2.length; 
+
+    } 
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
+    function orderAlphabetically(movies) {
+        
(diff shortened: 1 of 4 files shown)


starter_code/src/movies.js
```diff
@@ -1,15 +1,148 @@
...
+    let newMovies = [...movies];
+    if ( movies.length == 0  ){
+       return newMovies ;
+    }
+    
+    newMovies.sort(function(moviea, movieb) { 
+        return moviea.year - movieb.year  ;
+    });
+    
+    newMovies.sort((a, b) => (a.year > b.year) ? 1 : (a.year === b.year) ? ((a.title > b.title) ? 1 : -1) : -1 )
```
ta-web-paris: Here you could blend in your sort functions.
```

    let newMovies = [...movies];
    if ( movies.length == 0  ){
       return newMovies ;
    }
    
    newMovies.sort(function(moviea, movieb) { 
        return moviea.year === movieb.year ? moviea.title.localeCompare(movieb.title) : moviea.year - movieb.year
        // More detailed : 
        // if(moviea.year === movieb.year){
        //     return moviea.title.localeCompare(movieb.title)
        // }else return moviea.year - movieb.year;
    });
    return newMovies;    
}    
```
More about localeCompare: 

starter_code/src/movies.js
```diff
@@ -1,15 +1,148 @@
...
+            if (a.title.toLowerCase() > b.title.toLowerCase() ) return 1;   
+            if (a.title.toLowerCase() < b.title.toLowerCase() ) return -1;
+            return 0;
+        })
+        
+        let moviesTitle = [];
+        newMovies.forEach(element => moviesTitle.push(element.title));
+        
+        return moviesTitle.slice(0, 20) ;
+    };
```
ta-web-paris: Here since you had to transform the original array and get only the titles, you could have used map for it which creates a new copy of the array.

Then sort it and then slice it.

starter_code/src/movies.js
```diff
@@ -1,15 +1,148 @@
...
+    function ratesAverage(movies) {
+        
+        if (movies.length == 0){
+            return 0;
+        }      
+        
+        let movieNumber = 0;        
+        var total = movies.reduce(function(accumulator, currentValue){            
+            if ( !isNaN(currentValue.rate) ){
+                movieNumber = movieNumber+1;
```
ta-web-paris: I guess you can remove your movieNumber variable here :D

starter_code/src/movies.js
```diff
@@ -1,15 +1,148 @@
...
+            return {
+              name: movie.name,
+              title:  movie.title,
+              year: movie.year,
+              director: movie.director,
+              duration: Number(convertDuration(movie.duration)),
+              genre:movie.genre,
+              rate: movie.rate,
+            }
+          });
```
ta-web-paris: Nice way of creating a new object, to make your life easier you could have done it this way:

Here is a simpler solution since you want to get all the keys and values but only want to change one key / value pair 

```
        
        const movies1 = movies.map(movie => {
            return {
                ...movie,
              duration: Number(convertDuration(movie.duration)),
           
            }
          });
          
        return movies1 ;
    }
```

ta-web-paris:
Good solutions ! Well done, closing the PR.

PR #1117, opened 2020-01-11
[BER-FT] Karo

karo-cz:


starter_code/src/movies.js
```diff
@@ -1,15 +1,157 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(array) {
+  let sortedArray = array.slice().sort(function(a, b) {
+    if (a["year"] === b["year"]) {
+      return a["title"].localeCompare(b["title"]);
+    }
```
itstheandre: I dont know if this was you testing using bracket notation with the object , but you could also use the "dot" notation.
```
a.year
```

karo-cz: I thought that if the key is in " ", then I need to access it with [""].

Thanks :)

PR #1119, opened 2020-01-11
[PAR-WDFT-0120] Pierre Etryhard - ongoing

PierreEtryhard:


Jepassaispar:
All looks good 👍 just few details : 
```
    if (arr.length === 0) return 0;
    let newTime = [...arr];
    let duration = 0;
    newTime = arr.map(function(value) {
        return {
            name: value.name,
            title:  value.title,
            year: value.year,
            director: value.director,
            duration: hoursToMinutes(value.duration),
            genre: value.genre,
            rate: value.rate,         
        }
    });
    return newTime;
}
```
Be carefull of the variable you are declaring, I saw 3 var like the `let duration = 0` that are not used in your functions.

When you return `name: value.name` it is undefined, I guess it was the same as the title.

Those are just innatentions, you got the concepts, very good job :)

PR #1120, opened 2020-01-11
[PAR-WDFT] - Nina, Paul D, Anthony L [ongoing]

Gorlanova:
my brain is melting right now, thanks and have a nice week-end !

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,15 +1,77 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(array) {
+    if (array.length === 0) return null;
+    let newArray = array.sort((a, b) => {
+        if (a.year - b.year === 0) {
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,15 +1,77 @@
...
 
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(array) {
+    if (array.length === 0) return null;
+    let newArray = array
+        .map(movie => movie.title)
+        .sort((a, b) => {
+            if (a < b) return -1
+            else return 1
+        })
```
ta-web-paris: you could use 

```
  .sort((a, b) => {
      return a.localeCompare(b);
    });
```
More about localeCompare: (it returns either 1, -1, or 0) exactly what the sort method expects :D 

Gorlanova:
Nice and handy ! Thanks :)

ta-web-paris:
Nice solutions, if you want to make it shorted you could chain the functions.

But everything works so it's good for me :D
Closing the PR.

PR #1122, opened 2020-01-11
[FT BER] Nora Nicklis

Norizinus:
There are two errors from Jasmine that I cannot resolve. And I can also not reproduce them. 

They both state "movie.duration.split is not a function" (I'm performing  .split(" ") on a string, line 336) 
I've looked into the test script that runs in the background and these are the two things that are being checked that seem to cause the error: 
1. expect(typeof turnHoursToMinutes(movies)).toBe("object")
2. expect(turnHoursToMinutes(movies)).not.toEqual(movies)

I've run both commands in the console and get the expected result. Maybe you can see something I can't... because I give up ¯\_(ツ)_/¯

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -14,249 +14,6 @@ function orderByYear(itemsToSort) {
   return sortedItems;
 }
 
-let unsortedMovies = [
-  {
-    title: "Saving Private Ryan",
-    year: 1998,
-    director: "Steven Spielberg",
-    duration: "2h 49min",
-    genre: ["Drama", "War"]
-  },
-  {
-    title: "The Shawshank Redemption",
-    year: 1994,
-    director: "Frank Darabont",
-    duration: "2h 22min",
-    genre: ["Crime", "Drama"],
-    rate: 9.3
-  },
-  {
-    title: "The Godfather",
-    year: 1972,
-    director: "Francis Ford Coppola",
-    duration: "2h 55min",
-    genre: ["Crime", "Drama"],
-    rate: 9.2
-  },
-  {
-    title: "The Godfather: Part II",
-    year: 1974,
-    director: "Francis Ford Coppola",
-    duration: "3h 22min",
-    genre: ["Crime", "Drama"],
-    rate: 9
-  },
-  {
-    title: "The Dark Knight",
-    year: 2008,
-    director: "Christopher Nolan",
-    duration: "2h 32min",
-    genre: ["Action", "Crime", "Drama", "Thriller"],
-    rate: 9
-  },
-  {
-    title: "12 Angry Men",
-    year: 1957,
-    director: "Sidney Lumet",
-    duration: "1h 36min",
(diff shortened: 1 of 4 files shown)


starter_code/src/movies.js
```diff
@@ -2,14 +2,399 @@
...
+  sortedItems.sort(function(a, b) {
+    if (a.year === b.year) {
+      return a.title.localeCompare(b.title);
+    }
+    return a.year - b.year;
+  });
+  return sortedItems;
+}
+
+let unsortedMovies = [
```
fenrew: This is not necessary as all the movies is unsorted within the data file and since the data file has been linked in the html file you can console.log(movies) and you will see the whole array of unsorted movies

starter_code/src/movies.js
```diff
@@ -2,14 +2,399 @@
...
 // Iteration 5: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(movies) {
+  console.log(movies);
+  let dramaMovies = movies.filter(function(movie) {
+    // console.log(movie.genre);
+    // console.log(movie.genre.includes("Drama"))
+    return movie.genre.includes("Drama");
+  });
+  return ratesAverage(dramaMovies);
```
fenrew: Nice!

starter_code/src/movies.js
```diff
@@ -2,14 +2,399 @@
...
+}
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes(movies) {
+  let moviesInMinutes = [...movies];
+
+  let movieDurations = moviesInMinutes.map(function(movie) {
+    let timeValues = movie.duration.split(" ");
```
fenrew: To get rid of the .split is not a function error you can write this line above line 336:
let movieDur =  movie.duration + ''
let timeValues = movieDur.spli(" ")

The reason for this is probably because movie.duration is a number and by adding '' to it, it'll be converted to a string. I belive .toString() will also work. This also solves the "must return a number" fault in Jasmine

starter_code/src/movies.js
```diff
@@ -2,14 +2,399 @@
...
+    // console.log(movie.genre.includes("Drama"))
+    return movie.genre.includes("Drama");
+  });
+  return ratesAverage(dramaMovies);
+}
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes(movies) {
+  let moviesInMinutes = [...movies];
```
fenrew: This right here is giving you the "you need to return a new array" mistake. So I believe the issue is that when you clone it like this you only clone the actual array, and not everything nested within them. You can read a bit about it by searching "Deep clone js", and as always, you can ask me if you don't understand this, it's just a bit difficult to explain over messages. Solution:
// let moviesInMinutes = [...movies], instead do it like this:
let moviesInMinutes = JSON.parse(JSON.stringify(movies))

This is one way to create a Deep copy that has it's own memory reference

PR #1123, opened 2020-01-12
[PAR - WDFT] Remy/Wendy . [ongoing]

kurosaek:
ITERATION 3 :
working but not accepted by Jasmine

ITERATION 6 : 
my condition (if) is not working but seems ok when i'm testing it on a Repl.it

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,15 +1,108 @@
+const films = require("./data")
+
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(arr) {
+    if (arr.length ===1) {
+        return arr
+    }
+    var newArray = [...arr];
+    let sortedMovies = newArray.sort(function(a,b) {
+        if (a.year !== b.year) {
+            return a.year - b.year;
+        } else return a.title.localeCompare(b.title)
+    }) 
+    return sortedMovies
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
 
+function howManyMovies(arr) {
+    var counter = 0;
+    arr.filter(function(a) {
+        if (a.director === "Steven Spielberg" && a.genre.includes("Drama")){
+            counter++
+        }
+    }); return counter
+}
+
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
+
+function orderAlphabetically(arr) {
+    var newArray = [];
+    if (arr.length < 20) {
+        newArray = arr.map(function(a){
+            return a.title;
+        })
+    } if (arr.length > 20) {
+        newArray = arr.splice(0, 20).map(function(a){
+            return a.title
+        }).sort(function(a,b){
+           return a.localeCompare(b)
+        })
+    }; return newArray;
+}
+
+// WORKING BUT Not ACCEPTED BY JASMINE?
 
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+            return a.title;
+        })
+    } if (arr.length > 20) {
+        newArray = arr.splice(0, 20).map(function(a){
+            return a.title
+        }).sort(function(a,b){
+           return a.localeCompare(b)
+        })
+    }; return newArray;
+}
```
ta-web-paris: It's not accepted by Jasmine because it's not working when the array has less than 20 elements :D.
You still have to  sort the array even if there are less than 20 elements.

starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+        return 0
+    }
+    var sumRates = 0;
+    arr.map(function(a){
+        if (Boolean(a.rate) === false) {
+            a.rate = 0
+        }
+        return sumRates += a.rate
+    }); return Math.round((sumRates / arr.length)*100) / 100
+}
```
ta-web-paris: You could have used the reduce() method here.

starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+    });
+    if (dramaMovies.length === 0) {
+        return 0
+    };
+        dramaMovies.map(function(a){
+        sumDramaRate += a.rate
+    }); 
+    return Math.round((sumDramaRate / dramaMovies.length)*100) / 100
+}
+
```
ta-web-paris: Here you are creating an empty array in which your are pushing the values from the filter.
The filter creates a new array according to the conditions you put set in the call back function:
You should have done it this way: 
```

var dramaMovies = arr.filter(function(a){
                                        return a.genre.inculudes("Drama")
}
```
The filter method will only include in the output array the elements that return true.

You could have used your ratesAverage function aswell here by first filtering the first array and then sending the filtered array as a parameter.

starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes(arr){
+    let newArr = [...arr];
+    newArr.map(function(a) {
```
ta-web-paris: You don't need to spread the arr to create a new reference, the map function already does that for you

`let newArr = arr.map()`

starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+            duration = parseFloat(a.duration)*60
+        /*  2ND TRY
+        if (a.duration.charAt(1) === "h" && a.duration.charAt(a.duration.length-1) !== "n") {   // if 2nd = "h" and no "n" at the end : hour only
+            a.duration = parseFloat(a.duration)*60
+        }if (a.duration.charAt(1) !== "h") {        // if 2nd is not "h" : minute only
+            a.duration = parseFloat(a.duration)
+        */
+        }else a.duration = (parseFloat(a.duration)*60) + (parseFloat(a.duration.slice(2,8)))    // number of hour * 60 + parseFloat of the "xxmin" part
+    }); return newArr
+}
```
ta-web-paris: Your tests are not passing because even though you created a new array, the objects inside the array are still the same reference.

you should create a copy of the object aswell, a way to do this would be by using the object spread operator:

const copy = {...originalObject}

then modify the duration on copied object

ta-web-paris:
Well done overall, closing the PR, come see me if you have any questions.

PR #1124, opened 2020-01-12
[PAR - WDFT] - Shannon - progress

tranquildoe:
I got Iterations 1-3 done and wanted to send them in for feedback for now.
Do we need to have all the labs finished and correct by Monday?


TatijanaRajcic:
Hey Shannon :) Do you want us to correct your lab now, or do you want to add something to it before we do?

tranquildoe:
Hi Tatijana,
Oh, I thought Louis pushed the rest of it so it should be done? We worked
together.

TatijanaRajcic:
Hey Shannon! Nice work!

Some advice to make your code shorter: 
- For your function orderByYear : you could have used the built-in function localeCompare. Basically, it does what you are doing here:

```
    if (a.year - b.year) {
      return a.year - b.year;
    } else if (a.title > b.title) {
      return 1;
      return -1;
    } else return 0;
```

But shorter :) 

- For howManyMovies, you don't need to write the "else" condition while using filter. That would be enough: 

```
  let filteredMovies = array.filter(function(movie) { return movie.director === "Steven Spielberg" && movie.genre.includes("Drama")) }
  return filteredMovies.length
} 
```

- For your function orderAlphabetically, same thing as the first one: you could use localeCompare for your first step. Or even use sort() without any argument since it does the job. Also, you could chain your higher order functions to make your code shorter.

```
  let newArr = arr.map(obj => obj.title).sort().slice(0,20);
  return newArr;
}
```

Very good work anyway :))

PR #1126, opened 2020-01-12
[BER0120-FT]-Vaso

vasilikibalidou:
I am getting an error at the bonus iteration about the localeCompare but I am not sure how to fix it

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -12,13 +12,29 @@ function orderByYear(movies) {
 }
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
 function howManyMovies(movies) {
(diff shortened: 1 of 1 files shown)


vasilikibalidou:
BER0120-FT

fenrew:
First off: You've made a very nice and clean code. It is short and precise, please keep this awesome work up!:)

So your error is occuring because your orderByYear function compares the titles (movie.title while Jasmine feeds your function this array: [
  { year: 2000, rate: 9 },
  { year: 2000, rate: 8 },
  { year: 1978, rate: 10 },
  { year: 1978, rate: 7 }
]

What happens then is that your movie.title will be undefined since there are no .title in this array ==> "Can't do localCompare of undefined".

Other than this, every function seems quite optimized and beautiful!

PR #1127, opened 2020-01-12
[WD - FT] Alex Perdikakis

Perdikal:


starter_code/src/movies.js
```diff
@@ -1,15 +1,87 @@
...
+    } else if (result.length === 1 && result[0].includes("m")) {
+      return parseInt(result[0]);
+    } else {
+      return (
+        parseInt(result[0].replace(/\D/g, "")) * 60 +
+        parseInt(result[1].replace(/\D/g, ""))
+      );
+    }
+  };
+  return movies.map(movie => ({
```
fenrew: Did you do this to satisfy the Jasmine's "Return a new array"? If so what you do here is to manually creating a deep copy of your movies array. Another way you could solve this is to write: let newMoviesArray = JSON.parse(JSON.stringify(array))

JSON.stringify converts your object into a string, and JSON.parse converts your string to an Object. This leads to a deep copy of your movies array that has it's own memory reference.

PR #1128, opened 2020-01-12
done

JohnnyAL:
I could not for the life of me figure out what number 4 wanted in jasmine when it says "It should return 0 if there is no movie." Can you help to clarify this for me??

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2,14 +2,119 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(movieList) {
+  let sortedMovies = movieList.slice().sort(function(a, b) {
+    if (a.year === b.year) {
+      return a.title.localeCompare(b.title);
+    }
+    return a.year - b.year;
+  });
+  return sortedMovies;
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+
+function howManyMovies(movieList) {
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -2,14 +2,119 @@
...
+        return true;
+      }
+    });
+  return top20;
+}
+
 // Iteration 4: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movieList) {
+  let ratesTotal = movieList
```
fenrew: Just add this between line 48 and 49:   if(movieList.length < 1) return 0

starter_code/src/movies.js
```diff
@@ -2,14 +2,119 @@
...
+}
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes(movieList) {
+  let minutesMovieList = movieList.map(function(movie) {
+    if (!movie.duration.includes("h")) {
+      return {
+        ...movie,
+        duration: Number(movie.duration[0] + movie.duration[1])
```
fenrew: I'm pretty sure that movie.duration[0] + movie.duration[1] is basically string concatenation and then the string is turned into a number. You can instead use something called substring() or slice() to make it look a bit better. However, I may be incorrect here

JohnnyAL:
Haha yea the turnHoursToMinutes function is all over the place. But I made
it work and I was tired lol. But using split definitely makes more sense.

Thanks for the feedback!

PR #1130, opened 2020-01-12
BEFT012020_Akua

akups:


starter_code/src/movies.js
```diff
@@ -2,14 +2,132 @@
...
+  const howManyMovies = movie
+    .filter(function(movie) {
+      if (
+        movie.director === "Steven Spielberg" &&
+        movie.genre.includes("Drama")
+      ) {
+        return true;
+      }
+    })
+    .map(function(movie) {
```
fenrew: This .map function is not necessary and can be removed without affected Jasmine

starter_code/src/movies.js
```diff
@@ -2,14 +2,132 @@
...
+      return a.title.localeCompare(b.title);
+    })
+    .map(function(movie) {
+      return movie.title;
+    });
+
+  if (allMovies.length < 1) {
+    return [];
+  }
+  if (orderAlphabetically.length > 20) {
```
fenrew: This if statement here (if(orderAlphabetically.length > 20)) is not necessary as I'm pretty sure slice will only slice an array up to 20 (slice(0, 20)) and if an array is shorter, it will slice it accordingly. A bit of a clunky explanation but.. If you remove the if statement the code still works perfectly fine with Jasmine:)

starter_code/src/movies.js
```diff
@@ -2,14 +2,132 @@
...
+  });
+  return ratesAverage(dramaMovies);
+}
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
 // BONUS Iteration: Best yearly rate average - Best yearly rate average
+function bestYearAvg(movies) {
+  // I order the movies by years.
+  const moviesOrdered = orderByYear(movies);
```
fenrew: Your orderByYear function requires an array that has the property .title to function as it sorts the array by .title. Jasmine feeds your function with this array that doesn't have .title:
[
      { year: 2000, rate: 9 },
      { year: 2000, rate: 8 },
      { year: 1978, rate: 10 },
      { year: 1978, rate: 7 }
    ];

This causes the final mistake in your code that gives the error code: "Can not do .localCompare of undefined". You'd have to work around this to get it perfect:)

PR #1134, opened 2020-01-12
[PAR WEBFT] Paul Carilllion - Je pense que le code est bon mais deux points ne passent pas dans Jas… [ongoing]

pcarillion:
…mine...


starter_code/src/movies.js
```diff
@@ -2,14 +2,93 @@
...
+
+function turnHoursToMinutes(arr) {
+ let newArr = [...arr];
+ newArr.duration = newArr.map(function(arr){
+   return arr.duration = functionST(arr.duration);
+ }
+ );
+ return newArr;
+}
+
```
ta-web-paris: 1) . Line 86: you don't need to make a copy of the array when you are map, the map function returns a new array.

2) The reason your tests are not passing is because the elements inside the first array are objects, so you end up with a reference to the same objects in the new array.
So in the map, you should create a new object, modify the duration of that new object and return that new object so you end up with a new object that has it's duration modified.

Here's a solution one solution :
```
  return arr.map(function(movie) {
    return {
      ...movie,
      duration: functionST(movie.duration)
    };
  });
}
```

Here's another one with the spread operator :

```
  return arr.map(function(movie) {
    const copy = { ...movie };
    copy.duration = functionST(movie.duration);
    return copy;
  });
}
```

starter_code/src/movies.js
```diff
@@ -2,14 +2,93 @@
...
+  let newArr = arr.filter(function(element) {return element.genre.indexOf("Drama") !=-1})
+  if (newArr.length === 0) {return 0};
+  let sum = newArr.reduce(function(accumulator, currentValue) {
+    if (typeof currentValue.rate !== "number") {currentValue.rate = 0};
+    return accumulator + currentValue.rate
+  }, 0);
+  let average = sum / newArr.length;
+  return Number(average.toFixed(2));
+}
+
```
ta-web-paris: Here you could have used the ratesAverage() function that you created above. You have implemented the same average logic twice :D

starter_code/src/movies.js
```diff
@@ -2,14 +2,93 @@
...
+function orderByYear(arr) {
+  let newArr = [...arr];
+  newArr.sort(function(a, b){ 
+    if (a.year - b.year !== 0) {
+      return a.year - b.year;
+      }
+    else  {
+      if (a.title > b.title) {return 1}
+      else if (a.title < b.title) {return -1}
+      else {return 0};
```
ta-web-paris: return a.title.localeCompare(b.title) would resolve to the same result.

starter_code/src/movies.js
```diff
@@ -2,14 +2,93 @@
...
+  });
+  return newArr;
+}
+orderByYear(movies);
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
 
+function howManyMovies(arr) {
+  let newArr = arr.filter(function(element){
+    return (element.director === "Steven Spielberg") && (element.genre.indexOf("Drama")) != -1
```
ta-web-paris: Have a look at the includes() method on the Array.

ex: 

`return (element.director === "Steven Spielberg") && (element.genre.includes("Drama")`

ta-web-paris:
Everything is good apart for the last tests which are not passing, closing the PR :)

PR #1135, opened 2020-01-12
[PAR-WDFT] Kathleen & Alix ongoing

AlixdeR:
done (there are two solutions each time cause we felt more confortable with different methods!)


TatijanaRajcic:
Hey girls! Great job! Here are some advice:
- orderByYear
Your first solution is more efficient as in the second one you are applying two sorts, so your code has to process twice all the data. It's clearer to use the spread operator [... arr] that to map in order to make a copy of the existent array. To simplify this part of your code...:
```
        if (a.title<b.title) {
          return -1;
        } else if (a.title>b.title) {
          return 1;
        } else {
          return 0;
        };
```
...you could use a function called localeCompare :) it basically does what you do here but shorter
- howManyMovies:
Great solutions! I like the fact that you used both includes and indexOf, as well as arrow functions :) You could save one line by directly returning the result of your functions : 
```
function howManyMovies(arr) {
  return arr.filter(el => (el.director === "Steven Spielberg") && (el.genre.includes("Drama"))).length;
}
```
- orderAlphabetically:
Both of your solutions work but the second one is shorter so preferable :) 

- ratesAverage:
your both map the array so that you end up with an array of only rates ! That works well :) Another way of doing it would be to say in the reduce function that you are only interested in the rates. See the chapter "Sum of values in an object array" here in the doc: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce . If you use this method, don't forget to set an initial value to your function.

- dramaMoviesRate:
You could use one of the function you just created before to make this one shorter:
```
function dramaMoviesRate(array) {
  let dramaMovies = array.filter(movie => movie.genre.includes("Drama"))
}
```
- turnHoursToMinutes
Both of your solutions work but the second one is easier to read!

Great job girls :)

PR #1136, opened 2020-01-12
done [PAR-WDFT] Andrew NB ongoing tati

andrewnunezbibby:
Its weird: For Iteration 6 I am failing a few tests but when I test the function with console logs - in a node shell - i am getting values that tell me i should be passing a few of those tests. My function IS returning an Array(object), it IS a new Array and not the original, and the data is being returned as a number. I'm gonna keep working, but I just wanted to get a PR in

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,15 +1,74 @@
-/* eslint no-restricted-globals: 'off' */
-
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+orderByYear = (moviesArr) => {
+    let sortedMovies = [...moviesArr].sort((a, b) => {
+        if (a.year === b.year) return a.title < b.title ? -1 : 1
+        return a.year - b.year;
+    })
+    return sortedMovies;
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+howManyMovies = (moviesArr) => {
+    let results = moviesArr.filter(movie => {
+        return movie.director === "Steven Spielberg" && movie.genre.indexOf("Drama") !== -1
+    })
+    return results.length
+}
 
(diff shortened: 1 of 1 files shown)


TatijanaRajcic:
Hey Andrew! Your code is really great! Some advice (but it's mostly other way of doing it):

- in your orderAlphabetically function, you could have used map instead of reduce as a last step. I'm saying that because I can see that you are trying to reproduce an array with the accumulator, since you are losing it while using reduce. If you use map, it should be simpler :)

- regarding the bonus: I actually get a empty array while I execute your code. I guess on the second line you meant to apply map on arr and not convertedMovies. Then, one problem you have is that even if you are using map, which creates a copy of the original array, while refering to a object, you refers to the same. See painpoint for iteration 6 that Franck posted: 
https://github.com/frankmarmier/cohort_0120_pain-points_reinforcements/blob/master/pain_points/greatest-movies.md
Therefore, i sligthly changed your code:
```
turnHoursToMinutes = arr => {
    convertedMovies = [];
    let newArr = arr.map(movie => {
        let copyMovy = {...movie}
        let totalTime = 0;
        let duration = copyMovy.duration;
        // console.log(duration);
        let hours = parseFloat(duration.slice(0, (duration.indexOf("h")))) * 60
        let minutes = parseFloat(duration.slice((duration.indexOf(" ") + 1), duration.indexOf("min")))
        // console.log("Hours" + hours + "minutes" + minutes)
        copyMovy.duration = totalTime + hours + minutes
        // console.log(movie.duration)
        // console.log(movie);
        convertedMovies.push(copyMovy);
    })
    // console.log(convertedMovies)
    return convertedMovies;
}
turnHoursToMinutes(movies);
```
Now you are passing most of the tests. You still need to change your code to take care of the special cases, when there are only hours or minutes
Don't hesitate if you have questions ! :)

PR #1140, opened 2020-01-13
[BER - WDFT - 012020] Mohamed Ciwan

ciwan-aram:


starter_code/src/movies.js
```diff
@@ -2,14 +2,106 @@
...
+    return ratesAverage(dramaMovies);
+}
+console.log(dramaMoviesRate(movies));
+
+
+
 
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
+function turnHoursToMinutes(el) {
+    const moviesInMin = el.map(function (el) {
```
DalinaLeila: Hint: you would need to split at the empty space, and turn into a number. (+ replace the string "min" and "h" with empty strings)

Then you would have to multiply the hours by 60 to get the minutes and add that to the minutes you already have. 
Create a shallow copy and replace the old duration with the new one. 

There are many ways to do it of course, this is one of them only :). It's tricky but I'm sure you can figure it out! 💯

ciwan-aram: Thanks for detailed hint! we worked on it again in class today and it is clear now, just need to practice on a couple of different examples.
Vielen Dank :)

PR #1142, opened 2020-01-13
[PAR-WDFT] Camille ongoing

camilledlr:


clarabesset:
Good job for the most part - You're missing a few things on the last iteration, here's a solution - read it and let me know if you want to go over it together :) You can split the function into two main actions: converting the string into a number, removing the "h" and "m"s and having only number, and then converting it all in minutes.

`function convertToMin(hourAndMin) {
  if (hourAndMin.includes("h")) {
    return (
      Number(hourAndMin.substring(0, hourAndMin.indexOf("h"))) * 60 +
      Number(
        hourAndMin.substring(
          hourAndMin.indexOf(" ") + 1,
          hourAndMin.indexOf("m")
        )
      )
    );
  } else
    return Number(
      hourAndMin.substring(hourAndMin.indexOf(" ") + 1, hourAndMin.indexOf("m"))
    );
}

  return movies.map(x => {
    const y = Object.assign({}, x);
    y.duration = convertToMin(String(y.duration));
    return y;
  });
}`

PR #1147, opened 2020-01-13
[PAR-WDFT-0120] - María.MUNIZAGA ongoing

mmunizaga:
Hi, my partner was Sylvie but she told me she would do it later because was late with the other labs and preferred to do this later.


TatijanaRajcic:
Hey Maria! Your code is really great I'm impressed:)
One small thing: for your function dramaMoviesRate you could have used the function you just wrote above (ratesAverage) on the array you get with the filter, so you don't have to write the same code again. 
But honestly it's very good! Congrats!

PR #1149, opened 2020-01-14
[PAR - WD - FT] Jonathan Oreja - ongoing

Jarchiii:
/ Iteration 6: Time Format - Turn duration of the movies from hours to minutes

-> Je ne comprends pas pourquoi je modifie le tableau orignal.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,15 +1,178 @@
-/* eslint no-restricted-globals: 'off' */
 
-// Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+
+
+
+
+
+
+  function orderByYear(movieList) {
+      const newArr = [...movieList]
+      newArr.sort((a,b) => {
+          if (a.year > b.year) {
+              return 1 ;
+          }
+          if (a.year < b.year) {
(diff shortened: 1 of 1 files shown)


Jepassaispar:
Great, your functions are looking very neat !

I just refactorized your function reducer : 

```
function reducer(accumulator, currentValue) {
  if (currentValue.rate == "" || currentValue.rate == undefined) {
    return accumulator;
  } else return accumulator + currentValue.rate;
}
```

I really like the fact that you are creating a lot of functions, it is always great to do so as a developper to be able to re-use them.

if your array doesn't pass the test is because if you spread an array of objects into a new variable, your objects in your new array will be linked to the objects in your old array, ex : 
```
var a = [{name: "cat", age: 9}];
 var b = [...a];
a[0].age=10;
console.log(b[0].age)
``` 
will give you 10

So that's what you could do to compensate that : 
```
function converter(element) {
  var element1 = { ...element };
  element1.duration = convertHour(element1.duration);
  return element1;
}
```
The only thing is that your convertHour function return NaN when the minutes are <10, if you have some time you can correct it but anyway you got it, great job 👍

PR #1159, opened 2020-01-15
[PAR-FT-012020] - Youval BUGAJSKI ongoing

YuvalBu:


TatijanaRajcic:
Hey Youval :) Good job! You used several higher order functions so it's exactly what was expected!
Here are some advice:
- For your function orderByYear, you could have used the function [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare). It basically does the exact same thing you are doing here: 
```
         if(a.title > b.title) return 1
         if(a.title < b.title) return -1
         return 0
```
But it's great that you did it by yourself too :)

- Your function howManyMovies works great! You could have it on one line:
```
  return movies.filter(movie => movie.genre.includes("Drama") && movie.director === "Steven Spielberg").length
}
```

- For orderAlphabetically: same remark as in the first one: have a look at localeCompare :). You could also use sort() because this function on words when we don't specify any parameter. Here a shorter version:

```
  return newArr = arr.map(obj => obj.title).sort().slice(0,20);
}

```

- For ratesAverage one test is failling: when a movie does not have a rate. You have to make sure that you use movies.rate on movies that actually have a rate. Here is your code corrected:

```
function ratesAverage(movies){

    /* for (let i=0; i<movies.length; i++){
        if (isNaN(movies[i].rate)){
            movies[i].rate === 0
        }

        */
   
    let moviesRates = movies.reduce(function(sum, movies){
        if (movies.rate) { // HERE IS THE CONDITION
          return sum + movies.rate
        } 
        return sum // IF THE MOVIE DOES NOT HAVE A RATE, YOU JUST RETURN THE CURRENT ACCUMULATOR
   }, 0);

    let result = moviesRates/movies.length;

    return Number(result.toFixed(2));

}
```

TatijanaRajcic:
- For your function turnHoursToMinutes, your test are failling beause you are modifying the original objects. Indeed: when you use map, you make a copy of the original array. But if this array is full of objects, then modifying the copied array will modify the original objects. In other words, you should also make a copy of the objects themselves.

Here are some modifications. I commented the code : 
```

    let movieDuration = movies.map(movie => {
      let newMovie = {...movie} // MAKE A COPY OF THE OBJECTS
      if (newMovie.duration){ // MAKE SURE DURATION EXISTS
        return newMovie.duration.toString().split(' ') // MODIFY THOSE COPIES AND NOT THE ORIGINAL OBJECTS
      }
      return "0"
    });
    // console.log(movieDuration)
    let timeInMinutes = [];
  
    for (let i = 0 ; i < movieDuration.length ; i++) {
  
      let hours = 0;
      let minutes = 0;
  
      if (movieDuration[i][0].includes("h")) {
        hours = (movieDuration[i][0]).substring(0, movieDuration[i][0].length-1);
      }
  
      if (!movieDuration[i][0].includes("h")) {  
        minutes = (movieDuration[i][0]).substring(0, movieDuration[i][0].length-3);
  
      } else if (movieDuration[i].length > 1) {
        minutes = movieDuration[i][1].substring(0, movieDuration[i][1].length-3)
      };
  
      timeInMinutes.push(Number(hours * 60 + Number(minutes)));
    }
  
    // DOING moviesInMinutes = [...movies] is not enough because you are still referencing to the original objects
    // With the following method you are making a "true copy" of the original array

    let moviesInMinutes = movies.map(movie => {
      let newMovie = {...movie} 
      return newMovie
    });

  
    for (let i=0; i< moviesInMinutes.length; i++) {
      moviesInMinutes[i].duration = Number(timeInMinutes[i]);
    }
  
    // console.log(typeof moviesInMinutes[0].duration)
    return moviesInMinutes;
  }
```

Don't hesitate to write me or come to me if something is not clear :)

Good job!!

PR #1161, opened 2020-01-16
[WEB-BCN-Henriette-Federico-William]

federicogiorgino:


starter_code/src/movies.js
```diff
@@ -1,15 +1,81 @@
...
+  const arr20 = arrSorted.slice(0, 20);
+
+  const titles20 = arr20.map(function(movie) {
+    return movie.title;
+  });
+  return titles20;
+}
+
 // Iteration 4: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(arr) {
```
tawebbcn: First thing, you can check if the array passed to the function is empty or not (It should return 0 if there is no movie):
if(!arr.length) return 0;

starter_code/src/movies.js
```diff
@@ -1,15 +1,81 @@
...
+  const arr2 = [...arr];
+  const ratesArr = arr2.map(function(movie) {
+    return movie.rate;
+  });
+
+  let rateTotal = ratesArr.reduce((sum, movie) => {
+    return sum + movie.rate;
```
tawebbcn: Here, instead of the movie element, you already have the rate of the movie (because you created a new ratesArr above with only the rates).
So you would have:
 let rateTotal = ratesArr.reduce((sum, rate) => {
    return sum + rate;
}

In addition, one of the tests requires to return a result even if one of the movies does not have rate, so you can add a check inside the reduce method to check if "rate" exists or not:
if(!movie.rate) { return sum; }
else { return sum + movie.rate; }

starter_code/src/movies.js
```diff
@@ -1,15 +1,81 @@
...
+  }, 0);
+
+  let averageRate = (rateTotal / ratesArr.length).toFixed(2);
+  //   let averageRateRounder = Math.round(averageRate)
```
tawebbcn: to return a number rounded to 2 decimals, you can do:
let averageRate = Number((rateTotal / ratesArr.length).toFixed(2));
return averageRate;

starter_code/src/movies.js
```diff
@@ -76,6 +81,18 @@ function ratesAverage(arr) {
 }
 
 // Iteration 5: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(arr) {
+  const arrDram = arr.filter(e => {
+    return e.genre.includes("Drama");
+  });
+
```
tawebbcn: To check if there is no Drama movie (one of the tests):
if(!arrDram.length) {
  return 0;
}

starter_code/src/movies.js
```diff
@@ -76,6 +81,18 @@ function ratesAverage(arr) {
...
+  let rateTotal = arrDram.reduce((sum, movie) => {
+    return sum + movie.rate;
+  }, 0);
+
+  return rateTotal / arrDram.length;
+}
```
tawebbcn: same than previous iteration, you round the number return to 2 decimals with the .toFixed() method, and you convert the whole thing to a Number type:
return Number((rateTotal / arrDram.length).toFixed(2));

PR #1162, opened 2020-01-16
[BCN-FTWD-Valentin/Abel]

valentin-muller:


valentin-muller:
Didn't use Jasmine but the code should work though.

tawebbcn:
It's a bit hard to give you proper feedback for this exercise since you didn't use Jasmine, most of the lab don't pass the test...
I see that you have a good logic but it's more complicated to evaluate you this way
For next LABs, the use of Jasmine is not an option, specially if you want to receive proper feedback.

PR #1163, opened 2020-01-16
[BCN WDFT Rodrigo & Jorge] Initial commit

rodmartinezmedina:


starter_code/src/movies.js
```diff
@@ -1,12 +1,92 @@
...
+			}
+			else if (a.title > b.title) {
+				return 1;
+			}
+			else {
+				return 0;
+			}	 
+		});
+	const first20 = newArray.slice(0, 19);
+
```
tawebbcn: The slice( ) method doesn't include the last given element, so here it would be newArray.slice(0, 20);

starter_code/src/movies.js
```diff
@@ -73,6 +73,10 @@ function ratesAverage (arr) {
 	
 	const ratesArr = [];
 
+	if (arr = []) {
+		return 0;
+	}
+
```
tawebbcn: to check if an array is empty, you want to check if length of array is equal to 0
if (arr.length === 0) 
or 
if (!arr.length)

starter_code/src/movies.js
```diff
@@ -93,25 +93,28 @@ ratesAverage(movies);
...
-		if (rate => 8) {
-
-		};
+  for (let i = 0; i < sortDrama.length; i++) {
+    dramaRatesArr.push(sortDrama[i].rate);
+  }
 
+  const rateSum = dramaRatesArr.reduce((acc, num) => {
+    return acc + num;
+	}, 0);
```
tawebbcn: here you can check, inside the reduce method, if "num" exists (if movie rate exists)

		if(!num) {
			return acc;
		} else {
			return acc + num;
		}

starter_code/src/movies.js
```diff
@@ -93,25 +93,28 @@ ratesAverage(movies);
 
 
 // Iteration 5: Drama movies - Get the average of Drama Movies
+//Can't solve more. Getting errors. Can't figure out.
+function dramaMoviesRate(arr) {
+  const sortDrama = arr.filter(function(e) {
+    return e.genre.includes("Drama");
+  });
```
tawebbcn: here you can return 0 if the drama movies array (sortDrama) is empty:
if(!sortDrama.length) return 0;

starter_code/src/movies.js
```diff
@@ -116,5 +114,19 @@ dramaMoviesRate(movies);
 
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes(arr) {
```
tawebbcn: The tests ask you to return a new array, so here you could use the map() method to loop over the original array, since this method creates a new array.

PR #1164, opened 2020-01-16
[BCN-FTWD-012020] Stefan Berkenhoff - David Bronne 

StefanK2ff:


starter_code/src/movies.js
```diff
@@ -2,14 +2,120 @@
...
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
 
+function howManyMovies(movieArray, director, type) {
+    const filterResults = movieArray.filter(function(movie) {
+        return movie.director === director && movie.genre.indexOf(type) >= 0 // optional: .includes
+    });
+    //console.log(filterResults);
+    return filterResults.length
+}
+
```
tawebbcn: Here, you should pass only the movie array as argument of the function and filter by returning only if director = 'Spielberg' amd type='drama'

starter_code/src/movies.js
```diff
@@ -2,14 +2,120 @@
...
+    let sortedMovies = [...movieArray];
+    sortedMovies.sort(function (a, b) {
+            if (a.title > b.title) {
+                return -1
+            } else if (a.title < b.title) {
+                return 1
+            } else {
+                return 0
+            }
+        }
```
tawebbcn: Here you sort the array inversely alphabetically, you should invert greater and lower and it works

starter_code/src/movies.js
```diff
@@ -2,14 +2,120 @@
...
+            newDuration = parseInt(hours)*60
+        }
+        
+        movie.duration = newDuration;
+        console.log((typeof newDuration) + newDuration);
+        return movie
+    })
+    return moviesMinDuration
+}
+
```
tawebbcn: If you return a copy of the movie, it should be working

StefanK2ff:
> Good job in general, Be careful to read well the indications of what you have to do to pass the test properly but good logic in general :)

Did an updated with the learning from today: all green now.

PR #1165, opened 2020-01-16
[BCN-FTWD-012020] Neha Khachhibhoya - Jean Pierre Hrubant

annkay108:
Still, some work to do but maybe we will get it done till tomorrow.

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,15 +1,108 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies){
+
+    let filmsByYear = [...movies]
+    filmsByYear.sort(function(a, b){
+      if (a.year < b.year) {
+        return -1
+      }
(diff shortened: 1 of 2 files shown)


starter_code/src/movies.js
```diff
@@ -1,15 +1,108 @@
...
+    let drama = movies.filter(function(e){
+        return e.genre.includes("Drama");
+      })
+    return ratesAverage(drama);
+}
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
+function turnHoursToMinutes(movies){
+  let hoursToMinutes = [...movies];
+  let duration = hoursToMinutes.map(function (e){
```
tawebbcn: Here you're making a shallow copy with only duration so it takes the array points to the original one and the test reads it as the original

tawebbcn:
You guys did a very good job! The logic is good and you passed most of the tests. The only thing i can tell you guys is to name your variables in a more understandable or meaningful name. For example, instead of naming an array "duration" (that looks like its a single duration) maybe you can call it **durationArr** and same applies to other vars. 

Very good job :)

PR #1166, opened 2020-01-17
{BCN-WEB-FT]

JuiceDrinker:


JuiceDrinker:
Adi & Val

tawebbcn:
Good job, Remember to put both name on the title of the pull request ;)
For the last iteration you should return a copy of each movie inside the new array and it should work !
Well done !

PR #1169, opened 2020-01-18
done

fhoman:


starter_code/src/movies.js
```diff
@@ -1,15 +1,259 @@
...
+    }
+
+    return sumAverage + currentvalue.rate / arr.length;
+
+}
+
+const ratesAverage = function ratesAverage(movieArr) {
+
+
+    let averageNumber;
```
LloydC: You can also write it in one line:  let averageNumber = movieArr.reduce(calculateAverageRating, 0)

fhoman: oke, thanks

PR #1170, opened 2020-01-20
done

MichielSnoek:


starter_code/src/movies.js
```diff
@@ -2,14 +2,103 @@
...
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
 
+const orderAlphabetically = function(arr) {
+    let arrTwo = JSON.parse(JSON.stringify(arr))
+    arrTwo.sort(function(a,b) {
+        if (a.title > b.title) return 1
+        if (a.title < b.title) return -1
+    })
+
+    let newArr = [];
```
LloydC: Another valid approach is let newArr = arrTwo.map(eachMovie => eachMovie.title).slice(0, 20);

Since we know there are more than 20 movies inside data.js it's ok to do it that way as well.

PR #1214, opened 2020-01-23
[LIS JAN 20 Aline Pasetti]

alinepasetti:
missing the last and bonus :(


alinepasetti:
Regarding 6 - "The duration of the movie should return the correct number for really long movie - 341 minutes" should be checked, tests worked on my side

PR #1223, opened 2020-01-24
[LIS JAN 20 - Manuel Rocha]

TemidoRocha:


soniameller:
Hey Manuel! This is the interface where we can communicate for lab doubts. You can tag us @ironhackLis-webTeam so that we receive a notification.

About the code:
```
  var s = arr
    .map((x, i) => {
```
   Here the x of the first parameter  represents the item, you don't need to use the array and the index to access it.
```
      if (arr[i].duration === undefined) {
        return null;
      } else return arr[i].duration.split(' ');
    })

```
   In the following map, the z represents each element on the array, so mapping inside is not correct because z is no longer an array
```
    .map(z => z.map((y, i) => Number(parseInt(y))))
    .map(m => {
      return m.reduce((a, b) => a * 60 + b);
    });
  const t = [...arr];
  for (let i = 0; i < t.length; i++) {
    t[i].duration = s[i];
  }

  console.log(t);
  return t;
}
```

This is an example of the value of z in the map(z)
Captura de Pantalla 2020-01-28 a la(s) 09 41 35

If you try to apply a map to z, the following is the result:

Captura de Pantalla 2020-01-28 a la(s) 09 42 07

From there, the function is no longer working. I would recommend you to test the result of every part of the  functions you code while you are typing it so you have a better understanding of the values you are getting. Don't worry too much about this, it takes time! 🙃

[Here the solution for this lab](https://gist.github.com/josecarneiro/ae551463d49a8b98047ff1b26f7cbde2)

TemidoRocha:
Thanks a lot.

I’ll try it again.

Best regards,
Manuel Rocha

PR #1242, opened 2020-02-08
BER WDPT Camille

camilleory:


starter_code/src/movies.js
```diff
@@ -2,14 +2,106 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(arr) {
+    let sortedArray = arr.sort((a, b) => {
```
numbing: its better to make a copy of array and work on that one so we dont modife the original array ,
like this : let copyArray = arr.slice()

starter_code/src/movies.js
```diff
@@ -2,14 +2,106 @@
...
+    let moviesDurationInMinutes = movies
+    for (let i=0; i<movies.length; i++){
+      moviesDurationInMinutes[i].duration = newDuration[i]
+    }
+    return moviesDurationInMinutes
+    }
+    
+    turnHoursToMinutes(movies)
+
+    // It doesn't pass the test on SpecRunner but I tested in in Repl where it seems ok? --> https://repl.it/@camilleory/movies
```
numbing: woow nice job, you did great job

PR #1247, opened 2020-02-08
BER PT Olga

Olga-OK:


starter_code/src/movies.js
```diff
@@ -1,15 +1,125 @@
...
+    console.log (movies)
+}
+*/
+
+// IT IS ALIVE!
+function howManyMovies(arr) {
+    let filteredArray = arr.filter(function (obj) {
+
+            if (obj.director === "Steven Spielberg") && obj.genre.includes("Drama")) {
+            return true;
```
numbing: you could use && operator here to write little bit cleaner code but its nothing wrong with what you wrote

starter_code/src/movies.js
```diff
@@ -1,15 +1,125 @@
...
+  return sum + el
+    }, 0)
+
+    let averageDramaRatings = sumRates/movieRates.length
+
+    console.log(averageDramaRatings)
+    return averageDramaRatings
+}
+
+dramaMoviesRate (movies)
```
numbing: really nice job, you doing great

PR #1251, opened 2020-02-08
[BER-WDPT-202002]) - Sonja L.

Asphalthexe:
stuck at iteration 4

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -12,7 +12,7 @@ function orderByYear(moviesArray) {
                 return 1;
             } else {
                 return -1;
-            } 
+            }
         }
         return a.year - b.year
     })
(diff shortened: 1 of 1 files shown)


starter_code/src/movies.js
```diff
@@ -1,15 +1,148 @@
...
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+function howManyMovies(steveArray) {
+    let result = 0
+    steveArray.filter(function (dramaSteve) {
+        if (dramaSteve.genre === "Drama" && dramaSteve.director === "Steven Spielberg") {
+            result++;
+        } else {
+            return 0;
+        }
+    })
```
numbing: if here we use .length we can get the length of new Array that filter just created so we dont need result but its good way of thinking nice

PR #1277, opened 2020-02-22
[LIS - WDPT - FEB] Stefano

oddouu:
Strange: Iteration 6 shows a spec failure for "The duration of the movie should be a number" when looked from the Spec list. However, once I click on the spec in red, it becomes green. 
@ironhack-LIS-PTWD Do you know why this would happen?


ironhack-LIS-PTWD:
Hey, I'm sorry but not too sure why this would happen either. 
But not to worry, your code is working 👍 Good job!

PR #1281, opened 2020-02-23
[LIS-PT-FEB2020]- André R.

andreElCoder:
@ironhack-LIS-PTWD

I am changing the original movie vector on turnHoursToMinutes function.
I have tried to use the spread operator, however not able to solve it. 
I know I am changing this property inside the map function.
Please help me :) !

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -1,15 +1,86 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(arrayUnsorted){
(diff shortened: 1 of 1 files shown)


ironhack-LIS-PTWD:
Hey Andre,
Hopefully you managed to find the solution by now but if not, please refer to this solution:

`function turnHoursToMinutes(array) {
    return array.map(movie => {
        const durationAsString = movie.duration;
        let duration = 0;
        for (let value of durationAsString.split(' ')) {
            const number = parseInt(value);
            if (value.includes('h')) {
                duration += number * 60;
            } else if (value.includes('min')) {
                duration += number;
            }
        }
        return {
            ...movie,
            duration
        };
    });
}`

Thanks -Lissy

PR #1283, opened 2020-02-23
LIS-PTWD-FEB-Catarina & Bruno

ctlaranjeiro:
@ironhack-LIS-PTWD Help! 
Can't complete Iteration 6, had a problem when trying to convert strings to numbers. It would return NaN if I didn't set it to filteredDurationNumbers[0].map(Number). And can't figure out how to solve the Bonus

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2,14 +2,213 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(array) {
+    let sortedByYear = [...array].sort((a, b) => {
+            if(a.year < b.year) {
+              return -1
+            }
+            else if(a.year > b.year) {
+              return 1
+            }
+            else if(a.year === b.year) {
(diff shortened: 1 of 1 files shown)


ironhack-LIS-PTWD:
Hey, 
Did you try using parseInt() :)

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt

Thanks, 
Lissy

ctlaranjeiro:
Hey Lissy!

Yeah, we both did. It still returned NaN. I tried to give it another go today, but there has to be something we are doing wrong, 'cause it still returns NaN.

I tried to use substring() to eliminate the "h", the space, and the "min", so it would return me an array with two strings, like ["2", "35"]. Then i tried Number, parseInt, parseFloat, and they all returned NaN.

Will we correct this in class?

See you in a bit,
Catarina


Obter o Outlook para Android<https://aka.ms/ghei36>
________________________________

PR #1285, opened 2020-02-25
LIS - WDPT - FEB - Julia & Wilson 

juliamendesc:
@ironhack-LIS-PTWD

Hi girls, I am submitting my lab in partnership with Wilson, but I must confess I still have trouble understanding the parseInt which Guilherme mentioned could also be used in the turnHoursToMinutes(). Do you have the solution with parseInt so I can try first to see if I understand the syntax?

Thank you!

### DIFF
--- a/starter_code/src/movies.js
+++ b/starter_code/src/movies.js
@@ -2,14 +2,138 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(array){
+    let newDB = JSON.parse(JSON.stringify(array));
+    newDB.sort((a,b) => {
+      if (a.year === b.year){
+        if (a.title < b.title){
+          return -1;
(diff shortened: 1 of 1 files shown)


ironhack-LIS-PTWD:
Hi Julia,
Good job for submitting :) 
Here is the solution for the iteration, hopefully if you go through it slowly it will make more sense :)
`
function turnHoursToMinutes(array) {
    return array.map(movie => {
        const durationAsString = movie.duration;
        let duration = 0;
        for (let value of durationAsString.split(' ')) {
            const number = parseInt(value);
            if (value.includes('h')) {
                duration += number * 60;
            } else if (value.includes('min')) {
                duration += number;
            }
        }
        return {
            ...movie,
            duration
        };
    });
}`

Let me know if you need anything else 
-Lissy

PR #1300, opened 2020-03-27
BER-FTWD:Yosra.Yousfi

yosrayousfi:


src/movies.js
```diff
@@ -1,15 +1,95 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+const orderByYear = movies =>
+  [...movies].sort((firstEl, secondEl) => {
+    if (firstEl.year > secondEl.year) return 1;
```
fenrew: You actually don't have to return 1 or -1, etc. what you instead can write is a number that is either greater or lower than 0. With this is mind what you can do is just return this:

.sort(
 if (firstEl.year === secondEl.year) return firstEl.title.localeCompare(secondEl.title);
 return firstEl.year - secondEl.year
)

This is because firstEl.year - secondEl.year will return a negative number, a positive number, or 0

src/movies.js
```diff
@@ -1,15 +1,95 @@
...
+  return Math.round(avgRate * 100) / 100;
+  // Or return +avg.toFixed(2);
+};
 
 // Iteration 5: Drama movies - Get the average of Drama Movies
+const dramaMoviesRate = movies =>
+  ratesAverage(movies.filter(movie => movie.genre.includes("Drama")));
 
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
+const durationConversion = durationString => {
```
fenrew: Niiice!

PR #1302, opened 2020-03-27
BER-FTWD-Manuel

mtoscano91:
Couldn't finish last iteration. Got stuck updating the array with my new calculated array

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -85,20 +85,19 @@ function dramaMoviesRate(array) {
     
 function turnHoursToMinutes(array) {
     //Pick duration to modify
-   let splitDuration = array.map(function(movie){
+   let splitDuration = JSON.parse(JSON.stringify(array)).map(function(movie){
      //Split with H and check if it is an hour or min
      if (!movie.duration.split('h')[0].includes('min')) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,15 +1,110 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(array) {
+    //Create new Array and sort it
+    let newArray = array.slice().sort(function (a,b) {
+        return a.year - b.year || a.title.localeCompare(b.title);
```
fenrew: Clever!

src/movies.js
```diff
@@ -1,15 +1,110 @@
...
+
 
 // Iteration 4: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(array) {
+    // Check if array empty
+    if (array == 0) return 0;
+    //Get only movie rates
+    let onlyRate = array.map(function(movie){
+        return movie.rate || 0;
```
fenrew: This works, though I'd personally use filter for this as it is exactly what filter excels at

PR #1303, opened 2020-03-27
BER-FTWD-Olha

olhadanylchenko:
bonus is super complicated. hopefully, will finish it during weekend.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -63,3 +63,41 @@ const turnHoursToMinutes = arr => {
   return result;
 };
 // BONUS Iteration: Best yearly rate average - Best yearly rate average
+
+const bestYearAvg = arr => {
+  if (arr.length === 0) return null;
+
+  // making an array of unique years
+  const arrOfYears = arr.map(movie => movie.year);
+  const uniqueYears = arrOfYears.reduce((acc, curval) => {
+    if (!acc.includes(curval)) {
+      return [...acc, curval];
+    } else return acc;
+  }, []);
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,15 +1,103 @@
...
+};
+// }
+// ratesAverage(arr)
 
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
+const handleTime = string => {
+  if (!string.includes("h")) {
+    return parseInt(string.replace(/\D/g, ""));
+  } else if (!string.includes("min")) {
+    return parseInt(string.replace(/\D/g, "")) * 60;
```
fenrew: You can also use a .match here. It would be something along the lines of:
.match(/(\d+h)\s(\d+min))

src/movies.js
```diff
@@ -1,15 +1,103 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+const orderByYear = arr => {
+  const sorted = [...arr].sort((a, b) => (a.year > b.year ? 1 : -1));
```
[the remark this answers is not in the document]
olhadanylchenko: thank you Markus! I know, i was lazy and wanted to proceed asap to the hardest one, so i half-assed this one and stopped caring when it passed the tests (magically). I'll be more careful next time :)

src/movies.js
olhadanylchenko: thank you! i don't really know much about regex, i just googled and played around until it worked.

PR #1304, opened 2020-03-27
BER-FTWD-Phillip

phpaul89:
29.3.20:
- finished Bonus Iteration
- cleaned code

28.3.20:
- rewritten iteration 6

27.3.20:
- w/o Bonus
- Iteration 6 implements a helper function to convert duration format into minutes-only -> Jasmine seems to have problems with this solution as the following tests couldn't be processes:

"Should return an array --> fail
Should return a new array, not update the original one --> fail
The duration of the movie should be a number --> fail"


src/movies.js
```diff
@@ -1,15 +1,3123 @@
...
+    let counter = 0;
+    const spielbergsFinest = arr.filter(function (movie){
+        if(movie.director == "Steven Spielberg" && movie.genre.includes("Drama")){
+            return counter++;
+        }
+        
+        return 0;
+    });
+  
+    return counter;
```
fenrew: You can also return spielbergsFinest.length ;)

src/movies.js
```diff
@@ -1,15 +1,3123 @@
...
+    return parseFloat((sumRate/arr.length).toFixed(2));
+}
+
 // Iteration 5: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(arr){
+
+    const dramas = arr.filter(movie => movie.genre.includes("Drama"));
+
+    return ratesAverage(dramas);
```
fenrew: Nice!

src/movies.js
```diff
@@ -1,15 +1,3123 @@
...
+      let newDuration = 0;
+      let durationType = movie.duration;
+      const checkH = durationType.includes("h"); //returns true or false
+      const checkMin = durationType.includes("min"); //returns true or false
+    
+        if(checkH && checkMin){
+          let hMin = durationType.split(" ");
+          let newH = hMin[0].replace("h","");
+          let newMin = hMin[1].replace("min","");
+          newDuration = (Number(newH)*60) + (Number(newMin));
```
fenrew: You actually dont have to replace "h" and "min", parseInt does this far you:
parseInt("30 min") returns 30

PR #1307, opened 2020-03-28
BER-FTWD-Nate

mcfwesh:


src/movies.js
```diff
@@ -1,15 +1,88 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies){
+    movies = JSON.parse((JSON.stringify(movies)))
+     return movies.sort((a, b)=>{
+      if(a['title'] < b['title']){
+        return -1;
+      }
+      if (a['title'] > b['title']){
```
fonsogms: you didnt need to do that many condition you could have done something like this:
```
if (a.year===b.year){
        }
        return a.year-b.year;
```
That way you could save a lot of code :)

src/movies.js
```diff
@@ -1,15 +1,88 @@
...
+      }
+     }).sort((a, b)=> a['year'] - b['year'])
+    }
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+function howManyMovies(movies){
+    return movies.filter(movie =>{
+      return (movie['director'] === "Steven Spielberg" && movie['genre'].includes('Drama'))
+    }).length
+  }
```
fonsogms: nice job

src/movies.js
```diff
@@ -1,15 +1,88 @@
...
 
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies){
+    movies = JSON.parse((JSON.stringify(movies)))
```
fonsogms: nice approach!
A few comments:
1. You didnt need to JSON.parse movies :)
2. Also here you could have saved a lot of conditions with localeCompare again 
3. Since you  also return only and array of strings you can map the string from the beginning and sort the mapped array :)
example:
```
  let titles = movies.map(function(movie){
        return movie.title;
    });
    titles.sort(function(a,b){
        return a.localeCompare(b);
    });
    return titles.slice(0,20);
```

PR #1308, opened 2020-03-28
BER-WDFT-Marcia

marciafurtado:


src/movies.js
```diff
@@ -2,14 +2,139 @@
...
 // Iteration 4: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(arr){
+    if(arr.length === 0) return 0;
+
+
+    const sumOfallRates = arr.reduce(function(acc, val){
+        if(val.rate !== undefined){  // I thought this test: "Return average even 
+            return acc + val.rate;   // if one of the movies does not have rate!" would be solved here.
+        }
```
fonsogms: Here you needed to add an `else { return acc}` to return acc for the movies with no rates :)

src/movies.js
```diff
@@ -2,14 +2,139 @@
...
+      return parseFloat(averageRate.toFixed(2));
+  }
+  
+  dramaMoviesRate(movies);
+
 // Iteration 6: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes(arr){  
+
+    let newMoviesArray = arr.slice(); // here, arr is already returning an object, I don't understand why
```
fonsogms: It does return and array, but when you check type of newMoviesArray it tells you object cause technically and Array is an object

fonsogms: The problem with the "return an array" in jasmine is because you have an error before returning during the test :)

fonsogms: imagen

PR #1309, opened 2020-03-28
BER-FTWD-Chantel

chantel-taki:
Currently stuck on iteration 6.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,58 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
-
+function orderByYear(arr) {
+    const newArray = [...arr];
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,15 +1,58 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
-
+function orderByYear(arr) {
+    const newArray = [...arr];
+    newArray.sort(function (a, b) {
+        if ( a.year === b.year) {
```
fonsogms: here you could have made your life easier with localeCompare().
Example:
```
  let ordered = movies.slice().sort(function(a,b){
        if (a.year===b.year){
        }
        return a.year-b.year;
    })
    return ordered;
```

PR #1314, opened 2020-03-29
[PAR - WDFT - 0320] Anne Patana & Florian Jomain ongoing

florian-jomain:
…r programming.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2967,3 +2967,5 @@ let movies = [
       "rate": 8
     }
   ]
+
+  module.exports = movies;
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,121 @@
-/* eslint no-restricted-globals: 'off' */
+// const movies = require("./data.js")
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
-// Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+function orderByYear(arr) {
+    const copyOfArray = [...arr];
+    copyOfArray.sort((a, b) => {
+        if (a.year === b.year) {
+            return (a.title).localeCompare(b.title)
+            } else {
+            return a.year - b.year;
+            }
+    });
+    return copyOfArray;
+}
+
+// // // // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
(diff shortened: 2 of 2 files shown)


TatijanaRajcic:
Hey guys! Nice job !! :))) Here are some remarks to refactorize your code:
- Iteration 1 : you could use a ternary operator to make you code shorter:

```
function orderByYear(arr) {
    const copyOfArray = [...arr];
    copyOfArray.sort((a, b) => a.year === b.year ? (a.title).localeCompare(b.title) : a.year - b.year);
    return copyOfArray;
}
```

- Iteration 2: you do not need the "else" here. Filter will only keep the elements for which the callback functions returns true. So you don't care about the others. So your function becomes:

```
function howManyMovies(arr) {
    const spielbergMovies = arr.filter(movie => {
        if (movie.director === "Steven Spielberg" && movie.genre.includes("Drama")) {
            return true;
        } 
    });
    return spielbergMovies.length;
}
```

- Iteration 3: sort works 2 ways: without an argument, or with a callback function as a argument. Without any argument, the sort functions sorts in ascending alphabetical order. If you want a custom sorting, you need to specify a comparison function as a callback. In your example, you could have done the map first, so you would have end up with an array of only titles. Therefore, you could have use sort WITHOUT any argument, since you want to sort the titles by alphabetical order. Here, you are obliged to specify a custom sorting callback function, because you want to access the "title" key of every movie. But if you do things the other way, you don't have to. Therefore, the code can simply  be:

```
function orderAlphabetically(moviesArr) {
  return moviesArr
    .map(movie => movie.title)
    .sort()
    .slice(0, 20);
}
```

Here, you could use slice instead of using a loop. You way works, this one is only shorter :)

- Iteration 4: again, you do not need the "else" statement inside your filter callback. Another way of doing it, without having to do a filter before the reduce, would be to have some conditional statements inside the reduce itself. This way works too:

```
  let myArr = [...array]
  if (!array.length) return 0
  let myTotal = myArr.reduce(function(acc,movie) {
    if (movie.rate != null) {
      return acc + movie.rate
    } else {
      return acc
    }
  },0)
  return parseFloat((myTotal / myArr.length).toFixed(2))
}
```

- Iteration 5: do not need the "else" statement :) Nice use of the ratesAverage function :)

- Iteration 6: well done!

Keep up the good work guys !

PR #1316, opened 2020-03-29
BER-FTWD-Robert

robertbozsik:


src/movies.js
```diff
@@ -2,14 +2,176 @@
...
+
+// 3.
+// const durationMin = splitHClearNumber * 60 + splitMinClearNumber;
+// console.log(durationMin);
+
+// 4.
+function turnHoursToMinutes(moviesArray) {
+    const minutes = moviesArray.slice().map(function (minute) {
+        const splitH = minute.duration.split('h');
+        const splitHClearNumber = Number(splitH[0]);
```
Jeff-Moraes: Notice that some of the movies only have the minutes, they have less than 60 minutes, so when you split at this string and use the Number method on that you have NaN.

PR #1317, opened 2020-03-29
PAR-FT-032020 Yuval Bugajski

YuvalBu:


TatijanaRajcic:
Hey Yuval! Thanks for the lab! It's really good!
Here are some remarks:
- Iteration 1: you could use a ternary operator to make your code shorter:
```

	// We use the ... operator to create a deep clone of a given array
	let newArr = [ ...movies ];
	newArr.sort((a, b) => a.year === b.year ? a.title.localeCompare(b.title) : a.year - b.year);
	return newArr;
}
```

- Iteration 2 and 3: perfect ! 
- Iteration 4 : you don't have to do a loop to modify the original array by adding a rate, when there is not. You could write some conditions in your reduce function directly:

```

	let averageRate = movies.map((movie) => movie.rate).reduce((a, b) => {
		if (b) return a+b
		else return a
	}, 0 ) / movies.length;

	return Number(averageRate.toFixed(2));
}
```

Btw, instead of writing a and b as the parameters of the reduce function, I would advise you to give them clear name. Such as total and currentRate

```

	let averageRate = movies.map((movie) => movie.rate).reduce((total, currentRate) => {
		if (currentRate) return total + currentRate
		else return total
	}, 0 ) / movies.length;

	return Number(averageRate.toFixed(2));
}
```

- Iteration 5: perfect
- Iteration 6: works well :)
- Bonus: if you want to give it a try; don't hesitate to reach out :)

Good job!

PR #1319, opened 2020-03-30
done

dionifss:
Please really check the point 6, I would like to review it with a TI to understand why it doesnt work, because with console logs it seems it works ;)


dionifss:
all finally done ;)

PR #1321, opened 2020-03-30
[PAR WDFT 032020] William H. & Sam L. ongoing

williamhzo:


TatijanaRajcic:
Hey guys! What a great work! You passed every test so I'll only focus on some small improvements:
- Iteration 1: you could use the ternary operator to make you code shorter:

Instead of:
```
  if (a.year !== b.year) {
    return a.year - b.year;
  } else {
  }
};
```

you could have: 

```
  return a.year !== b.year ? a.year - b.year : a.title.localeCompare(b.title);
};
```

- Iteration 2: perfect!
- Iteration 3: you could chain all the HOF. But it's fine like that too :)
- Iteration 4: i simplified your code using the ternary operator, and changing the way you use reduce a little bit:

```
  let sumRates
  if (array.length > 0) {
    sumRates = array.reduce((result, item) => item.rate ? result + item.rate : result, 0)
  } else {
    return 0;
  }
  return +(sumRates / array.length).toFixed(2);
}
```

Because now you are basically using reduce, but not for the right reason. The purpose of reduce is to "reduce" an array, to a single value. Right now, you are using reduce but basically a for loop would do the trick with your logic (i.e. incrementing sumRates at every iteration). So either replace your reduce with a for loop, or change your reduce function so that you actually use its return value :)

- iteration 5: perfect use of the already defined function :)
- Iteration 6: nice use of deep clone! As I told you Sam, you are obligated here to use a deep clone, as you have an array of objects, so making a copy of the array is not enough.
- Last bonus: really nice guys !!!

Really good work!

PR #1343, opened 2020-04-02
[BCN-FT-032020] Hélène Stephan

HeleneStephan:
That's in progress. I don't know yet if I'll have time to finish it today.

I don't understand why there is an error on the iteration 4 because with my own tests it seems to work so I'll need to check Jasmine tests if I manage to find enough time to do so !


tawebbcn:
Very good job overall 👏
Well done for doing the 1st bonus, please refer to Uros' solution for more details :)

PR #1344, opened 2020-04-02
donde 5

alpagor:
Hecho 4 junto con Ariadna!

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,14 +1,131 @@
 /* eslint no-restricted-globals: 'off' */
 
-// Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+// Iteration 1: Ordering by year - Order by year, ascending (in growing order) 
+
+
+function orderByYear (movies) {
+
+
+    const newMovies = [...movies];
+    
+    newMovies.sort(function(movieA, movieB){
+        if(movieA.year < movieB.year){
+            return -1;
+        }
+        if(movieA.year > movieB.year){
+            return 1;
+        }
+        if(movieA.year === movieB.year){
+            if(movieA.title < movieB.title){
+                return -1
+            }
+            if(movieA.title > movieB.title){
+                return 1
+            }
+            if(movieA.title === movieB.title){
+                return 0
+            }
+            return 0;
+        }
+    });
+    return newMovies
+    }
+
+
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
 
+function howManyMovies(movies) {
+	const dramaMovies = movies.filter(function(movie){
+		
+			if (movie.genre.includes("Drama") && movie.director === "Steven Spielberg"){
+			return true 	
+			} 
+	});
+	return dramaMovies.length
+}
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,15 +1,132 @@
...
+            if(movieA.title < movieB.title){
+                return -1
+            }
+            if(movieA.title > movieB.title){
+                return 1
+            }
+            if(movieA.title === movieB.title){
+                return 0
+            }
+            return 0;
```
tawebbcn, line 28: This return is redundant as you are already considering all possible cases and returning 0 when titles are the same. So you could remove it.

src/movies.js
```diff
@@ -1,15 +1,132 @@
...
+			if (movie.genre.includes("Drama") && movie.director === "Steven Spielberg"){
+			return true 	
+			} 
+	});
+	return dramaMovies.length
+}
+
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(movies) {
```
tawebbcn, line 51: This function runs good and it is easy to understand. However, to improve it performance wise you could do the map before the sort. That way you will be sorting only the list of titles and not all of the movie objects.

src/movies.js
```diff
@@ -1,15 +1,132 @@
...
+                    return 0
+                }
+    }).map(function(movie){
+    return movie.title;
+    });
+    
+    return alfaOrder.slice(0, 20);
+    
+    }
+    orderAlphabetically(movies);
```
tawebbcn, line 74: Remember that for Jasmine Tests you don't need to run the functions, Jasmine will do that for you and check on different cases :)

src/movies.js
```diff
@@ -1,15 +1,132 @@
...
+                    return -1
+                }
+            
+                if(movieA.title > movieB.title){
+                    return 1
+                }
+                if(movieA.title === movieB.title){
+                    return 0
+                }
+    }).map(function(movie){
```
tawebbcn, line 67: Good job on joining methods together, however, if they have multiple lines, we would recommend to separate them for better readability.

src/movies.js
```diff
@@ -1,15 +1,132 @@
...
+    orderAlphabetically(movies);
+
+
 // Iteration 4: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movies) {
+    if (movies.length <= 0){
+        return 0
+    } else {
+
```
tawebbcn, line 83: This should have one more level of indentation, remember to watch these for readability.

src/movies.js
```diff
@@ -1,15 +1,132 @@
...
+    ratesAverage(movies);
+    
+    // ratesAverage(movies);
+
 // Iteration 5: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(movies){
+	
+    const dramaMovies = movies.filter(function(movie){
+                 return movie.genre.includes("Drama")
```
tawebbcn, line 108: Watch indentation.

tawebbcn:
You missed one Iteration but overall really good job. Really good use of previously learned tools like .includes, .slice, .toFixed and .parseFloat!. Try to improve indentation by your own or run the auto-format extension you should have installed so you see how the code could be improved for readability.

PR #1345, opened 2020-04-02
Greatest Movies

arimagic-8bit:
Lab acabado hasta la iteración número 5. 

Saludos! :D

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -2,13 +2,110 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(movies) {
+  const newArray = [...movies];
+  newArray.sort(function(movA, movB) {
+    if (movA.year < movB.year) {
+      return -1;
+    }
+    if (movA.year > movB.year) {
+      return 1;
+    }
+    if (movA.year === movB.year) {
+      if (movA.title < movB.title) {
+        return -1;
+      }
+      if (movA.title > movB.title) {
+        return 1;
+      }
+      if (movA.title === movB.title) {
+        return 0;
+      }
+      return 0;
+    }
+  });
+  return newArray;
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+
+function howManyMovies(movies) {
+  const dramaMovies = movies.filter(function(movie) {
+    if (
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -2,14 +2,111 @@
...
+      if (movA.title < movB.title) {
+        return -1;
+      }
+      if (movA.title > movB.title) {
+        return 1;
+      }
+      if (movA.title === movB.title) {
+        return 0;
+      }
+      return 0;
```
tawebbcn, line 24: This return is redundant as you are already considering all possible cases and returning 0 when titles are the same. So you could remove it.

src/movies.js
```diff
@@ -2,14 +2,111 @@
...
+    ) {
+      return true;
+    }
+  });
+  return dramaMovies.length;
+}
+
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(movies) {
```
tawebbcn, line 46: This function runs good and it is easy to understand. However, to improve it performance wise you could do the map before the sort. That way you will be sorting only the list of titles and not all of the movie objects.

src/movies.js
```diff
@@ -2,14 +2,111 @@
...
+      return 0;
+    })
+    .map(function(movie) {
+      return movie.title;
+    });
+
+  return alfaOrder.slice(0, 20);
+}
+
+orderAlphabetically(movies);
```
tawebbcn, line 68: Remember that for Jasmine Tests you don't need to invoke these functions, Jasmine will do that for you and check on different cases :)

src/movies.js
```diff
@@ -2,14 +2,111 @@
...
+      }
+      if (movA.title > movB.title) {
+        return 1;
+      }
+      if (movA.title === movB.title) {
+        return 0;
+      }
+      return 0;
+    })
+    .map(function(movie) {
```
tawebbcn, line 61: Good job on joining methods together, however, if they have multiple lines, we would recommend to separate them for better readability.

tawebbcn:
You missed one Iteration but overall really good job. Great use of previously learned tools like .includes, .slice, .toFixed and .parseFloat!
Try to improve spacing in between variable declarations, functions and conditionals for better readability.
Remember to change the title name of your pull request to [BCN-FT-032020] + your name or names (If worked in pairs)

PR #1346, opened 2020-04-02
[BCN-FT-Alejandro-Olle]

alex-olle:
Iteration 6 tried but missing

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -2,14 +2,137 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(array) {
+  let arrYear = array.slice();
+
+  arrYear.sort(function(a, b) {
+    if (a.year < b.year) {
+      return -1;
+    }
+    if (a.year > b.year) {
+      return 1;
+    }
+    if (a.year === b.year) {
+      if (a.title < b.title) {
+        return -1;
+      }
+      if (a.title > b.title) {
+        return 1;
+      }
+    }
+  });
+
+  return arrYear;
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+
+function howManyMovies(array) {
+  let moviesNum = 0;
+
+  const newArray = array.filter(whichMovie => {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -2,14 +2,137 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(array) {
+  let arrYear = array.slice();
```
tawebbcn, line 6: Although slice works good in this case, semantically, maybe it would make more sense to use array spreading to copy the array ( let arrYear = [...array] ), because you are not trying to slice a particular part of the array.

src/movies.js
```diff
@@ -2,14 +2,137 @@
...
+  newArray.sort(function(a, b) {
+    if (a.title < b.title) {
+      return -1;
+    }
+    if (a.title > b.title) {
+      return 1;
+    }
+  });
+
+  const titleArr = newArray.map(function(movieTitle) {
```
tawebbcn, line 60: Performance wise, it would be better to do the map first to get the list of titles and then sort only the array of titles. That way the sort doesn't have to go over all of the objects inside the movies array.

src/movies.js
```diff
@@ -2,14 +2,137 @@
...
+
+  let sumRate = newArr.reduce(function(acc, currentvalue) {
+    let updatedAccumulator = acc + currentvalue;
+
+    return updatedAccumulator;
+  }, 0);
+
+  let avg = sumRate / newArr.length;
+
+  return Math.round(avg * 100) / 100;
```
tawebbcn, line 87: Perfect use of this method to get the desired number format.

tawebbcn:
Really good job on this Lab! everything was very well indented, formatted and spaced. Good use of previously learned methods like .round, .slice and .includes as well as the new methods. You could try to improve If/else conditionals and use "else if" when the cases apply (Meaning they are working on the same variable check conditions).
Overall the logic used was great and it seems you tried hard on the last iteration and it was good logic to check every character for "h" or "m" and then slice the number part of the string to be used, but the execution had some small issues in the code.
Keep up the good work!

PR #1347, opened 2020-04-02
lab-javascript-greatest-movies

RicardVillalba:
hi! I  complete the iterations 1 to 5. I tried to do everything using concepts explained in class and following lab instructions but there are a lot of things not approved by jasmine. I hope to fix it before tomorrow morning.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,14 +1,2101 @@
 /* eslint no-restricted-globals: 'off' */
+let movies = [
+  {
+    title: "The Shawshank Redemption",
+    year: 1994,
+    director: "Frank Darabont",
+    duration: "2h 22min",
+    genre: ["Crime", "Drama"],
+    rate: 9.3
+  },
+  {
+    title: "The Godfather",
+    year: 1972,
+    director: "Francis Ford Coppola",
+    duration: "2h 55min",
+    genre: ["Crime", "Drama"],
+    rate: 9.2
+  },
+  {
+    title: "The Godfather: Part II",
+    year: 1974,
+    director: "Francis Ford Coppola",
+    duration: "3h 22min",
+    genre: ["Crime", "Drama"],
+    rate: 9
+  },
+  {
+    title: "The Dark Knight",
+    year: 2008,
+    director: "Christopher Nolan",
+    duration: "2h 32min",
+    genre: ["Action", "Crime", "Drama", "Thriller"],
+    rate: 9
+  },
+  {
+    title: "12 Angry Men",
+    year: 1957,
+    director: "Sidney Lumet",
+    duration: "1h 36min",
+    genre: ["Crime", "Drama"],
+    rate: 8.9
+  },
+  {
+    title: 'Schindler"s List',
+    year: 1993,
+    director: "Steven Spielberg",
+    duration: "3h 15min",
+    genre: ["Biography", "Drama", "History"],
+    rate: 8.9
+  },
+  {
+    title: "Pulp Fiction",
+    year: 1994,
+    director: "Quentin Tarantino",
+    duration: "2h 34min",
+    genre: ["Crime", "Drama"],
+    rate: 8.9
+  },
+  {
+    title: "The Lord of the Rings: The Return of the King",
+    year: 2003,
+    director: "Peter Jackson",
+    duration: "3h 21min",
+    genre: ["Adventure", "Drama", "Fantasy"],
+    rate: 8.9
+  },
+  {
+    title: "Il buono, il brutto, il cattivo",
+    year: 1966,
+    director: "Sergio Leone",
+    duration: "3h 2min",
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,15 +1,2102 @@
...
+  },
+  {
+    title: "Dead Poets Society",
+    year: 1989,
+    director: "Peter Weir",
+    duration: "2h 8min",
+    genre: ["Comedy", "Drama"],
+    rate: 8
+  }
+];
```
tawebbcn, line 2005: There was no need to add the data to this file. Jasmine checks it by passing it through the function argument.

src/movies.js
```diff
@@ -1,15 +1,2102 @@
...
+      }
+      return 0;
+    }
+  });
+  return newArray;
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
 
+function howManyMovies(movies) {
```
tawebbcn, line 2034: Couple of issues with this function:

- The requirement was to show movies with two conditions (director and genre).
- If you check the objects, the genre property is not a string, it is another array! so to see if that array includes the string "Drama" you would have to use a method like .includes() or .indexOf() .
- Unless the requirement it specifically says to console.log the result, you should instead return it, the way you did in the function before.

src/movies.js
```diff
@@ -1,15 +1,2102 @@
...
+  const newArray = movies.filter(function(genereObj, i, movies) {
+    if (genereObj.genre == "Drama") {
+      return true;
+    }
+  });
+  console.log("newArray", newArray);
+}
+howManyMovies(movies);
```
tawebbcn, line 2042: Similar to how you did in the first iteration, the function doesn't need to be invoked. Jasmine will do this check.

src/movies.js
```diff
@@ -1,15 +1,2102 @@
...
+    }
+    if (objA.title > objB.title) {
+      return 1;
+    }
+    if (objA.title === objB.title) {
+      return 0;
+    }
+  });
+
+  newList = movies.splice(0, 19);
```
tawebbcn, line 2059: In the splice method, the second argument is the amount of elements you want returned, not the index of the last element. It works differently from slice. so in this case you should have return movies.splice(0, 20).

src/movies.js
```diff
@@ -1,15 +1,2102 @@
...
+      return true;
+    }
+  });
+  console.log("newArray", newArray);
+}
+howManyMovies(movies);
+
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(movies) {
```
tawebbcn, line 2046: Remember both sort and splice mutate the original array (even when passed as a parameter). The instructions asked you not to do that, so you would need to spread the array into a new variable. The same way you did in the first step of iteration 1.
Also, return the result instead of console.log

src/movies.js
```diff
@@ -1,15 +1,2102 @@
...
 
+function ratesAverage(movies) {
+  const eAverage = movies.reduce(function(acc, averageObj, i, movies) {
+    let updatedMovies = acc + averageObj.rate;
+
+    return updatedMovies;
+  }, 0);
+
+  var num = eAverage / movies.length;
+  var n = num.toFixed(2);
```
tawebbcn, line 2074: It is a good approach to use toFixed() but this method converts the number into a string. So in order to fix it you would have to converting it back into a number by wrapping everything into a parserFloat()

src/movies.js
```diff
@@ -1,15 +1,2102 @@
...
+  console.log(n);
+}
+ratesAverage(movies);
+
 // Iteration 5: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(movies) {
+  const newArray = movies.filter(function(genereObj, i, movies) {
+    if (genereObj.genre == "Drama") {
```
tawebbcn, line 2083: Same as Iteration 2. property genre is an array, so you will need to use .includes or .indexOf

src/movies.js
```diff
@@ -1,15 +1,2102 @@
...
+  });
+
+  const eAverage = newArray.reduce(function(acc, averageObj, i, newArray) {
+    let updatedMovies = acc + averageObj.rate;
+
+    return updatedMovies;
+  }, 0);
+
+  var num = eAverage / newArray.length;
+  var n = num.toFixed(2);
```
tawebbcn, line 2095: Same as Iteration 4. You will need to change the string into a number.

tawebbcn:
Only one Iteration done with Jasmine. The rest were almost done but there were some issues that prevented from achieve the required results like passing strings instead of numbers, mutating original arrays, not checking the objects in which you need to work on and not returning results. 
Overall your logic and application of methods is good! but you need to work on reading the instructions carefully and also finding and solving the smaller issues in your code.

PR #1349, opened 2020-04-02
done - [BCN-FT-ALEX A.]

skanndar:
Got a blocker with the  hours to mins function...

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -2,14 +2,131 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(arr) {
+  const arrCopy = [...arr];
+  arrCopy.sort(function(movie1, movie2) {
+    if (movie1.year > movie2.year) {
+      return 1;
+    } else if (movie1.year < movie2.year) {
+      return -1;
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -2,14 +2,131 @@
...
+  });
+  return arrCopy;
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
 
+function howManyMovies(arr) {
+  const arr2 = arr.filter(function(obj) {
+    if (
+      obj.genre.indexOf("Drama") !== -1 &&
```
tawebbcn, line 28: nice one! you could also use the .includes("Drama") method here

src/movies.js
```diff
@@ -2,14 +2,131 @@
...
+    arr2.map(element => {
+      if (element.rate) {
+        rating += element.rate;
+        counter++;
+        return rating;
+      }
+    });
+    let avgRate = rating / counter;
+
+    return Number(avgRate.toFixed(2));
```
tawebbcn, line 103: Here you could also shortcut it by reusing the previous function:
`return ratesAverage(arr2)`

PR #1350, opened 2020-04-02
[BCN-FT-Maxime Lagrange]

Gitrader:
Done until iteration 5.

Since I couldn't make Jasmine work I try to do it with the logic

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -2,13 +2,121 @@
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(newArr) {
+  let array2 = [...newArr];
+  let orderByYear = array2.sort(function(objA, objB) {
+    if (objA.year < objB.year) {
+      return -1;
+    }
+    if (objA.year > objB.year) {
+      return 1;
+    }
+    if (objA.year === objB.year) {
+      if (objA.title < objB.title) {
+        return -1;
+      }
+      if (objA.title > objB.title) {
+        return 1;
+      }
+      if (objA.title === objB.title) {
+        return 0;
+      }
+    }
+  });
+  return orderByYear;
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct
+
+/*Go ahead and create a howManyMovies() function that receives an array as a parameter and filter 👀 
+the array so we can have only the drama movies where Steven Spielberg is the director.
+*/
+
+function howManyMovies(array) {
+  /* if (num=0){
+
+    };
+    return num
+    */
+  if (array.length === 0) {
+    return 0;
+  }
+  const newArray = array.filter(function(obj, i, originalArr) {
+    if (obj.director === "Steven Spielberg" && obj.genre.includes("Drama")) {
+      return true;
+    }
+  });
+
+  return newArray.length;
+}
 
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(newArray) {
+  let array3 = [...newArray];
+  let order = array3
+    .sort(function(objA, objB) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -2,14 +2,122 @@
...
+function howManyMovies(array) {
+  /* if (num=0){
+
+    };
+    return num
+    */
+  if (array.length === 0) {
+    return 0;
+  }
+  const newArray = array.filter(function(obj, i, originalArr) {
```
tawebbcn, line 44: Just a little tip, try to avoid leaving unused parameters like i and originalArray.

src/movies.js
```diff
@@ -2,14 +2,122 @@
...
+    }
+  });
+
+  return newArray.length;
+}
+
 // Iteration 3: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(newArray) {
+  let array3 = [...newArray];
```
tawebbcn, line 56: Avoid using names for your variables like array3 and be more specific with naming. In this case you could use array for the function parameter and arrayCopy for the variable, for example.

src/movies.js
```diff
@@ -2,14 +2,122 @@
...
+        return -1;
+      }
+      if (objA.title > objB.title) {
+        return 1;
+      }
+      if (objA.title === objB.title) {
+        return 0;
+      }
+    })
+    .map(function(movie, index, originalArr) {
```
tawebbcn, line 69: Is good that you are linking methods but when they have several lines of code, it is better to separate them. Better for readability as the methods used might get lost in the code.

src/movies.js
```diff
@@ -2,14 +2,122 @@
...
+    });
+  if (order.length > 20) {
+    return order.slice(0, 20);
+  }
+  return order;
+}
+
 // Iteration 4: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(newArr) {
```
tawebbcn, line 80: You had good logic on this function but there were a couple issues:

- the Jasmine check for movies between 8 and 8.9 is something done automatically by jasmine, you in the function you were supposed to just get all the rates.
- Because you are applying .reduce directly to the .map you are working with the list of rates. Because of that, inside the .reduce you are not passing an object but a rate number. So instead of doing rateObj.rate you would just do rateObj. Naming of parameters probably confused you here.
- Remember .reduce applies the changes to the variable being declared ( rates ), so to get the average in the variable average you should use the variable rates and not the variable updatedAcc. The variable updatedAcc has a scope inside the .reduce method and cannot be accesed from outside.
- Finally, the .toFixed method changes the number to a string, so you will hace to wrap everything into a parseFloar( ) to change it back to a number.

src/movies.js
```diff
@@ -2,14 +2,122 @@
...
+    }, 0);
+  let average = updatedAcc / array4.length;
+  let averageRounded = average.toFixed(2);
+
+  return rates;
+}
+
 // Iteration 5: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(newArray) {
```
tawebbcn, line 99: Similar to last one, good logic but some parts were wrong along the way:

- in the .filter, you are already working on each element of the array, so for the conditional, you don't need to do: "newArray ===". Only leave " obj.genre.includes("Drama") " for the validation.
- in the .reduce method, there is a typo "moviej". Also, since you mapped only the rates, you don't need to do movie.rate (The parameter movie here is only the rate number)
- See the same two last feedback points from the previous function.
- You are returning the reduced number, not the averageRounded variable.

tawebbcn:
Really good job on the first 3 iterations but you had a lot of issues on the 4th and 5th. You have a very good logic but try to separate the methods and apply them to different variables so you get a better understanding on how they are applying the changes to each variable and what parameters are you actually working on. That will help you apply the logic better.

PR #1351, opened 2020-04-02
[BCN-FT-032020] Daniel Hamekasi

wangagat:
didnt have time to finish the 6th and bonus :(

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,14 +1,90 @@
 /* eslint no-restricted-globals: 'off' */
 
 // Iteration 1: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies) {
+  let tempArray = [...movies];
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,14 +1,90 @@
...
+    }
+  });
+
+  if (dramaMovies.length === 0) {
+    return 0;
+  } else {
+    let averageDramaRating =
+      dramaMovies.reduce((accumulator, movie) => {
+        return accumulator + movie.rate;
+      }, 0) / movies.length;
```
tawebbcn, line 83: The last test is failing because you're dividing the sum of drama movies rates by the length of the original movies array (instead of dividing by the **dramaMovies** array length).
A shorter way would be to reuse the function you created just before: `return ratesAverage(dramaMovies)`

PR #1354, opened 2020-04-03
Arslane Gharout FT March BCN

iamarslane:


movies.js
```diff
@@ -0,0 +1,69 @@
...
+}
+
+// ugh i don't know how to resolve this one (iteration 4)
+
+// Iteration 5: Drama movies - Get the average of Drama Movies
+
+function dramaMoviesRate(array){
+    let genreFilter = array.filter(movie => {return movie.genre.includes("Drama");})
+    return ratesAverage(genreFilter);
+}
```
tawebbcn, line 56: excellent, very straight-forward 👍

movies.js
```diff
@@ -0,0 +1,69 @@
...
+return title.slice(0,20);
+}
+
+// Iteration 4: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(array){
+const rating = array.reduce(accumulator, currentValue){
+    return currentValue.rate / array.length 
+}
+}
```
tawebbcn, line 47: Nice start, it's a good idea to use the reduce() method here.
Watch out the syntax though, (accumulator, currentValue) are the 2 arguments of the callback function inside reduce(). You could do something like this:

```
  if (array.length === 0) return 0;
  const rating = array.reduce((accumulator, currentValue) => {
    if(currentValue.rate) return accumulator + currentValue.rate;
    else return accumulator;
    }, 0) / array.length;
  return parseFloat(rating.toFixed(2));
}
```

PR #1413, opened 2020-06-05
[BER-WDFT-JUN20] Marco Muser

marcomuser:


src/movies.js
```diff
@@ -1,14 +1,56 @@
...
+        return movie.director
+    })
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(movies) {
+    return movies.filter(function(movie) {
+        return movie.director === "Steven Spielberg";
```
CarlosZiegler, line 13: Hello, you can use  "&&" AND operator, then you no need the next filter. Like this
return movie.director === "Steven Spielberg" && movie.genre.includes("Drama").

marcomuser: Thanks for the hint! I just implemented and added it to this pull request.

PR #1415, opened 2020-06-05
WDFT-Berlin-June-2020-Luis-Schmitz

luis3110:
First three iterations, when I had questions Majd helped me out. Great guy!

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -41,9 +41,53 @@ function ratesAverage(movies) {
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(movies){
+    let dramaMovies = movies.filter(function (movie) {
+        if (movie.genre.includes("Drama")) {
+          return movie;
+        }
+      });
+      if (dramaMovies.length === 0) {
+          return 0;
+      };
+    let sumRates = dramaMovies.reduce(function(acc, val){
+        return acc + (val.rate || 0)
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,94 @@
...
+          return 0; 
+      }
+    });
+    let spielbergDrama = dramaMovies.filter(function (movie) {
+      if (movie.director.includes("Spielberg")) {
+        count += 1;
+      }
+    });
+    return count;
+  }
```
Jeff-Moraes, line 28: following the logic that you created for this interaction, this is one of the possible solution
```function howManyMovies(array) {
  let count = 0;
  let dramaMovies = array.filter(function (movie) {
      return movie;
    }
  });
  let spielbergDrama = dramaMovies.filter(function (movie) {
    if (movie.director === "Steven Spielberg") {
      count += 1;
    }
  });
  return count;
}
```
instead of using the method includes() to check the string "Steven" you should use the boolean operators === to check if the value of the propriety director is the same as the string "Steven Spielberg".

luis3110: Okay thanks Jeff. I thought with includes() I would make sure that it also gets it if there are more directors. But if this is also possible with === I think it is more reliable :)

luis3110:
Thanks for the review. It was fun to work on the Lab, and I was very thankful for Majd and his help :)!

PR #1417, opened 2020-06-05
[BER WDFT JUN20] Pauline Didier

Tyouxik:
Mandatory iterations done
Attempt to do Iteration 7

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,53 @@
+
+
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(movies) {
+    return movies.map(function(movie){
+        return movie.director
+      })
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(movies) {
+    const spielbergDrama = movies.filter(movie => movie.director === "Steven Spielberg" && movie.genre.includes("Drama"))
+    return spielbergDrama.length
+  }
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
-
+function ratesAverage(movies) {
+    if (movies.length === 0) {return 0};
+    let totalRates = movies.reduce(function(sum, movie) {
+      return sum + movie.rate;
+    }, 0)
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,15 +1,53 @@
...
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(movies) {
+    const spielbergDrama = movies.filter(movie => movie.director === "Steven Spielberg" && movie.genre.includes("Drama"))
+    return spielbergDrama.length
+  }
```
Jeff-Moraes, line 17: super nice solution!

src/movies.js
```diff
@@ -1,15 +1,53 @@
...
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
-
+function ratesAverage(movies) {
+    if (movies.length === 0) {return 0};
+    let totalRates = movies.reduce(function(sum, movie) {
```
Jeff-Moraes, line 21: to solve the error
`Return average even if one of the movies does not have rate!`
you can just add an if statement for the movies that hasn't a rate
```if (!movie.rate) return sum;```
so for the ones that hasn't a rate value you can just return the sum.

File: src/movies.js [hunk 3]
```diff
@@ -1,15 +1,53 @@
...
+    let averageRate = totalRates / movies.length
+    return Number(averageRate.toFixed(2))
+}
 // Iteration 4: Drama movies - Get the average of Drama Movies
-
+function dramaMoviesRate(movies) {
+    let dramaMovies = movies.filter(function (movie) {
+        return movie.genre.includes("Drama");})
+    return ratesAverage(dramaMovies);
+}
```
Jeff-Moraes, line 32: Super nice that you reused the ratesAverage function.

src/movies.js
```diff
@@ -1,15 +1,53 @@
...
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-
+function orderByYear(movies) {
+    return movies.slice().sort(function(a, b) {
+        if (a === b){
```
Jeff-Moraes, line 36: here you should compare the propriety year of the movies as
```if (a.year === b.year) {```
this will solve the error
`If two movies have the same year, order them alphabetically by their title`

PR #1418, opened 2020-06-06
Berlin-WDFT-<Majd Sabik>

majdsabik:


src/movies.js
```diff
@@ -1,16 +1,107 @@
...
+  movies.reduce(function (acc, val) {
+    if (val.genre.includes("Drama")) {
+      rateSum += val.rate;
+      count++;
+      noDrama = false;
+    }
+  }, 0);
+  if (noDrama) return 0;
+  return parseFloat((rateSum / count).toFixed(2));
+}
```
Jeff-Moraes, line 44: This is nice solution!
For a shorter solution, you can filter the movies that include the Drama genre and then past this new array as an argument to the ratesAverage(filteredMoviesArr) function. ;)

src/movies.js
```diff
@@ -1,16 +1,107 @@
...
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-
+function orderByYear(movies) {
+  let clone = movies.slice();
+  clone.sort((a, b) => (a.year > b.year ? 1 : -1));
+  return clone;
+}
```
Jeff-Moraes, line 50: Uow.. super nice and short solution!!

src/movies.js
```diff
@@ -1,16 +1,107 @@
...
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-
+function orderAlphabetically(movies) {
+  let clone = movies.slice();
+  let sorted = clone.sort((a, b) => (a.title > b.title ? 1 : -1)).slice(0, 20);
```
Jeff-Moraes, line 54: I liked that you used the ternary statement here

src/movies.js
majdsabik: That's an elegant solution.
Thanks for the review and the helpful comments :D

PR #1420, opened 2020-06-06
[BER WDFT] Eduarda Liu

eduardaliu:


src/movies.js
```diff
@@ -1,16 +1,84 @@
...
+const orderAlphabetically = array => {
+    const orderedTitles = array.map (movie => movie.title).sort(); // ternary + cleaning up syntax when you only have a return value
+    if (orderedTitles.length > 20) return orderedTitles.slice (0,20);
+    else return orderedTitles
+}
+
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 
+const turnHoursToMinutes = array => {
+    const updatedMovies = [...array];
```
fonsogms, line 67: for some reason the testing doesnt pass the original array test.
So you could do updatedMovies=array.splice().
and it will work.

PR #1423, opened 2020-06-07
Berlin WDFT June 2020 Verena Mayrock

verenamay:


src/movies.js
```diff
@@ -1,16 +1,84 @@
...
 
+// create new Array
+const dramaMovies = movies.filter(function(drama) {
+    return drama.genre.includes('Drama') && drama.genre.length==1}); 
+
+// console.log(dramaMovies);
+
+function dramaMoviesRate(dramaMovies) {
+const averageDramaWithAllDecimals = dramaMovies.reduce(function(sum, val) {
+return sum+val.rate/dramaMovies.length;
```
fonsogms, line 46: here you could have filtered the movies by gender with the include method.
and then call your own previous fuction for avergae
example:
```
 const dramaMovies = moviesArr.filter(function (movie) {
          return movie.genre.includes("Drama");
      });
```

src/movies.js
```diff
@@ -1,16 +1,84 @@
...
+  
+  howManyMovies(movies);
+
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movies) {
+    const averageWithAllDecimals = movies.reduce(function(sum, val) {
+      return sum+val.rate/movies.length;
+    }, 0)
```
fonsogms, line 29: here you needed to check in some of the elements if they had any rate at all because some of the movies didnt have any rate so you could have created a small if statement in the reduce fuction to check that part:
```
if (typeof val.rate==="number") return sum+val.rate;
        else return sum;
```

src/movies.js
```diff
@@ -1,16 +1,84 @@
...
+};
+
+dramaMoviesRate(dramaMovies); 
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(movies) {
+    const newSortedArray=movies.slice();
+    const years= newSortedArray.sort(function(a,b) {
+      return a.year-b.year;
```
fonsogms, line 59: here you needed to make also an if statement to check if the the year is the same to order by title.
And to order by title you ca use localeCompare()
example:
```
 if (older.year==newer.year) {
    return a.title.localeCompare(b.title)
  } else return older.year-newer.year;
  })
```

PR #1424, opened 2020-06-07
WDFT Berlin Nina 

nina-sm:


src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+
+    return array.map((movie) => movie["director"])
+
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies() {
```
fonsogms, line 13: you forgot to add here array as a an argument in the howManyMovies function:
example:
```
 howManyMovies(array) 
```

PR #1425, opened 2020-06-07
Yuri [Berlin WDFT June 2020] 

yuri-la-torre-ortiz:
Done.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,79 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors (arrayOfMovies) {
+    let mappedArray = arrayOfMovies.map(function(movie) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,79 @@
...
+    let title2 = b.toLowerCase();
+    if (title1 < title2) return -1;
+    if (title1 > title2) return 1;
+    if (title1 === title2) return 0;
+  })
+  if (titleList.length < 20){
+    return titleList.slice(0,titleList.length)
+  }
+  return titleList.slice(0,20)
+  };
```
timbotimber, line 76: Definitely works! You could refactor this with a localecompare as well. I'm also not sure you need to use the if statement as well as the slice. But if it works, it works!

PR #1427, opened 2020-06-08
[WDV-06.2020-Berlin]-Romulo

Romulo91:
"done"- without Bonus


src/movies.js
```diff
@@ -1,16 +1,91 @@
...
 
+
+function howManyMovies(a){ 
+    const newArr = a.filter(i => {
+      return i.director === `Steven Spielberg` && i.genre.indexOf(`Drama`)!=-1
+    },)
+    return newArr.length;
+  }
+
+
```
timbotimber, line 21: Nice work. Could also use includes as an alternative to indexOf -1

PR #1428, opened 2020-06-08
Eliott - Fournet / Web-Dev-Berlin-FT-June2020

Eliott-Fournet:
Just miss the 7th iteration, coming ASAP.  ( ͡° ͜ʖ ͡°)


src/movies.js
```diff
@@ -1,16 +1,111 @@
...
+    return 1;
+  }}
+    } else {
+
+    return a.year - b.year;
+    }});
+
+    return yearSorted;
+}
+
```
timbotimber, line 86: Looks good. An alternative you could use localecompare as opposed to the .toUpperCase loop

PR #1430, opened 2020-06-08
Rachel de la Torre didn't finish but is learning more every day!

Rachel-de-la-torre:
Just keep swimming!

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,12 +1,65 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+const getAllDirectors = movies.map(function(movie){
+    return movie.director
+})
+
+console.log(getAllDirectors)
+
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+
+// const getAllDirectors = movies.map(function(movie){
+//     return !movies.includes(movie);
+// })
+
+// console.log(getAllDirectors)
+
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+
+
+// function howManyMovies(arr)
+
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,12 +1,65 @@
...
+
+
+    // const placesWithPool = places.filter(function (place) {
+    //     return place.pool === 'yes';
+    // }).map(function (place) {
+    //     return {
+    //         title: place.title
+    //     }
+    // });
+
```
timbotimber, line 43: the start of this is pretty accurate. You don't need the second if though and it would have worked: 

```
function howManyMovies(arr) {
  let spielbergMovies = arr.filter(function(movie) {
    if (
      movie.director === "Steven Spielberg" &&
      movie.genre.includes("Drama")
    ) {
      return true;
    }
  });
  return spielbergMovies.length;
}
```

src/movies.js
```diff
@@ -1,12 +1,65 @@
...
+function ratesAverage(arr) {
+    const average = movies.reduce(function(sum,val){
+     return sum + val.rate || 0;
+     },0);{
+return Number((average / arr.length).toFixed(2));
+}
+
+// .filter 
+
```
timbotimber, line 55: Nice, this looks good.

src/movies.js
```diff
@@ -1,12 +1,65 @@
...
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(movies){
+    const dramaMovie = movies.filter (function(movie)}
+    return movie.genre.indexOf("Drama")
+})
+}
```
timbotimber, line 62: the ```indexof``` method doesn't quite work here. A different approach might be to filter the array (like you already have) then us an if statement to find the genre with an ```.includes``` method: 

```
function dramaMoviesRate(arr) {
  let dramaMoviesRated = arr.filter(function(movie) {
      return true;
    }
  });
  return ratesAverage(dramaMoviesRated);
}
```

PR #1456, opened 2020-06-08
[WEB MAD] Míriam Sánchez

msanchezalcon:
This lab has been one of the hardest so far. The more I try to revise simpler ways of building functions, the easier it is for me to actually learn them. There are some red crosses I couldn't figure out in Jasmine, but I will try to get back to it a later point and will also try to complete the bonus.

My main issues come with deeper structures and several conditions within the same function. I still need A LOT of practice and to learn by example. Once I see examples I have that ahá moment, but I feel like I can't come up with many ideas myself.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,49 @@
 // Iteration 1: All directors? - Get the array of all directors.
+const getAllDirectors = (movies) => movies.map((elm) => elm.director);
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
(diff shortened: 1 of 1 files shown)


ta-web-mad:
Hi , I know it can be overwhelming but try to take slow, you did a great job, there were small things that caused the test to fail, for example you were missing a return in this filter:
```
function howManyMovies(movies) {
  let spielbergDrama = movies.filter((elm) => {
    **return** elm.genre.includes("Drama") && elm.director === "Steven Spielberg";
  });
  return spielbergDrama.length;
}
```
As you can see these are small details, so don't worry and  keep up the great work.

Dayan R.

PR #1468, opened 2020-06-14
[REMOTE-WDPT-0602] - Matthew Howard

mattalanhoward:


valmgisbert:
wonderful use of `new Set()` and other methods! try to stick to one way (either ES5 or ES6) for better practice; you can comment the alternate way to keep it for reference but to be consistent keep it either one or the other. Great on commenting your steps! Also kudos on practicing different ways to do the same thing (like sorting alphabetically). Remember to clear your `console.logs`. Great work!

PR #1476, opened 2020-06-15
solve all iterations, except for parts of the bonus

willemprins93:


valmgisbert:
Such clean code! I love it! try to stick to either ES5 or ES6 on your file; as long as you don't change the name of your functions, it should work just fine. Loved every step commented, keep up that good practice. Excellent work!

PR #1478, opened 2020-06-15
PTMIA062020 AndrewChen

achen002:


src/movies.js
```diff
@@ -1,16 +1,249 @@
...
+
+    if (array1.length === 0) {
+        return 0;
+      }
 
+    const spielbergDrama = array1.filter(function(movie) {
+      //checks to see if directore is Steven
+      if(movie.director ==="Steven Spielberg") {
+          //iterates over the genre key to see if drama
+          for(let i = 0; i < movie.genre.length; i ++) {
```
sandrabosk, line 22: A cleaner and shorter way to achieve the same would be using the `.includes()` method:
```js
function howManyMovies(someMovies) {
  return someMovies.filter( mov => mov.director === 'Steven Spielberg' && mov.genre.includes('Drama')).length;
}
```

src/movies.js
```diff
@@ -1,16 +1,249 @@
...
+ 
+  
+  },0);
+  let avgRating = movieRating / array1.length;
+  
+  return Number(avgRating.toFixed(2));
+  }
+  
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(array) {
```
sandrabosk, line 52: Since you have already created the `ratesAverage ()` function, you could just reuse it here:
```js
function dramaMoviesRate(someMovies) {
  let drMovies = someMovies.filter(eachMovie => eachMovie.genre.includes('Drama'));
  return ratesAverage(drMovies);
}
```

src/movies.js
```diff
@@ -1,16 +1,249 @@
...
+  }
+  
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(array) {
+    if (array.length === 0) {
+        return 0;
+      }
+    
+      const sortedByYear = array.sort(function (movie1, movie2) {
```
sandrabosk, line 86: `.sort()` method will mutate your original array so, to avoid that, we should first make a copy of the original array:

```js
function orderByYear(array) {
  const newArr = [...array];
  newArr.sort(...)
}
```

src/movies.js
```diff
@@ -1,16 +1,249 @@
...
+      });
+    
+      return sortedByYear;
+    }
+    
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(array) {
+    let newArray = [];
+    array.forEach(function(elem) {
```
sandrabosk, line 103: t seems you are just creating a copy of the array here. You can use `.slice()` or `[...arr]` approach to achieve the same result.

src/movies.js
```diff
@@ -1,16 +1,249 @@
...
+      const sortedByAlpha = newArray.sort(function (movie1,movie2) {
+        if(movie1.title < movie2.title) return -1;
+        if(movie1.title > movie2.title) return 1;
+    
+      });
+      const onlyTitles = sortedByAlpha.map(function(movie) {
+        return movie.title;
+      });
+    
+      if(sortedByAlpha.length > 20) {
```
sandrabosk, line 117: Although this approach works, you could have use `.slice(0, 20)` to get the first 20 titles.

src/movies.js
```diff
@@ -1,16 +1,249 @@
...
+  for(let i = 0; i < finalArray.length; i++) {
+    finalArray[i].duration = firstDigitToNumber[i];
+  
+  }
+  
+   return finalArray;
+  }
 // BONUS - Iteration 8: Best yearly rate average - Best yearly rate average
+
+function bestYearAvg(array) {
```
sandrabosk, line 178: Find a bit shorter solution below:

```js
function bestYearAvg(lotsOfMovies) {
  if (!lotsOfMovies.length) return null;

  let masterObject = {};

  lotsOfMovies.forEach(eachMovie => {
    if (!masterObject[eachMovie.year]) {
      masterObject[eachMovie.year] = [eachMovie];
    } else {
      masterObject[eachMovie.year].push(eachMovie);
    }
  });

  let highest = 0;
  let theActualYear;
  for (let theYear in masterObject) {
    if (ratesAverage(masterObject[theYear]) > highest) {
      highest = ratesAverage(masterObject[theYear]);
      theActualYear = theYear;
    }
  }
  return `The best year was ${theActualYear} with an average rate of ${highest}`;
}```

sandrabosk:
Hi  Andrew, besides small improvements which are already pointed out, you provided good solutions overall! Keep it up!

PR #1480, opened 2020-06-15
[MIA-PT-062020] Lee Fung-A-Fat  

leefaf:
Here is my best attempt to the hw. Included notes with collaboration and help

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,113 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(arr)
+{ const allDir = arr.map(movie => {
+  return movie.director;
+  }); return allDir;
+  }
+  getAllDirectors(movies);
+
+  // A little help with writing the Syntax with Mikayla and Nick
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+// Having trouble to get it  will still keep working on it. 
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(arr){
+    const spielbergMovies = arr.filter(drama => {
+      if(drama.director === 'Steven Spielberg' && drama.genre.includes('Drama'))
+        return drama;
+    });
+       return spielbergMovies.length;
+    }
+    howManyMovies(movies);
+
+    // colab with Mikayla but logic makes sense. Collab with Mikayla to practise the Syntax
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(arr){
+    const sumOfRates = arr.reduce((accum, currentRate) => {
+      if(currentRate.rate){
+        accum+=currentRate.rate;
+      }return accum;
+      }, 0);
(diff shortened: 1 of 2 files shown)


src/movies.js
```diff
@@ -1,16 +1,113 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(arr)
```
sandrabosk, line 3: Although this solution is correct, I need to point out that it doesn't follow any good practices with the syntax and code organization. We put certain things on a new line for a reason to make it visually better represented. This code could be turned into something like this:
```js
   const directors = movies.map(oneMovie => oneMovie.director);
   return directors;
}
```

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+// Having trouble to get it  will still keep working on it. 
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(arr){
+    const spielbergMovies = arr.filter(drama => {
+      if(drama.director === 'Steven Spielberg' && drama.genre.includes('Drama'))
```
sandrabosk, line 20: Great approach!

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+function ratesAverage(arr){
+    const sumOfRates = arr.reduce((accum, currentRate) => {
+      if(currentRate.rate){
+        accum+=currentRate.rate;
+      }return accum;
+      }, 0);
+    const avgOfRates = sumOfRates/arr.length;
+   if(arr.length === 0){
+     return 0;
+   } return parseFloat(avgOfRates.toFixed(2));
```
sandrabosk, line 40: Great. Can you think of some other ways to turn string to number?

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+  // A help with writing the Syntax with Mikayla and Nick 
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(arr){
+    if(arr.length === 0) return 0;
+    var dramaType = movies.filter(function(drama){
+        return drama.genre.includes("Drama");
+      });
+      return ratesAverage(dramaType);
```
sandrabosk, line 53: Good! You reused previously defined function!

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+      return 0
+    }
+      const sortedArrAlpha = arr.sort(function(a, b){
+          if(a.title > b.title) return 1;
+          if(a.title < b.title) return -1;
+          return 0;
+          })
+          return sortedArrAlpha;
+          if (sortedArrAlpha.length > 20){
+     constNewList = sortedArrAlpha.length.slice(0,19);
```
sandrabosk, line 98: Besides returning just first 20 of movies, we are asked to return only the titles, not the whole objects. So your code needs additional `.map()` method:
```js
sortedArrAlpha.map(movie => movie.title).slice(0, 20);
return sortedArrAlpha;
```

sandrabosk:
Good job Lee! Great approach to solving problems. Keep it up!

PR #1482, opened 2020-06-15
[PTMIA062020] Alexey Ugodnikov

alex-ugodnikov:
working on iteration 7 was very hard and I couldn't complete a few tests. Needed some guidance on using Regular Expressions

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,81 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors (arrayOfMovies) {
+    let directors = arrayOfMovies.map(dir => dir.director);
+   return directors;
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies (arrayOfMovies) {
+    let filteredArray = arrayOfMovies.filter(movie => 
+     (movie.director === "Steven Spielberg")&&(movie.genre.includes("Drama")))
+    return filteredArray.length; 
+  }
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(arrayOfMovies) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,89 @@
...
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(arrayOfMovies) {
+    if (!arrayOfMovies.length) return 0;
+    let sumOfRates = arrayOfMovies.reduce((accumValue, currValue) => {
+      if (currValue.rate > 0) 
+      return accumValue + currValue.rate; 
+      return accumValue;
+    }, 0)
+    return parseFloat((sumOfRates / arrayOfMovies.length).toFixed(2));
```
sandrabosk, line 27: Great approach. Can you think of some other ways to turn string to numbers?

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+  }
+
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate (arrayOfMovies) {
+    let dramaMovies = arrayOfMovies.filter(movie => (movie.genre.includes("Drama")));
+    if (!dramaMovies.length) return 0;
+    return ratesAverage(dramaMovies);
```
sandrabosk, line 36: Amazing approach to reusability. Way to go!

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically (arrayOfMovies) {
+    let sortedArr = arrayOfMovies.slice().sort((a,b) => {
+      if (a.title > b.title) return 1;
+      if (a.title < b.title) return -1;
+    }); 
+    let limitArr = [];
+    for (let i=0; (i<sortedArr.length)&&i<20; i++) {
```
sandrabosk, line 59: Using loop could be one way to go about this. However a cheaper way would be using just simple `.slice(0, 20)` method.  Your solution could be somewhat like this:
```js
sortedArr.map(eachMovie => eachMovie.title).slice(0, 20);
return sortedArr;
```

sandrabosk:
Overall, very good approach to solving problems. Keep it up!

PR #1484, opened 2020-06-15
[Remote-PT-062020] Nuno Pacheco

nuno-pacheco:


valmgisbert:
Remember that you don't need to `console.log` an invoked function. Nice consistency! try to practice arrow functions to practice the transition to ES6. Great work!

PR #1486, opened 2020-06-15
[REMOTE-WDPT-0602] - Pedro Lamy

PedroMMLamy:


valmgisbert:
Keep the consistency between ES5 vs ES6; as long as you don't change the function name it will work. Also, don't copy vars from data file in your `movies.js` because it will mess with Jasmine and render your code useless. Check your syntax errors and where you close your functions, try to read carefully the instructions to see what the exercise and Jasmine are asking of you, and work on spacing for a cleaner code. If you can, keep working on this. Still, very nice work!

PR #1488, opened 2020-06-16
[REMOTE-WDPT-0602] Cristina Matamoro

cris-developer:
src/movies.js


valmgisbert:
If you go with, say ES5, comment the other so it doesn't confuse Jasmine (or yourself!); also as long as you don't change the function name, you can change in to ES6 format with no problem. Careful with spaces between method invoking, and around your code structure. Very good work!

PR #1489, opened 2020-06-16
REMOTE-WDPT-0602] - Giovanna Izzo

Hachiko9:
Had a lot of problems with this lab


valmgisbert:
Wonderful code! I am curious as to why you changed the `data` let to a const. WOO SOME ONE-LINERS! Amazing job!

PR #1490, opened 2020-06-16
[REMOTE-WDPT-0602] - Alejandro Gonzalez - LAB - Gratest movies.

alejanglez:
LAB-javascript-greatest-movies

Iterations 1-6 complete without bonus, iteration 5, doesnt work for Jasmine.


valmgisbert:
Try to remain consistent between ES5 vs ES6; if you don't change the function name then you're okay. Some methods clash against what Jasmine wants, you should check your code for them; also, be careful with using extra or unidentified functions for Jasmine since it can break her. Check your spacing along your code to make it easier to read. Nice job!

PR #1491, opened 2020-06-16
[REMOTE PTWD JUNE 2020] - Lynn W-B

lynn-w-b:
All iterations completed but not yet passing Jasmine tests...to be continued.


valmgisbert:
The main struggle you have here with Jasmine is that you're not declaring your functions, so it doesn't recognise them. Check for typos and where you're closing your blocks of code. Remember to stay ES5 vs ES6 consistent; it's okay if you stick to ES5 and then refactor or practice later, but go with the most clear one for you first. Same thing with one-liners; try to work on more thorough and step-by-step code and then refactor to one-liners to avoid errors and losing them. Whenever you can, please come back and work on this. Still, good work!

PR #1492, opened 2020-06-16
[REMOTE-WDPT-0602] - Juan Jose Carreño

JJexcelsior:


valmgisbert:
This is an incredibly clean code! Nice use of ternaries. As long as you understand it when you come back to it, great format; you can comment some steps otherwise if it helps you. Remember to stay consistent between ES5 vs ES6; you'll be fine as long as you don't change the name. Excellent work!

PR #1494, opened 2020-06-16
[REMOTE-WDPT-0602] - André Bacalhau

AndrewMastercode:


valmgisbert:
Remember to stick to either ES5 vs ES6; it'll be fine as long as you don't change the name. In your err in `ratesAverage`, take into account that some movies don't have rating, and that might affect your result. Very nice and structured code! Good work!

PR #1496, opened 2020-06-17
[MIA-WDPT-062020] Ana Ortez

anaortez23:
Iterations 1-6 need help last part of 7

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,99 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+const getAllDirectors = moviesArr => {
+    console.log('here')
+    let getDirectors = moviesArr.map(el => el.director);
+    return getDirectors;
+   
+};
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+
+let directors = getAllDirectors()
+
+const filteredDirectors = directors.filter((director, index) => directors.indexOf(director) === index);
+
+console.log(filteredDirectors)
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+const howManyMovies = moviesArr => {
+    let filteredDrama = moviesArr.filter(function (movie) { return movie.genre.includes('Drama') && movie.director === 'Steven Spielberg' })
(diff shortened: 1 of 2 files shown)


src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+    let theAvg = theSum / movies.length;
+    
+    return theAvg.toFixed(2)
+  
+  };
+  ratesAverage()
+
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
+const avgDramaRate = moviesArr => {
```
sandrabosk, line 42: You need to use the same function names as they are described in tests. So here the correct name would be `dramaMoviesRate()`. Also you have passed the parameter `moviesArr` which you don't use in the function. Instead, you are using `movies`. These are small issues but it is important to keep an eye on everything.

src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+orderByYear()
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+orderAlphabetically = arr => {
+    let orderedArr = movies.slice().sort((a,b) => {
+    if (a.title < b.title) return -1;
+    if (a.title > b.title) return 1;
+    return 0;
+    });
+    let shortArray = orderedArr.slice(0,20);
```
sandrabosk, line 76: They want us to return only titles so we should also map through the existing array and take only the titles. Something along these lines:
```
let shortArray = orderedArr.map(eachMovie => eachMovie.title).slice(0, 20);
return shortArray;
```

src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+  
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+const ratesAverage = arr => {
+    let theSum = movies.reduce((accum, currValue) => accum + currValue.rate, 0);
+  
+    let theAvg = theSum / movies.length;
+    
+    return theAvg.toFixed(2)
```
sandrabosk, line 35: I am pretty sure you will get the string here so turning it to number would be required. You can use either `parseInt()`, `Number()` or simply `+` sign. Your solution would look something like this:
```
return Number(theAvg.toFixed(2));
```

sandrabosk:
Great solution, congrats. Keep it up. Just make sure you pay attention to tests and follow the given naming conventions and test cases.

PR #1515, opened 2020-07-09
[WDFT-REM-2007] ALVARO SANCHEZ LAMADRID

AlvaroSLam:
Iterations 1, 2, 3 and 4 completed

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,9 +1,44 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(movieArray) {
+    let directors = movieArray.map(function(item){
+        return item.director;
+    })
+
+    return directors
+}
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+    
+function howManyMovies(movieArray) {
+    let filterMovies = movieArray.filter(function(item){
+        return item.director === "Steven Spielberg" && movieArray.includes("Drama")
+        }         
+    })
+    return filteredMovies.length;
+}
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(movieArray) {
+    if (movieArray.length == 0) {return 0;}
+
+    let rate = movieArray.reduce(function (acc, elem) {
+        let i = 0;
+        if (!movieArray.rate){
+            i++return acc +0} else { 
+                return acc+movieArray.rate}}, 0)
+
+        let scoreavg = (rate /(movieArray.length-i))
+        return Math.round(scoreavg * 100) / 100;
+    }
+    
+
+
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,51 +1,103 @@
 // Iteration 1: All directors? - Get the array of all directors.
 
+
 function getAllDirectors(movieArray) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,103 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+
+function getAllDirectors(movieArray) {
+    let directors = movieArray.map(function(item){
+        return item.director;
+    })
+}
```
IH-WebDev-TA-Remote, line 8: For the function to do something, you will need to return the directors array you created.

src/movies.js
```diff
@@ -1,16 +1,103 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+
+function getAllDirectors(movieArray) {
+    let directors = movieArray.map(function(item){
```
IH-WebDev-TA-Remote, line 5: instead of naming the parameter "item", you could name it "movie" or "eachMovie". It would be more understandable since every item is a movie object.

src/movies.js
```diff
@@ -1,16 +1,103 @@
...
+}
+    
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movieArray){
+    let filteredList = movieArray.filter(function(item){
+        return item.["director"] === "Steven Spielberg" && item.["genre"].includes("Drama")
```
IH-WebDev-TA-Remote, line 17: Good use of .includes!

src/movies.js
```diff
@@ -1,16 +1,103 @@
...
+    })
+    return filteredList.length;
+}
+
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movieArray) {
+    if (movies.length == 0){
```
IH-WebDev-TA-Remote, line 26: remember to use triple equal instead of double equal when comparing. Best practices are important!

src/movies.js
```diff
@@ -1,16 +1,103 @@
...
+        return 0;
+    }
+
+
+    let ratesAverage = movieArray.reduce(function(acc, item){
+        return acc + item.["rate"] / movieArray.length
+    }, 0)
+    return ratesAverage.toFixed(2)
```
IH-WebDev-TA-Remote, line 34: Good use of toFixed, however this method converts the number into a string and you want to return a number. So in order to fix this you will need to do one of the many ways to convert it back to a number (Number. , parseFloat or the + symbol)

src/movies.js
```diff
@@ -1,16 +1,103 @@
...
+}
+
+
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(movieArray) {
+    let dramaMovies = movieArray.filter(function(item){
+        return item.["genre"].["Drama"]
```
IH-WebDev-TA-Remote, line 43: Drama is not a key, it is a value. So to check only movies with the drama genre you need to do similar to how you did in a previous iteration ( item["genre"].includes("Drama") )

src/movies.js
```diff
@@ -1,16 +1,103 @@
...
+    })
+    return ratesAverage(dramaMovies)
```
IH-WebDev-TA-Remote, line 45: Awesome re-use of previous function!

IH-WebDev-TA-Remote:
Very good job! I would say other than those small issues in the code, you manage to do and understand the mandatory part. Remember to check your code for syntax errors (those little red curly lines) or by check the bottom left of your VS code.
Keep it up! and don't be afraid to try those bonus iteration, even if you don't manage to finish them you will learn something out of it. And check the solution to understand how to do them.

PR #1516, opened 2020-07-09
[WDFT-REM-2007] Elise Dujardin

elisedjn:
All done! Thank you Jorge for your help on the iteration 7! 
There are some stuff I've done which are working but I don't always know why 😅

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,187 @@
 // Iteration 1: All directors? - Get the array of all directors.
+const getAllDirectors = (moviesArr) => {
+    let directorsArr = moviesArr.map(function(movie){
+        return movie["director"];
+    });
+    return directorsArr;
+}
+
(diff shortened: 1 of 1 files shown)


IH-WebDev-TA-Remote:
Amazing work! All tests passing and the code looks very clean and easy to understand. Also the comments are always appreciated! :)
You are doing sooo good that my only feedback would be to start getting comfortable using arrow functions, reduced syntax and chaining methods. This is definitely a more advanced way of coding but if you start now it would really help you later.

for example, in iteration 6:
Instead of:

 alphaOrderMoviesArr.sort(function (a, b){
      if (a > b){
          return 1;
      } else if (a < b) {
          return -1;
      } else {
          return 0;
      }
  })

You can do: 

alphaOrderMoviesArr.sort((a, b) => a - b)

Keep up the great work! you are doing amazing :)

PR #1517, opened 2020-07-09
wd-remote 2007- katie stamey

kmstamey:


src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(movies){
+
+    let occurance = movies.filter(function(movie) {
```
IH-WebDev-TA-Remote, line 18: Try to come up with more descriptive variables of what would be inside it like filteredArray or filteredMovies

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+
+        if (movie.director != "Steven Spielberg"){
+            return false;
+
```
IH-WebDev-TA-Remote, line 22: Watch spacing, you can add them between conditionals or functions but try to avoid single spaces inside statements.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+    return occurance.length;
+     
+}
+
+
+howManyMovies(movies);
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(movies){
+    if (movies.length == 0) {
```
IH-WebDev-TA-Remote, line 40: For best practices, remember to use triple equal ( === ) instead of double equal ( == )

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+
+    let avg = total/movies.length;
+    return avg;
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(movies){
+    let occurances = movies.filter(function(movie) {
+
+        if (movie.genre.indexOf("Drama")== -1 ) {
```
IH-WebDev-TA-Remote, line 61: Using indexOf is perfectly fine, but there is another method called .includes that returns a boolean if an instance of the string appears in the array. So that one would be even better for cases like this.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+    }, 0);
+
+    let avg = total/occurances.length;
+    return avg;
+}
+
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies) {
+    let sorted = movies.sort(function(a, b){
```
IH-WebDev-TA-Remote, line 82: Remember .sort is a method that mutates the original. Because the iteration ask you to not mutate the original you need to start with a copy of the parameter movies and then work on that. Otherwise the reference points to the original array.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+        return a.year > b.year;
+    });
```
IH-WebDev-TA-Remote, line 84: if you are trying the reduced syntax of the sort method (one line) it should be something like: 
return a.year - b.year;
Instead of:
return a.year > b.year;

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+
+    return sorted;
+}
+
+
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies) {
+    let alpha = movies.sort(function(a,b){
+        return a.title.localeCompare(b.title);
```
IH-WebDev-TA-Remote, line 94: The sort method and the localCompare method, both can arrange things alphabetically, so using them both might be a little redundant.

IH-WebDev-TA-Remote:
Good work overall, I see you managed to solve most of the Lab! Keep up the good work :)

PR #1518, opened 2020-07-09
WebDev - Remote - July2020 Joel Roca Martinez

JoelRocaMartinez:
this lab is hell and this is just the begining LOL

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,108 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+function getAllDirectors(movies) {
+    let answer = movies.map(function(movie) {
+        return movie.director   
+    });
+    return answer
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(movies) {
+    
+    let result = movies.filter(function(movie) {
+        // Condition: Only drama movies & from Steven Spielberg
+        if (movies.director == "Steven Spielberg" && movies.genre.includes("Drama")) {
+            return movie
+        }
+    });   
+return result   
+}
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(movies) {
+    let answer = movies.reduce(function(acc, elem) {
+        return acc + elem.rate
+    }, 0)
+return answer / movies.length
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(movies) {
+    let drama = movies.genre.includes("Drama");
+    if (drama) {
+        let answer = movies.reduce(function(acc, elem) {
+            return acc + elem.rate
+        }, 0)
+        return answer / drama
+    }
+} 
+
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies) {
+    let newSortedArray = JSON.parse(JSON.stringify(movies))
+   
+    newSortedArray.sort(function(a, b) {
+        if (a.year > b.year) {
+            return 1;
+        } else if (a.year < b.year) {
+            return -1;
+        } else if (a.year = b.year) {
+            return newSortedArray.sort(function(a, b) {
+                if (a.title > b.title) {
+                    return 1;
+                } else if (a.title < b.title) {
+                    return -1;
+                } else {
+                    return 0;
+                }
+            })
+        }
+    })
+
+return newSortedArray;
+
+         
+}
+
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies) {
+    let sortedAlphabetically = JSON.parse(JSON.stringify(movies))
+    
+    sortedAlphabetically.sort(function(a, b) {
+        if(a.title > b.title) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,108 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+function getAllDirectors(movies) {
+    let answer = movies.map(function(movie) {
```
IH-WebDev-TA-Remote, line 5: Try to give your variables more meaningful names of what is the value. For this case, something like directorsArr could be.

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+    });
+    return answer
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(movies) {
+    
+    let result = movies.filter(function(movie) {
+        // Condition: Only drama movies & from Steven Spielberg
+        if (movies.director == "Steven Spielberg" && movies.genre.includes("Drama")) {
```
IH-WebDev-TA-Remote, line 16: instead of movies.director and movies.genre it should be movie.director and movie.genre . Remember movies is the original array being passed from the main function parameter, movie is each movie element being passed from the filter parameter.

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+            return movie
+        }
+    });   
+return result
```
IH-WebDev-TA-Remote, line 20: You need to return a quantity of elements, not the full array. For that use .length

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+function ratesAverage(movies) {
+    let answer = movies.reduce(function(acc, elem) {
+        return acc + elem.rate
+    }, 0)
+return answer / movies.length
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(movies) {
+    let drama = movies.genre.includes("Drama");
```
IH-WebDev-TA-Remote, line 33: The instructions say to return 0 if the array is empty. A single if condition at the beginning could do the trick.

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+    if (drama) {
+        let answer = movies.reduce(function(acc, elem) {
+            return acc + elem.rate
+        }, 0)
+        return answer / drama
```
IH-WebDev-TA-Remote, line 38: The returned number needed to be with only two decimals, to do this use toFixed and the turn in into a number.

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+        // Condition: Only drama movies & from Steven Spielberg
+        if (movies.director == "Steven Spielberg" && movies.genre.includes("Drama")) {
+            return movie
+        }
+    });   
+return result   
+}
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(movies) {
```
IH-WebDev-TA-Remote, line 24: The instructions say to return 0 if the array is empty. A single if condition at the beginning could do the trick.

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+    let answer = movies.reduce(function(acc, elem) {
+        return acc + elem.rate
+    }, 0)
+return answer / movies.length
```
IH-WebDev-TA-Remote, line 28: The returned number needed to be with only two decimals, to do this use toFixed and the turn in into a number.

src/movies.js
IH-WebDev-TA-Remote, line 33: movies is an array of elements. So you can't simply target the key genre of the array, for this you would need loop into each element of the array first.

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies) {
+    let newSortedArray = JSON.parse(JSON.stringify(movies))
+   
+    newSortedArray.sort(function(a, b) {
+        if (a.year > b.year) {
+            return 1;
+        } else if (a.year < b.year) {
+            return -1;
+        } else if (a.year = b.year) {
```
IH-WebDev-TA-Remote, line 52: instead of single equal, do triple equal. You are comparing two values, not assigning a new value.

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+            return newSortedArray.sort(function(a, b) {
+                if (a.title > b.title) {
```
IH-WebDev-TA-Remote, line 54: There is no need to do another sort inside the first sort. Try removing lines 53 (the sort) and 61 (the closing brackets of the sort) and it will work.

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+        } else {
+            return 0
+        }
+
+    })
 
+    let top20 = sortedAlphabetically.slice(0, 20);
+    console.log(top20)
+    
+ return sortedAlphabetically;
```
IH-WebDev-TA-Remote, line 89: you are creating the variable top20 of the 20 elements but then you are returning the whole sorted array.

IH-WebDev-TA-Remote:
Looks good overall but there is a lot of small mistakes that were breaking most of the tests. Try to go back to it now seen the solution and try it again. And if some test is not passing and you are sure you are doing the correct thing, try to read your code line by line and you will probably find those mistakes. See rubber duck debugging.
And yes... it is just the begining! :) #trustTheProcess

PR #1519, opened 2020-07-09
WEB-REM-2007 Bodei

BodeiBrouwer:
This is how far I got before Jasmine crashed. I tried relaunching many times but Jasmine kept loading and crashing.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,88 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(movies){
+    let answer1 = movies.map((movie) => {
+        return movie.director;
+    })
+return answer1;
+};
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies){
+    let drama;
+    let spiel = movies.filter((movie) => {
+    for (let i=0; i < movie.genre.length; i++){
+       drama = movie.genre[i] == "Drama" && movie.director == "Steven Spielberg";
+       if(drama){break;}
+    } 
+    return drama
+}) 
+return spiel.length;
+};
+
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+
+
+
+function ratesAverage(arrMovies) {
+    let sum = arrMovies.reduce(function(acc, elem){
+        return acc + elem.rate
+    }, 0)
+  let avgRate = sum / arrMovies.length
+  return avgRate
+};
+
+//return avgRate.toFixed(2);   breaks code
+
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(array) {
+    let drama = array.filter((element) => {
+    return element.genre == "drama";
+    })
+} 
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(movieArr) {
+    moviesCopy = JSON.parse(JSON.stringify(movieArr));
+    moviesCopy.sort (function(a,b){
+             if (a.year > b.year){
+                return 1;
+              }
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,89 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(movies){
+    let answer1 = movies.map((movie) => {
```
IH-WebDev-TA-Remote, line 4: Try to declare your variables with more meaningful names of what the value inside is. For example, directorsArr

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+        return movie.director;
+    })
+return answer1;
```
IH-WebDev-TA-Remote, line 7: Check indentation. This return should be at the same level as the last closing parentheses

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+};
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies){
+    let drama;
+    let spiel = movies.filter((movie) => {
+    for (let i=0; i < movie.genre.length; i++){
```
IH-WebDev-TA-Remote, line 17: The loop works well here, but you could use the method .includes to check the "Drama" inside the genre array. Looks more clean and you avoid running the loop.

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+return spiel.length;
+};
+
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+
+
+function ratesAverage(arrMovies) {
+    let sum = arrMovies.reduce(function(acc, elem){
```
IH-WebDev-TA-Remote, line 32: to check the jasmine test to return 0 if there is no movies you could start with a simple if condition that checks that and breaks the function by returning 0. (When a function encounters a return if will stop running the rest of the code inside.

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+        return acc + elem.rate
+    }, 0)
+  let avgRate = sum / arrMovies.length
+  return avgRate
```
IH-WebDev-TA-Remote, line 36: the value should be rounded to 2 decimals, you can achieve this by using the toFixed and parseInt methods.

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+
+//return avgRate.toFixed(2);   breaks code
+
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(array) {
+    let drama = array.filter((element) => {
+    return element.genre == "drama";
+    })
```
IH-WebDev-TA-Remote, line 47: Good start but the idea here was that you now applied the average to the drama array (either by creating new code or by reusing the ratesAverage function you created above.

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+                return 1;
+              }
+              else if(a.year < b.year) {
+                 return -1;
+              }
+              else {
+                return 0;
+              }
+        })
+}
```
IH-WebDev-TA-Remote, line 65: This part was really good! you just forgot to return the moviesCopy array :P

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(titles) {
+   for (let i=0; 1 <21 ; i++) {
```
IH-WebDev-TA-Remote, line 70: The problem here is that you are only checking the first 20 elements and ordering them. the idea was to order all items and then get the first 20. So for that you will need to do first a sort and the apply maybe a splice or a slice.

IH-WebDev-TA-Remote:
The logic overall seems good but try to go over the solution and check some extra tools that might have help you achieve the expected results, like toFixed, slice, splice, includes.
Also, the indentation in your code seemed to be a little off in some places so try to be more careful. Keep it up!

PR #1520, opened 2020-07-09
[WEB-REM-2007] Lanette Lanting

llanting:
Perfected iterations 1 till 5 as much as I could. Only got started on thinking about the logic for 7 & 8.

Not sure if my bonus for iteration 1 is working as it should, didn't see Jasmine for this.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -8,6 +8,17 @@ function getAllDirectors(movies) {
 
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+function getUniqueList(movies) {
+    let list = [];
+    for (let i = 0; i <movies.length; i++) {
+        let director = movies[i].director;
+        if (list.indexOf(director) < 0) {
+            list.push(director);
+        }
+    }
+    return list;
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
 function howManyMovies(movies) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -8,6 +8,17 @@ function getAllDirectors(movies) {
...
+function getUniqueList(movies) {
+    let list = [];
+    for (let i = 0; i <movies.length; i++) {
+        let director = movies[i].director;
+        if (list.indexOf(director) < 0) {
+            list.push(director);
+        }
+    }
+    return list;
+}
```
IH-WebDev-TA-Remote, line 20: This solution works perfectly but the idea was to use .map like we learn in the lectures. I would suggest you try this again using map so you get used to that method. You will be using that a lot moving forward :)

src/movies.js
```diff
@@ -1,16 +1,134 @@
...
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies) {
+    let drama;
+      let spiel = movies.filter(function(movie) {
+          for (let i = 0; i < movie.genre.length; i++) {
```
IH-WebDev-TA-Remote, line 27: Using a loop works fine here, but you could avoid it if you look into the .includes method for the genre key. The method can check if a string exist inside an array.

src/movies.js
```diff
@@ -1,16 +1,134 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(movies) {
+    let answer = movies.map(function(movie){
```
IH-WebDev-TA-Remote, line 3: Try to use more meaningful names for your variables, something that reflects the value that will be stored. For this iteration could be directorsArray or something like that.

IH-WebDev-TA-Remote:
Very good job overall! But don't be scared to try those bonuses, they are really challenging but even just by giving them a try it will be helpful for you :)

PR #1521, opened 2020-07-09
WDFT-REM-2007] BRITTA HELM

brittahelm:
got it to work in codepen but still a bit of red in jasmine on the bonus iterations, can't figure out what's not working there

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,149 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(arr) {
+    let newArr = arr.map(function(elem){
+        return elem.director;
+    })
+    return newArr;
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(arr) {
+    let spielbergArr = arr.filter(function(elem){
+        return elem.director == "Steven Spielberg" && elem.genre.includes('Drama');
+    })
+    return spielbergArr.length;
+}
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(arr) {
+    let rateArray = arr.filter(function(elem){
+        return elem.hasOwnProperty('rate') == true;
+    });
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,149 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(arr) {
+    let newArr = arr.map(function(elem){
```
IH-WebDev-TA-Remote, line 3: Try avoiding names like newArr when you can be little more descriptive when naming your variables, it is a small thing but helps greatly in the end! something like directorsArr could be better here :)

src/movies.js
```diff
@@ -1,16 +1,149 @@
...
+    return newArr;
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(arr) {
+    let spielbergArr = arr.filter(function(elem){
+        return elem.director == "Steven Spielberg" && elem.genre.includes('Drama');
```
IH-WebDev-TA-Remote, line 15: Perfect use of includes!

src/movies.js
```diff
@@ -1,16 +1,149 @@
...
+    })
+    return spielbergArr.length;
+}
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(arr) {
+    let rateArray = arr.filter(function(elem){
+        return elem.hasOwnProperty('rate') == true;
```
IH-WebDev-TA-Remote, line 24: Very clever using hasOwnProperty!

src/movies.js
```diff
@@ -1,16 +1,149 @@
...
+        return elem.title;
+    })
+    return titleArray.slice(0, 20);
+}    
+
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 
+
+function turnHoursToMinutes(arr) {
+
```
IH-WebDev-TA-Remote, line 105: Even though you are using .map to create a new array you are still accessing objects via the parameter elem. And because those point to the original objects you are mutating them, so either you will need to make a copy of the array before the map or make a copy of the object inside the map.

IH-WebDev-TA-Remote:
Amazing work! Keep it up :)

PR #1522, opened 2020-07-09
[WEB-REM-2007] RAFAEL DE LAZARO

rlaz133:
Still can't get to the bonus assignments on time. Will get there.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,112 @@
 // Iteration 1: All directors? - Get the array of all directors.
+const getAllDirectors =(movieArray) =>{
+    let directors = movieArray.map((movie)=>{
+    return movie.director;    
+    })
+    return directors;
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+const howManyMovies = (movieArray) =>{
+    let filmsbySteve = movieArray.filter((movie) =>{
+        return movie.director == 'Steven Spielberg' && movie.genre.includes('Drama');
+    })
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,112 @@
...
+    return directors;
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+const howManyMovies = (movieArray) =>{
+    let filmsbySteve = movieArray.filter((movie) =>{
+        return movie.director == 'Steven Spielberg' && movie.genre.includes('Drama');
```
IH-WebDev-TA-Remote, line 15: Good use of includes!

src/movies.js
```diff
@@ -1,16 +1,112 @@
...
+            return movie.title;
+    })
+    sortedArray.splice(20)
+    return sortedArray;
+}
+
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 
+//This crashes Jasmine, so I think one of the loops are infinite. Would finish with a reduce, but can't convert the strings into numbers properly.
+
```
IH-WebDev-TA-Remote, line 99: Yeah, something around that logic seems off.
First of all remember split does not mutate, so if you want to use you need to save the new array in a variable.
Also, that split will give you an odd result, for example on the duration "2h 55min", using .split("h" will give you an array of two elements. First one is "2" and second one is " 55min".
If you want to modify the hours then you will target the numbers separatly, not the elements on every iteration of the loop in the loop.

IH-WebDev-TA-Remote:
Very good job overall, very clean and easy to understand! Kudos!
Since I see you are already using arrow functions for the methods I would recommend you to check on their reduced syntax (one line). Practicing that would help you a lot along the way.

PR #1523, opened 2020-07-09
[WEB-REM-2007] Marta lourido

MartaLourido:
Here's my lab, I still have things to finish tomorrow since I haven't managed to do everything.
It's a bit challenging!
Hopefully, tomorrow I will resolve everything.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -23,46 +23,42 @@ function howManyMovies(movies) {
 
 function ratesAverage(movies) {
 
-    if (movies.length === 0) {
-      return 0
-    }
-    else { 
-        let total=movies.map(function(a,b){
-            return n*100;
-        }).reduce(function(a, b){
-            return a+(b || 0);
-        });
-        
-        Math.round((total/movies.length))/100;
-    }
-   
+  if (movies.length === 0) {
+    return 0;
+  }
+  let sumRates = movies.reduce(function(accumulator, film){
+    if (!film.rate){
+      return accumulator + 0 
+    } 
+      return accumulator + film.rate  
+  }, 0) / movies.length;
+
+  return Number((sumRates).toFixed(2));
 }
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
 function dramaMoviesRate(movies) {
-    let dramaMovies = movies.filter(movie => {
-      if (movies.length === 0) {
-            return 0
-      }
-          
-      if (movie.genre.indexOf('Drama') > -1) {
-        return true;
-      }
-       else if (movie.genre.includes('Drama')) {
-        return true;
-      }
-    });
-       if (dramaMovies.length) return 0;
-        return dramaMoviesRate;
-  }
+  let moviesd = movies.filter((movies) => movies.genre.includes("Drama"));
 
+  if (moviesd.length === 0) {
+    return 0;
+  }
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,115 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(movies) {
+    let alldirector = movies.map((movies) => movies.director);
```
IH-WebDev-TA-Remote, line 4: Perfect use of reduced map syntax!

src/movies.js
```diff
@@ -1,16 +1,115 @@
...
+  }
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies) {
+
+    let moviesd = movies.filter((movies) => movies.director === "Steven Spielberg" && movies.genre.includes("Drama"));
```
IH-WebDev-TA-Remote, line 15: Try to name your variables with more meaning. Instead of moviesd you could use filteredMovies or dramaSpielbergMovies.

src/movies.js
```diff
@@ -1,16 +1,115 @@
...
+    let thedramamovies = moviesd.length;
```
IH-WebDev-TA-Remote, line 16: As best practices, remember to make use of camel casing when naming your variables.

src/movies.js
```diff
@@ -1,16 +1,115 @@
...
+  }, 0) / moviesd.length;
+
+  return Number((sumRates).toFixed(2)); //we need 2 digits for get the float condition, so we need put 2 into the to.Fixed 
+}
+
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(movies) {
+    let gettheyears = [...movies] //it is the same that using JSON
```
IH-WebDev-TA-Remote, line 61: spread operator (those 3 dots) work similar to JSON.parse... , however it is not the same.
We will probably see more on spread operator later and explain better but the main difference is that spread makes a copy only of the first level of the array or object. Any nested elements will still point to the original reference. On the other hand, using JSON.parse... will make a full copy of all the nested elements, and none will point to the original reference. So even though it looks better using spread operator, it is better to use JSON.parse(JSON.stringify()) whenever you have nested elements.

src/movies.js
```diff
@@ -1,16 +1,115 @@
...
+      if (a.year > b.year) {
+        return 1;
+      } 
+      else if (a.year < b.year) {
+        return -1;
+      } 
+      else {
+        let firstMovie = a.title;
+        let secondMovie = b.title;
+        if (firstMovie > secondMovie) {
```
IH-WebDev-TA-Remote, line 72: creating the variables for a.title and b.title is fine but you can avoid some code by comparing those keys directly. So, something like (a.title > b.title)

IH-WebDev-TA-Remote:
Awesome job! aside from some small tips and feedback pieces I think everything was great :)
Keep up the good work and don't be afraid of those bonuses! even if you struggle a lot and don't manage to finish them you still learn a lot just by trying and failing.

PR #1524, opened 2020-07-09
[WDFT-REMOTE-2007] Alejandro Hernández

alejandrohj:
Hi TA,

Almost all the mandatory part done, I will try the rest tomorrow.

I have some problem on the iteration 6:
([(https://ironhack-5bv7024.slack.com/archives/C014U8HU5LJ/p1594336861019700)])

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,84 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(movies){
+       //code
+       let answer = movies.map((movie)=>{
+           return movie.director;
+       })
+       return answer;
+}
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(movies){
+    let dramaAndSteven;
+    let answer = movies.filter((movie)=>{
+        for(let i=0; i<movie.genre.length; i++){
+            dramaAndSteven = movie.genre[i] == "Drama" && movie.director == 'Steven Spielberg';
+          if(dramaAndSteven){break;}
+        }
+        return dramaAndSteven;
+    })
+    console.log(answer);
+    return answer.length;
+}
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(movies){
+    if(movies==0){return 0}
+    else{
+        let avrg;
+        let count = 0;
+        let answer = movies.reduce((acc, elem)=>{
+            if(elem!=0){
+                return acc + elem.rate;
+            }
+        },0)
+        avrg = answer/movies.length;
+        return parseFloat(avrg.toFixed(2));
+   }
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-
+function dramaMoviesRate(movies){
+    let dmSp;
+    let dramaMovies = movies.filter((movie)=>{
+        for(let i=0; i<movie.genre.length; i++){
+        dmSp = movie.genre[i] == "Drama";
+          if(dmSp){break;}
+        }
+        return dmSp;
+    })
+    return ratesAverage(dramaMovies);
+  
+}
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-
+function orderByYear(movies){
+    let byYear = movies.sort((a,b)=>{
+        if(a.year>b.year){return 1;}
+        else if(a.year<b.year){return -1;}
+        else{
+            if(a.title>b.title){return 1;}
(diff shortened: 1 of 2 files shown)


src/movies.js
```diff
@@ -1,16 +1,132 @@
...
+       return answer;
+}
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(movies){
+    let dramaAndSteven;
+    let answer = movies.filter((movie)=>{
+        for(let i=0; i<movie.genre.length; i++){
```
IH-WebDev-TA-Remote, line 15: The loop works good for this, but there is a method called .includes that looks for an element inside an array, so by using that you could avoid the loop. Will look more clean.

src/movies.js
```diff
@@ -1,16 +1,132 @@
...
+          if(dramaAndSteven){break;}
+        }
+        return dramaAndSteven;
+    })
+    console.log(answer);
+    return answer.length;
+}
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(movies){
+    if(movies==0){return 0}
```
IH-WebDev-TA-Remote, line 26: Remember to use triple equal instead of double equal as best practices.

src/movies.js
```diff
@@ -1,16 +1,132 @@
...
+    else{
+        let avrg;
+        let answer = movies.reduce((acc, elem)=>{
+            return acc + (elem.rate || 0);
```
IH-WebDev-TA-Remote, line 30: Perfect use of ||

src/movies.js
```diff
@@ -1,16 +1,132 @@
...
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-
+function dramaMoviesRate(movies){
+    let dmSp;
+    let dramaMovies = movies.filter((movie)=>{
+        for(let i=0; i<movie.genre.length; i++){
+        dmSp = movie.genre[i] == "Drama";
+          if(dmSp){break;}
```
IH-WebDev-TA-Remote, line 43: Just a tip: to reduce the syntax on conditionals even more you can remove the curly brackets if the statement is in one line.

src/movies.js
```diff
@@ -1,16 +1,132 @@
...
+            else{return 0;}
+        }
+    });
+    return byYear;
+}
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies){
+    let cloneMovie = JSON.parse(JSON.stringify(movies));
+    let the20first=[];
+    let titlesInAlphabeticalOrder = cloneMovie.sort((a,b)=>{
```
IH-WebDev-TA-Remote, line 68: Another tip, since you are using arrow functions, you can research the reduced syntax for the sort method when only used to do a simple sort. You could essentially turn 5 lines into 1 and still look very clean.

src/movies.js
```diff
@@ -1,16 +1,132 @@
...
+            else if(a.title<b.title){return -1;}
+            else{return 0;}
+        }
+    });
+    return byYear;
+}
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies){
+    let cloneMovie = JSON.parse(JSON.stringify(movies));
+    let the20first=[];
```
IH-WebDev-TA-Remote, line 67: if you are not pushing elements into this array, then there is no need to initialise it as empty. You could declare it in line 76 and put the spliced values in it.

src/movies.js
```diff
@@ -1,16 +1,132 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(movies){
+       //code
+       let answer = movies.map((movie)=>{
```
IH-WebDev-TA-Remote, line 4: Try to be more specific when naming your variables. Name then something that will reflect the value stored. For this, something like directorsArr.

src/movies.js
```diff
@@ -1,16 +1,132 @@
...
+        return movie
+    });
+    return filmDuration;
+}
 
 // BONUS - Iteration 8: Best yearly rate average - Best yearly rate average
+function bestYearAvg(movies){
+    let moviesByYear = {}
+    if(movies ==0){return null;}
+    else{
```
IH-WebDev-TA-Remote, line 111: This works perfect but you don't really need to attach the rest of the code to an else. To make it more clean you can just create a simple condition if (movies.length === 0) without an else and then continue with the rest of the code. If the condition happens, then the return inside will stop the rest of the code in the function from executing.

IH-WebDev-TA-Remote:
Amazing work on the lab. All tests are passing and everything is clean and easy to understand. I have thrown a couple of feedback and suggestions in the code itself but overall everything is great. Keep up the good work!

PR #1525, opened 2020-07-10
[WEB-REM-2007]Gunner Andersen

GunnerAg:
Still part not done, to be finished today

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,4 +1,4 @@
-// Iteration 1: All directors? - Get the array of all directors.
+/// Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 function getAllDirectors(movies){
     let directorsarray = movies.map(function(movie){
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,11 +1,34 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
-
+function getAllDirectors(movies){
+    let directorsarray = movies.map(function(movie){
```
IH-WebDev-TA-Remote, line 4: As best practices, remember to use camelCasing

src/movies.js
```diff
@@ -1,11 +1,34 @@
...
+          return movie.director
+    })
+  return directorsarray
+}
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(movies){
+    let SpielbergOnes = movies.filter(function(movie){
```
IH-WebDev-TA-Remote, line 11: Again, camelCasing . Variables should not start with uppercase.

src/movies.js
```diff
@@ -1,11 +1,34 @@
...
+        return movie.director == 'Steven Spielberg' && movie.genre.includes('Drama') == true;
```
IH-WebDev-TA-Remote, line 12: good use of .includes!

src/movies.js
```diff
@@ -1,11 +1,34 @@
...
+    })
+    return SpielbergOnes.length
+}
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(movies){
+    let avrg
+    let addallscores = movies.reduce(function(acc,movie){
+       return acc + movie.rate
+    },0)
+    avrg = addallscores/movies.length.toFixed(2)
```
IH-WebDev-TA-Remote, line 22: the toFixed is being applied to the movies.length and not to the division. You can wrap it in paranthesis to make it work.
Also, toFixed changes the number to a string, so you will need to convert it back to a number using one of several methods (Number. , parseFloat or the + symbol)

src/movies.js
```diff
@@ -1,11 +1,34 @@
...
+    return avrg
+} 
+
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(){
+    if(movies.genre.includes("Drama")){
```
IH-WebDev-TA-Remote, line 29: Make sense to use includes but there are several issues here:
- there is no parameter in the function.
- you need a map to target each movie object (similar to previous iterations)
- you are not returning anything in the function.

IH-WebDev-TA-Remote:
Good effort but there was still a lot to do in there. Only half of the iterations done and some issues, Please go back to the lab and finish at least the mandatory part. I will keep the lab open until you submit it again.

IH-WebDev-TA-Remote:
Awesome! Looks good now. Even if you do some iterations along with the solution, the important thing is that you understand everything that is happening in the code, how the methods work and that you can reapply that logic later.
Keep up the good work!

PR #1526, opened 2020-07-10
[WEB-REM-2007] Galicia

Srtagalicia:


src/movies.js
```diff
@@ -1,16 +1,95 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors (movies){
+    let directors = movies.map(function(movie){
+        return movies.directors;
```
IH-WebDev-TA-Remote, line 5: the return inside the map is having two issues.
- The property of the object is called director, not directors.
- It should be movie, not movies. Movies is the main function parameter and movie is the map parameter. To avoid this kind of issues you can avoid similar naming of your variables. So instead of movies do moviesArray and instead of movie do eachMovie.

src/movies.js
```diff
@@ -1,16 +1,95 @@
...
+    return directors;
+}   
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies){
+    let dramaMovies = movies.filter(function(movie){
+        movie.director == "Steven Spielberg" && movie.genre.includes("Drama");
```
IH-WebDev-TA-Remote, line 16: In these methods (map, filter, sort, etc...) you need to return something in the inner function. If this case you are evaluating the .director and .genre but you are not returning the evaluation. 
This might be confusing because when you reduce the syntax to a one line, you don't need to add the return keyword. However, when you do the map method statement in multiple lines inside curly brackets you need to add the return keyword.

src/movies.js
```diff
@@ -1,16 +1,95 @@
...
+    });
+    return dramaMovies.length;
+}
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movie){
```
IH-WebDev-TA-Remote, line 23: You are using movies during the function, however the parameter in this cased was named movie. You can note that the parameter is never being used because in your VS code it should look gray.

src/movies.js
```diff
@@ -1,16 +1,95 @@
...
+    }
+    let totalRates = movies.reduce(function(acc, movie){
+        return acc + movie.rate || 0            ;
+    }, 0);
+    return Number((totalRates/movies.length).toFixed(2));
+}
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+let dramaMovies = movies.filter((movie) => {
```
IH-WebDev-TA-Remote, line 35: This is what is breaking Jasmine. Remember that for these labs, all your code should be inside functions or objects. But declaring variables outside and doing returns it is breaking the testing system.

src/movies.js
```diff
@@ -1,16 +1,95 @@
...
+    if (a.title > b.title){
+        return 1;
+    }
+    else if (a.title < b.title){
+        return -1;
+    }
+    else {
+        return 0;
+    }
+    let movieTitle = cloneMovies.map(function (movie){
```
IH-WebDev-TA-Remote, line 77: This part of the code is never running (you can see a hint about it because in your code it should look a darkened color. The reason is that after a return (which is happening in any case of the conditional above) the code inside a function won't run. Return is the last thing that will run in a function. If you want this code to run it should be outside the map.

src/movies.js
```diff
@@ -1,16 +1,95 @@
...
+    let dramaMovies = movies.filter(function(movie){
+        movie.director == "Steven Spielberg" && movie.genre.includes("Drama");
+    });
+    return dramaMovies.length;
+}
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movie){
+    if (movies.length == 0){
```
IH-WebDev-TA-Remote, line 24: for best practices, always use triple equal instead of double equal when comparing values.

IH-WebDev-TA-Remote:
Good job overall, however you need to be extra careful that the variables and properties you are using are the correct ones. Whenever something is not working correctly try to read your own code line by line to figure out what might be causing the issue.
Keep it up!

PR #1527, opened 2020-07-10
first commit

gonzalo-development:
Unfortunately yesterday was not able to reach the target well(needed index), I knew most methods needed, but did mix loops in a disorder way. After today lesson, it all came up great

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,11 +1,53 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(movies){
+
+const directorF =  movies.map(function(movie){
+        return movie.director
+    })
+
+return directorF
+}
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies){
+    let drama;
+    let spiel = movies.filter((movie) => {
+    for (let i=0; i < movie.genre.length; i++){
+       drama = movie.genre[i] == "Drama" && movie.director == "Steven Spielberg";
+       if(drama){break;}
+    } 
+    return drama
+}) 
+return spiel.length;
+};
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movies){
+    let averageRate = 0;
+    movies.reduce(function(acc,value){
+        for(var i=0; i<movie.rate.length; i++){
+          averagerRate =  acc + value(movie.rate[i])  
+        }
+        return averageRate 
+    },0)
+    return averageRate.length;
+}
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(movies){
+    let averageDramaRate = 0;
+    movies.map(function(ele){
+        if(ele.genre == "drama"){
+
+        }
+
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
@@ -14,3 +56,4 @@
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 
 // BONUS - Iteration 8: Best yearly rate average - Best yearly rate average
+
--- a/src/movies.js
+++ b/src/movies.js
@@ -26,34 +26,88 @@ function howManyMovies(movies){
 return spiel.length;
 };
 
+function howManyMovies(movies){
+    let Drama = movies.filter(function(ele){
+      return ele.genre.includes("drama")
+    })
+    return Drama;
+  }
+
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
 function ratesAverage(movies){
-    let averageRate = 0;
-    movies.reduce(function(acc,value){
-        for(var i=0; i<movie.rate.length; i++){
-          averagerRate =  acc + value(movie.rate[i])  
-        }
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+    for (let i=0; i < movie.genre.length; i++){
+       drama = movie.genre[i] == "Drama" && movie.director == "Steven Spielberg";
+       if(drama){break;}
+    } 
+    return drama
+}) 
+return spiel.length;
+};
+
+function howManyMovies(movies){
```
IH-WebDev-TA-Remote, line 29: This is another function with the same name as before, which is causing issues. It will only run the second one, and in this case, this one is not checking for the director. If you comment or remove it the tests for the howManyMovies function will pass.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movies){
+  
+    let sum = movies.reduce(function(acc,value){
+      return acc + value.rate
+    },0)
+    return sum/sum.length;
```
IH-WebDev-TA-Remote, line 44: sum is the total value (a number), to get the average you need to divide by the amount of elements of the array (movies). sum.length will give you undefined.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+  }
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(movies){
+    let s = movies.filter(function(elw){
+      return elw.genre.includes("drama")
+    })
+    return rateAverage(s);
```
IH-WebDev-TA-Remote, line 52: Good that you reused the ratesAverage function but because that one is not working correctly you won't get the correct result here.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+  
+    let sum = movies.reduce(function(acc,value){
+      return acc + value.rate
+    },0)
+    return sum/sum.length;
+  }
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(movies){
+    let s = movies.filter(function(elw){
```
IH-WebDev-TA-Remote, line 49: As best practices, avoid these kind of names for your variables. Use names that reflect the expected value that will be stored inside. This is very important for other developers checking on your code and even for you when you go back to your code.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+   
+    let clone = JSON.parse(JSON.stringify(movies))
+    
+    let order =clone.sort(function(a,b){
+      if( a.year > b.year){
+        return 1;
+      }else if( a.year < b.year){
+        return -1;
+      }else{
+        return 0;
```
IH-WebDev-TA-Remote, line 67: To pass the test to order alphabetically if they have the same year, instead of returning 0 here you can add a new conditional that checks their title.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+      }
+    })
+    return order;
+  }
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies){
+    let cloneTitle = JSON.parse(JSON.stringify(movies));
+   
+   let orderTitle = cloneTitle.sort(function(){
```
IH-WebDev-TA-Remote, line 77: You forgot to add the a and b parameters to the sort function. After adding those almost all the tests will pass.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+     
+     if( a.title > b.title){
+       return 1;
+     }else if( a.title < b.title){
+       return -1;
+     }else{
+       return 0;
+     }
+     })
+   return orderTitle.splice(0,20)
```
IH-WebDev-TA-Remote, line 87: orderTitle is an array of objects. The expected result as per instructions is an array of strings. To do that you can use a map on the orderTitle to get only the titles, similar to how you returned an array of directors on the first iteration.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+   }
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
+function turnHoursToMinutes(movies){
```
IH-WebDev-TA-Remote, line 91: This function is having a syntax error that is breaking Jasmine. If you solve the issue or comment it Jasmine will run again.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+        let min= time[1].split("min");//string ["20", ""]
+        let sum = Number(hours[0])*60 + Number(min[0]);
+      }else if(movie.duration.includes('h')){ 
+        let hours= movie.duration.split('h');// string ["1", ""]
+        movie.duration = Number(hours[0])*60;
+      }else if(movie.duration.includes('min') ){
+         let min= movie.duration.split('min');
+        movie.duration = Number(min[0]);
+      }
+      return movie,
```
IH-WebDev-TA-Remote, line 107: And here is the syntax error. the comma at the end of this line.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+      })
+    return updateTime;
```
IH-WebDev-TA-Remote, line 109: Overall it was a good effort on solving the iteration,. Logic looks good but the conditional could have been implemented better. 
You have the sum variable with the result in minutes but you are never modifying the object movie that is being returned. Reason is that the condition is being resolved on the first if, because all duration values have a " " so the next two will never run. If you want those to run they should be independent conditionals from the first one.

IH-WebDev-TA-Remote:
Good job overall, logic seems good but you were having some issues with syntax and following along the expected results from the labs. Keep it up.

PR #1528, opened 2020-07-10
[WDFT-REM-2007] Maeva Merrou

maevamerrou:
finally worked !

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,112 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+
+let getAllDirectors = movies.map(function(movie){
+  return movie.director;
+});
+
+
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+
+
+  let howManyMovies= movies.filter(function(movie){
+    if (movie["genre"].includes("Drama") && movie["director"] === "Gary Trousdale")
+       return movie;
+   });
+
+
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,112 @@
...
+
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+
+let ratesTotal = movies.reduce(function(acc, movie){
+  return acc + movie.rate
+},0);
+
+console.log((ratesTotal / movies.length).toFixed(2));
```
IH-WebDev-TA-Remote, line 28: toFixed returns a string. You would need to do something else to return a number instead ;)

src/movies.js
```diff
@@ -1,16 +1,112 @@
...
+},0);
+
+let dramaMoviesRate = (ratesDrama / dramaMovies.length).toFixed(2)
+
+
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+
+let orderByYear = movies.sort(function(a, b){
```
IH-WebDev-TA-Remote, line 52: .sort mutates the original array, so in order to not mutate it (as per instructions) you would need to first create a deep copy of the array

src/movies.js
```diff
@@ -1,16 +1,112 @@
...
+   return -1;
+ } if (a.title > b.title){
+   return 1;
+ } else {
+   return 0;
+ }  
+});
+
+
+let twentyFirst = alphabeticalOrder.filter(function(movie){
```
IH-WebDev-TA-Remote, line 84: A filter works perfectly fine but a suggestion would be to use splice or slice instead, since you already know the indexes to be cut from the array.

IH-WebDev-TA-Remote:
Great work on the lab, the logic is very good and the code is clean. Kudos!
But still, it is important that you know how Jasmine works testing the functions you create, so if possible try to put go over the lab again and try to turn the code into functions to see the tests passing. Keep up the good work!

PR #1530, opened 2020-07-13
Marcella - Amsterdam wd pt June 2020

allecramarcella:
iteration 1-5 together with Yuka, rest alone

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -5,6 +5,15 @@ function getAllDirectors(movieArr){
   } 
 
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+function getAllDirectors(movieArr){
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,140 @@
...
+      return alpha
+     }) 
+     byYear = sortedYears
+    return byYear
+}
+    
+orderByYear(movies)
+
+
+// failed attempt to fix error (should return new array) in Jasmine
```
guido-an, line 80: When using sort() your are mutating the original array so you should create a safe copy of the array before sorting, for example: 
`function orderByYear(arr) {`
  `let byYear = [...arr];`
`  byYear.sort((a, b) => {`
   ` let alpha = a.year - b.year `
    `if (alpha === 0){`
   `   return a.title.localeCompare(b.title)`
  `  }`
`    return alpha`
  ` }) `
`  return byYear;`
`}`

PR #1531, opened 2020-07-13
[WDPT - AMS - 0620 Yuka]

yukajyeh:


src/movies.js
```diff
@@ -1,15 +1,114 @@
...
+     }) 
+      return [...sortedYears]
+    }
+    orderByYear(movies) 
+    
+
+
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(arr){
+  {
```
guido-an, line 105: This parenthesis in not needed

guido-an:
Really good Yuka!

src/movies.js
yuka210: Fixed 👯

PR #1532, opened 2020-07-14
done

Lisette-Sneha:


src/movies.js
```diff
@@ -1,16 +1,109 @@
...
+  return onlyTitle.length
+}
+// console.log(howManyMovies(movies))
 
-// Iteration 3: All rates average - Get the average of all rates with 2 decimals
+// // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage (array) {
+  const sumRatingCalc = array.reduce((total, movie) => {
+    return (total + (movie.rate || 0)) // Does this make sense??? [6, , ] should average out to 6 not 2
```
guido-an, line 28: This can be another solution: 
` const sumRatingCalc = array.reduce((total, movie) => {`
    `if (movie.rate) {`
    `  return total + movie.rate`
    `} else {`
  `    return total`
   ` }`
`  }, 0)`

PR #1533, opened 2020-07-14
[WDPTAMS0620] Anna Dorenskaia

AnnaDore:
Hi Guido!
I couldn't pass 3 tests for the mandatory tasks + I was fighting with the bonuses but couldn't solve even a part of it. Could you, please, share the code (if it is possible)? I would like to check the logic and put the code in a code visualizer. Thank you in advance! =)


guido-an:
Hi Anna, I am going to share with you other solutions for this exercise :)

PR #1553, opened 2020-08-14
WEBDEV-Berlin-Kevin Klein-fulltime

pkevinklein:


src/movies.js
```diff
@@ -1,16 +1,194 @@
...
+  let allDirectors = [];
+  allDirectors = moviesData
+  .map(function (movie){
+    
+    return movie.director;
+  })
+  .filter(function(d, i, self){
+    return i == self.indexOf(d);
+  });
+return allDirectors;
```
timbotimber, line 12: Nice! I think you could simplify this slightly without the second part. Also it's good practice to name the variables as opposed to using letters. I don't know what each of these values is doing!

PR #1555, opened 2020-08-14
WDFT_Berlin_GreatestMovies_DanParry

DanP20:


src/movies.js
```diff
@@ -1,15 +1,65 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(movies) {
+    let directors = movies.map(function(i) {
+        return i.director;
+    }) 
+    return directors;
+}
+
```
Tyouxik, line 8: Use clear variable names (eg. movie) instead of i, it will make you code more readable for you and other programmers.

PR #1558, opened 2020-08-15
[WDFT Berlin 08.20] Hamlet Hayrapetyan

Prime-arm:


src/movies.js
```diff
@@ -1,16 +1,102 @@
...
+        return movie.rate;
+      });
+      const sum = rating.reduce((max, val) => {
+      return max + val;
+    }); 
+      total = (sum / movAvg.length).toFixed(2);
+      return Number.parseFloat(total);
+
+    }
+
```
timbotimber, line 46: nice. You don't have to declare an empty array to begin with, you can just say let `rating = movAvg.map... `if you like

src/movies.js
```diff
@@ -1,16 +1,102 @@
...
+    //     return 0;
+    // }
+    // total = 0
+    // let rating = [];
+
+    // rating = dramAvg.map(funtion(movie) {
+    //     if 
+    // })
+}
+
```
timbotimber, line 60: Here you can filter the origin array to find the movies that have the drama genre then reuse your function from the previous iteration

PR #1559, opened 2020-08-15
BER_WEB-DEV_08 2020 Habid Badillo Espinosa

habidbesp:


src/movies.js
```diff
@@ -1,16 +1,78 @@
...
+      
+    }) 
+      return findDirDram.length
+  }
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movies){
+    const sumOfrates = movies.reduce(function(total, val){
+      if(val.rate === ''|| val.rate === undefined){
```
timbotimber, line 29: Don't think you need the or in here. Could just use `if(!val.rate)` which just means if the value gives you a falsy value.

PR #1563, opened 2020-08-16
[Berlin-FT-082020] Tolga Kara

TolgaKara:
Couldn't solve the Bonus - Iteration 8.
The rest was ok but some tests are confusing.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,103 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(movieArray) {
+	const directors = movieArray.map((x) => {
+		return x.director
+	})
+	/* console.log(directors) */
+	return directors
+}
+/*
+let director = getAllDirectors(movies)
+console.log(director) */
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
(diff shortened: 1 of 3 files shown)


src/movies.js
```diff
@@ -1,16 +1,152 @@
...
+		if (!val.rate) {
+			return acc + 0
+		}
+
+		return acc + val.rate
+	}, 0)
+
+	/* console.log(avgRates) */
+
+	return Math.round((avgRates / movies.length) * 100) / 100
```
Jeff-Moraes, line 48: you could also use the toFixed method 
```
+((avgRates / movies.length).toFixed(2))
```

src/movies.js
```diff
@@ -1,16 +1,152 @@
...
+		.sort(function (title1, title2) {
+			if (title1 < title2) {
+				return -1
+			}
+			if (title1 === title2) {
+				return 0
+			}
+			if (title1 > title2) {
+				return 1
+			}
```
Jeff-Moraes, line 94: here you can also use the localeCompare method
```
return title1.localeCompare(title2);
```

src/movies.js
```diff
@@ -1,16 +1,152 @@
...
+	})
+
+	let yearAvg = movies.reduce((acc, val) => {
+		if (acc) {
+			acc.push(val)
+		}
+	}, [])
+
+	return yearArray
+}
```
Jeff-Moraes, line 152: for this bonus iteration, you can check this solution
```
            return m.year === movie.year
        }));
            best.year = movie.year;
            best.rate = average;
        }
        return best;
}
```

PR #1564, opened 2020-08-16
Daniel Papastratidis Berlin WDFT

idanpap:


src/movies.js
```diff
@@ -1,16 +1,123 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(movies) {
+    const directorsList = movies.map(function(name) {
+            return name.director
+        })
```
Tyouxik, line 6: You should name your variable to what they are. If you map through an array of movies, each item is a movie (not a name). It will help you and others reading your code more easily.

src/movies.js
```diff
@@ -1,16 +1,123 @@
...
+    }
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(movies) {
+    const listOfSpielbergDramas = movies.filter(function (movie) {
+        if (movie.director === 'Steven Spielberg' && movie.genre.includes('Drama')) {
+          return true
+        }
```
Tyouxik, line 22: With filter, you only need a return statement with your filtering logic.
movies.filter(function (movie) {
        return movie.director === 'Steven Spielberg' && movie.genre.includes('Drama')
        })
Filter will return all the element that match movie.director === 'Steven Spielberg' && movie.genre.includes('Drama')

src/movies.js
```diff
@@ -1,16 +1,123 @@
...
+        if( a.title > b.title) {
+          return 1
+        }
+      }
+        return a.year - b.year
+    })
+    return sorted
+}
+
```
Tyouxik, line 73: This function can be simplified :)

src/movies.js
```diff
@@ -1,16 +1,123 @@
...
+            movie.duration = Number (movie.duration.replace("h", "") * 60)
+            } else if (!movie.duration.includes('h')) {
+              movie.duration = Number (movie.duration.replace("min", ""))
+            } else {
+        movie.duration = Number (movie.duration.replace('h', "").replace('min', '')[0].split(" ")*60) + Number (movie.duration.replace('h','').replace('min',"").split(' ')[1])}
+        return movie
+      })
+      return moviesWithDuration;
+  }
+
```
Tyouxik, line 101: Good approach :)

PR #1566, opened 2020-08-16
202008 - WDFS - FT - BER - Marlena Kowieska

marlykiwi:


src/movies.js
```diff
@@ -1,16 +1,163 @@
...
+            if (a.title < b.title) {
+                return -1;
+            }
+            return 0;
+        }
+        return a.year - b.year;  
+    }); 
+    return oldestFirst;
+}
+
```
timbotimber, line 80: again here, you don't need to add the extra functionality in the original method to get the same answer, for example: 

```
function sortIt(arr) {
  return arr.sort((a,b) => 
    a - b
  )
}
         
sortIt([1,5,7,9,6,2,5,7,])
```

src/movies.js
```diff
@@ -1,16 +1,163 @@
...
+            }
+            newArray.push({
+                ...item,
+                duration: +totalTime
+            })
+    }) 
+    return newArray;
+}
+
+
```
timbotimber, line 162: really impressive logic!

PR #1568, opened 2020-08-16
Khushboo

khushboo026:


src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(movies){
+    const moviesOfSt = movies.filter(function(mov){
+        count = 0;
+    if((mov.director === 'Steven Spielberg') && ((mov.genre[1]==='Drama')||mov.genre[0]==='Drama'||mov.genre[2]==='Drama'))
+        return movies;
```
timbotimber, line 17: you could use a` .includes() ` here to avoid the chained 'ors'

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+
+// // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(movies){
+    if(movies.length === 0) return 0;
+    const num = movies.reduce(function(acc,val){
+        if(val.rate) {
+        return acc + val.rate;
+        }else{ 
+        return acc
```
timbotimber, line 30: real nice logic

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+    })
+    if (sortedArray.length <= 20) return sortedArray 
+    else {
+    const newArray =[];
+   for(let i=0;i<20;i++){
+       newArray.push(sortedArray[i]);
+   }
+   return newArray;
+}
+}
```
timbotimber, line 73: Nice! Could shorten it a bit with slice as opposed to the loop to make the solution a bit neater

PR #1571, opened 2020-08-17
WDFT - BERLIN - Yoseob Shin - done but without bonuses

shinyo627:


src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+    howManyMovies(movies);
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+//summed value / number of elements
+function ratesAverage(movies) {
+    let sum = movies.reduce((total, rates) => {
+      return total + rates.rate; 
+    }, 0)
+    let average = sum / 250;
+    return Number(average.toFixed(2))
+}
```
Jeff-Moraes, line 33: here you needed to add an if statement to check if the movies array is empty so then you should return 0.
instead of hard coding to get the average like = sum / 250 you could have used the length of the movies array.
and to sum only the movie rate of the ones that have a rate you could add the if statement inside of the reduce method.
```
  let sum = movies.reduce((total, rates) => {
    if (rates.rate) {
      return total + rates.rate;
    } else {
      return total;
    }
  }, 0);
  let average = sum / movies.length;
  return Number(average.toFixed(2));
}
```

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+    }
+  }
+  return movieTitles;
 
-// BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
+};
+orderAlphabetically(movies)
 
+
+// BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
```
Jeff-Moraes, line 106: here is one solution for this bonus
```
    return movies.map(function (movie) {
        let convertedDuration = 0;
        if (movie.duration.includes('h')) {
            convertedDuration += Number(movie.duration.slice(0, movie.duration.indexOf('h'))) * 60;
        }
        if (movie.duration.includes('min')) {
            convertedDuration += Number(movie.duration.slice(movie.duration.indexOf(' ') + 1, movie.duration.indexOf('m')));
        }
        return { ...movie, duration: convertedDuration };
    });
}
```

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
+// function turnHours 
 // BONUS - Iteration 8: Best yearly rate average - Best yearly rate average
```
Jeff-Moraes, line 108:
```
            return m.year === movie.year
        }));
            best.year = movie.year;
            best.rate = average;
        }
        return best;
}
```

src/movies.js
```diff
@@ -1,16 +1,108 @@
...
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies) {
+    // returns array of first 20 titles alphabetically ordered
+    // return only title key
+    // if less than 20 movies then return all
+    let listOfMovies = [...movies];
+// below returns alphabetically arranged movie objects
+    listOfMovies.sort(function(a, b) {
+    var textA = a.title.toUpperCase();
+    var textB = b.title.toUpperCase();
```
Jeff-Moraes, line 87: remember that it is always recommended to use let or const instead of var. ;)

PR #1572, opened 2020-08-17
Berlin - FT - Veronica De Leon

veronicadeleonh:


src/movies.js
```diff
@@ -1,14 +1,88 @@
...
+  const sortedByYear = [...movies].sort(function (movie1, movie2) {
+    return movie1.year - movie2.year;
+    // if (movie1.year < movie2.year) {
+    //   return -1;
+    // }
+    // if (movie1.year > movie2.year) {
+    //   return 1;
+    // }
+    // return 0;
+  });
```
Jeff-Moraes, line 80: to pass on this test "If two movies have the same year, order them alphabetically by their title" you can add an if statement to your sort function
```
   if (movie1.year !== movie2.year) {
    } else {
    }
```

src/movies.js
```diff
@@ -1,14 +1,88 @@
...
+  return sortedByYear;
+}
+
+orderByYear(movies);
+
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
```
Jeff-Moraes, line 86: here is a solution for this iteration
```
    return movies.map(function (movie) {
        return movie.title;
    }).sort(function (a, b) {
        return a.localeCompare(b)
    }).slice(0, 20);
}
```

src/movies.js
```diff
@@ -1,14 +1,88 @@
...
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
```
Jeff-Moraes, line 88: and these are the solutions for the bonus
``` 
    return movies.map(function (movie) {
        let convertedDuration = 0;
        if (movie.duration.includes('h')) {
            convertedDuration += Number(movie.duration.slice(0, movie.duration.indexOf('h'))) * 60;
        }
        if (movie.duration.includes('min')) {
            convertedDuration += Number(movie.duration.slice(movie.duration.indexOf(' ') + 1, movie.duration.indexOf('m')));
        }
        return { ...movie, duration: convertedDuration };
    });
}
```
````
            return m.year === movie.year
        }));
            best.year = movie.year;
            best.rate = average;
        }
        return best;
}
````

PR #1574, opened 2020-08-17
WDFT- 08-2020-Daniela 

daniela-kirilova:


src/movies.js
```diff
@@ -1,15 +1,95 @@
-/* eslint no-restricted-globals: 'off' */
+function getAllDirectors(arr) {
+  return arr.map((movie) => {
+    return movie.duration;
+  });
+}
```
Tyouxik, line 5: Careful, you are returning the duration instead of the directors.

src/movies.js
```diff
@@ -1,15 +1,95 @@
...
+  let avarage = sum / movieArray.length;
+  return Number(avarage.toFixed(2));
+}
+
 // Iteration 5: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(moviesArray) {
+  if (moviesArray.length === 0) {
+    return 0;
+  }
```
Tyouxik, line 69: Here your if statement check if the array with all the movies is empty, however the test ask you to check if the array of drama movies made by Spielberg is empty.

PR #1586, opened 2020-08-21
[WDFT SAO 08.2020] Camila Fassarella

CamilaFass:
done, but nothing done yet, only iteration 1. Will update the rest this weekend.


CamilaFass:
done. pending bonus only.

PR #1589, opened 2020-08-23
WDFT-Berlin-Katrin

pineapplepan:


src/movies.js
```diff
@@ -1,17 +1,142 @@
 // Iteration 1: All directors? - Get the array of all directors.
 
+function getAllDirectors(movies) {
```
DavesEspinosa: Really good use of the map method with an arrow function, well done

src/movies.js
```diff
@@ -1,17 +1,142 @@
...
+    const mydirectors = movies.map((x) => x.director);
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+return mydirectors.filter((x, index) => mydirectors.indexOf(x) === index);
```
DavesEspinosa: It will more clear I guess, if instead of (x) you put exactly what are you looking for, but anyways, good work!

src/movies.js
```diff
@@ -1,17 +1,142 @@
...
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies) {
+    const spielbergMovies = movies.filter(
+        (x) => x.director == "Steven Spielberg" && x.genre.includes("Drama")
```
DavesEspinosa: The comparison there has to be "===", just to say, but still working, nice!

PR #1613, opened 2020-08-28
[PAR-WDFT-248] Jorell & Philippe

JorellLangmar:


N10T:
Hello Jorell & Phil,

**Nice commits** 🥇 

 
**iteration 1**
- 
- 👍 
- shorter solution : const getAllDirectors = movies => movies.map(movie => movie.director)

**iteration 2**
-
- 👍 

**iteration 3**
-
- 👍 

**iteration 4**
-
- re-use of the ratesAverage function 💯  

**iteration 5**
-
- const copy = [...movies] this is a copy of the array, but the movies are the same objects and not new objects
👍 

**iteration 6**
-
- forEach is for looping without changing the new array, map do the update👍 
- shorter way, chain functions :
const orderAlphabetically = movies =>[...movies].sort((a, b) =>a.title.localeCompare(b.title)).slice(0, 20).map((movie) => movie.title)

**Bonus 7 🥇 **
-
- No bonus

Good job 💯

PR #1615, opened 2020-08-29
[PAR-WDFT-248] Bouziane Bey / Vincent Proum

Bouziane95:


N10T:
Hello Bouziane & Vincent,

**Nice commits** 👍 


**iteration 1**
- 
- 👍 
- let unique = [...new Set(result)]; 👍 
- var copy = [...movies] thisis a copy of the array but the object are the same and keep the reference of the original element
=> var copy = movies.map(movie => { return {...movie} })

**iteration 2**
-
- 👍 
- isSspMovie is not super explicit isSpilbergMovie maybe

**iteration 3**
-
- 👍 

**iteration 4**
-
- 👍 

**iteration 5**
-
- nice splitting of the function 👍 

**iteration 6**
-
- 👍 
shorter way, chain functions :
const orderAlphabetically = movies =>[...movies].sort((a, b) =>a.title.localeCompare(b.title)).slice(0, 20).map((movie) => movie.title)

**Bonus 7 🥇 **
-
- good job 👍 
- first iteration commit is the answer for the copy

**Bonus 8 🥇 **
-
- you just started it :)


Good job 💯

PR #1616, opened 2020-08-30
[PAR-WDFT-248] MAXIMILIAN GRUBER

MaxJGruber:
All main iterations completed and tested.

Thanks!

Max


N10T:
Hello Max,

**Best practice:  commit for each iteration => "iteration 1 | done"**


**iteration 1**
- 
- 👍 
- x["director"] this way to access to a key of the object is useful with a variable like x[param]
- "x" is not explicit, prefer "movie" in this case
- with a fat arrow function : const getAllDirectors = arr => arr.map(x => x["director"])

**iteration 2**
-
- 👍

**iteration 3**
-
- 👍 

**iteration 4**
-
- 👍 

**iteration 5**
-
- nice splitting of the function 👍 

**iteration 6**
-
- 👍 
shorter way, chain functions :
const orderAlphabetically = arr => arr.map((x) => x["title"]).sort().slice(0, 20)

**Bonus 🥇 **
-
- no bonus :'(

Good job 💯

PR #1617, opened 2020-08-30
[PAR-WDFT-248] Laura Caillaux

Lcdevi:


N10T:
Hello Laura,

**Nice commits** 👍 


**iteration 1**
- 
- 👍 
-with a fat arrow function :   const getAllDirectors = array => array.map(film => film.director)

**iteration 2**
-
- "array" is scoped to the function, in this case, you can use "array in all of your functions
- 👍 

**iteration 3**
-
- 👍 

**iteration 4**
-
- 👍 

**iteration 5**
-
- 👍 
- My proposal of your function to be more readable
function orderByYear(array5) {
  var sortYears = [...array5].sort(function (a, b) {
    if (a.year > b.year) return 1;
    if (a.year === b.year) {
      if (a.title < b.title) return -1;
      if (a.title > b.title) return 1;
      return 0;
    }
  });
  return sortYears;
}

**iteration 6**
-
- 👍 
shorter way, chain your functions :
function orderAlphabetically(array6) {
  return [...array6]
    .map((str) => str.title)
    .sort(function (a, b) {
      if (a < b) return -1;
      if (a > b) return 1;
      return 0;
    })
    .slice(0, 20);
}

**Bonus 7 🥇 **
-
- good job 👍 
- to copy an object, and not the array arround you need to do it for each object
const copyBonusTime = bonusTime.map(object => {return {...object}})

**Bonus 8 🥇 **
-
- not done :)


Good job 💯

PR #1619, opened 2020-08-30
[PAR-WDFT-248] Frederico RATSIMANDRESY & Marina FUXA

Ratsimandresy:
first commit, done without bonus yet!


Fanny-Sbn:
Hello Rico and Marina. Thanks for submitting your lab.

**Iteration 1**
-
It's working but the goal of the exercise was to use an HOF, try to get the same result using .map

**Iteration 2**
-
Good
It's not compulsory to declare isSteven and IsDrama variables, you can directly return the comparison between both. 

(With a fat arrow function):

const howManyMovies = movies => {
  const stevenDramaMovies = movies.filter(movie => movie.director === "Steven Spielberg" && movie.genre.includes("Drama"));
  return stevenDramaMovies.length;
}

**Iteration 3**
-
Use an if, else statement. You don't need to use two times the Number() function in your return : 

return Number((sumOfMovieRates / lengthMovie).toFixed(2));

**Iteration 4**
-
👍 

**Iteration 5**
-
Nice 👍 
You just forgot a .year after movie2 line 75
Be careful with chained ternary operator, it can be hard to read sometimes.

**Iteration 6**
-
It's working but your function is overcomplicated. 
Try to simplify it, you don't need the first statement and don't forget to not mutate the original array. 

Good job guys ! 👍

PR #1620, opened 2020-08-31
[PAR - WDFT - 248] Lisa PONCET & Timothée NICOLE

lisapnct:
We didn't manage to do the last part of the 5th iteration :( 
Our idea was to sort movies alphabetically first and then based on the year but we were not able to order them alphabetically...


Fanny-Sbn:
Hello Lisa and Timothée ! Thanks for submitting your lab. 

**Iteration 1**
-
👍 

**Iteration 2**
-
👍 

**Iteration 3**
-
👍 Good
Just for you to know you can chain HOF : 

  if (movies.length === 0) {
    return 0;
  } else {
    var rateExists = movies
    .filter((movies) => movies.rate >= 0)
    .reduce(
      (acc, rateExists) => (acc += rateExists.rate),
      0
    );
    var averageRates = rateExists / movies.length;
    return Number(averageRates.toFixed(2));
  }
}

**Iteration 4**
-
👍  think about reusing your ratesAverage() function, you'll save a few lines. 

**Iteration 5**
-
You need to sort by year first and if two films have the same year, order them alphabetically. You try to sort alphabetically first, that's why is not working. 

**Iteration 6**
-
👍 

Good job guys ! Try to redo iteration 5 and maybe the bonuses too if you have time !

PR #1623, opened 2020-09-01
[PAR-WDFT-248]Tatiana Matzenbacher

tatimatz:


Fanny-Sbn:
Hello Tatiana ! Thanks for submitting your lab 😄 

**Iteration 1**
-
👍 
-with a fat arrow function : const getAllDirectors = array => array.map(film => film.director)

**Iteration 2**
-
👍 

**Iteration 3**
-
It's not working because we need to convert the rate which are in strings in number 

**Iteration 4**
-
👍 

**Iteration 5**
-
👍 

**Iteration 6**
-
Where is iteration 6?
Hint : You need to sort the array by movie title, then only keep the title for each movie (map()) and keep only the first 20 ones (splice())

tatimatz:
Thanks Fanny!

PR #1631, opened 2020-09-17
rem2009 katie stamey]

kmstamey:


src/movies.js
```diff
@@ -1,16 +1,99 @@
...
 
+function howManyMovies(movies){
+
+    let occurance = movies.filter(function(movie) {
+        if (movie.director != "Steven Spielberg"){
+            return false;
+        }
+        if (movie.genre.indexOf("Drama")== -1 ) {
+            return false;
+        }
```
IH-WebDev-TA-Remote, line 22: Because both conditions will return false, you can put them in the same conditional using an or operator ||
Also, using indexOf works good here, but this method is mostly used when you are interested in knowing the index in case it exists. Here you only care if the string is there or not. So using the method .includes would work better as this method automatically returns a boolean :)

src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+
+        return true;
+    });
+
+    return occurance.length;
+
+}
+
+
+howManyMovies(movies);
```
IH-WebDev-TA-Remote, line 32: no need to invoke the functions while working with Jasmine here

src/movies.js
```diff
@@ -1,16 +1,99 @@
...
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movies){
+    if (movies.length == 0) {
```
IH-WebDev-TA-Remote, line 36: remember to use triple equal instead of double equal unless you specifically want to convert data types. Best practices!

src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+       if(!curr.rate) {
+           return acc;
+       }
+
+        return acc + curr.rate;
+
+    }, 0);
+
+    let avg = total/movies.length;
+    return avg;
```
IH-WebDev-TA-Remote, line 50: the expected result was only with two decimals, you can get this with a mix of toFixed(2) and changing it back to a number!

src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+
+    let avg = total/occurances.length;
+    return avg;
+}
+
+// Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies) {
+    let sorted = movies.sort(function(a, b){
+        return a.year > b.year;
+    });
```
IH-WebDev-TA-Remote, line 78: Nice! already working with a reduced syntax of the sort method :)

src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+    let total = occurances.reduce(function(acc, curr){
+            return acc + curr.rate;
+    }, 0);
+
+    let avg = total/occurances.length;
+    return avg;
+}
+
+// Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies) {
```
IH-WebDev-TA-Remote, line 75: when you use the sort method, you are mutating the original array. Even it is being passed through a parameter. So everytime you use the sort method or other methods that mutate, it is important to use a copy of it. Via deep clone or sometimes shallow clone.

src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+    let sorted = movies.sort(function(a, b){
+        return a.year > b.year;
```
IH-WebDev-TA-Remote, line 77: the sort order is inverted I believe, so changing the > for a < would solve that

src/movies.js
```diff
@@ -1,16 +1,99 @@
...
+    });
+    return sorted;
+}
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies) {
```
IH-WebDev-TA-Remote, line 82: Same as above, making a copy of the array is very important here!

IH-WebDev-TA-Remote:
Very good job! I see you managed to do most tests of all mandatory iterations! Code is well indented and easy to read. Main feedback is remember when you need to make a copy of an array when using methods that mutate the original data. Keep it up!

PR #1632, opened 2020-09-17
[Rem-2009] Laura Salakari

LauraSalakari:
Yay JavaScript :)

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,212 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(movies) {
+    let allDir = movies.map((elem) => {
+        return elem.director;
+    })
+
+    return allDir;
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+function uniqueDirectors(movies) {
+    let dirs = getAllDirectors(movies);
+    let unique = [];
+
+    dirs.forEach((dir) => {
+        if (unique.indexOf(dir) == -1) {
+            unique.push(dir)
+        }
+    })
+
+    return unique;
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(arr) {
+    let spielberg = arr.filter((elem) => {
+        return (elem.director == "Steven Spielberg" && elem.genre.indexOf("Drama") != -1);
+    })
+
+    return spielberg.length;
+}
+
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(arr) {
+    if (arr != '') {
+        let ratings = arr.reduce((acc, elem) => {
+            if (elem.rate != undefined) {
+                return acc + elem.rate;
+            }
+            else {
+                return acc;
+            }
+
+        }, 0)
+
+        return Math.round((ratings / arr.length) * 100) / 100;
+    }
+    else {
+        return 0;
+    }
+
+}
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
+
+function dramaMoviesRate(arr) {
+    if (arr != '') {
+        let movies = 0;
+
+        let ratings = arr.reduce((acc, elem) => {
+            if (elem.genre.includes("Drama")) {
+                movies++;
+                return acc + elem.rate
+            }
+            else {
+                return acc;
+            }
+        }, 0)
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,212 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(movies) {
+    let allDir = movies.map((elem) => {
+        return elem.director;
+    })
```
IH-WebDev-TA-Remote, line 5: Awesome! using arrow functions like a boss. You can even reduce the syntax more and make this solution in one line! ;)

IH-WebDev-TA-Remote: `return movies.map(elem => elem.director)`

src/movies.js
```diff
@@ -1,16 +1,212 @@
...
+    return allDir;
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+function uniqueDirectors(movies) {
+    let dirs = getAllDirectors(movies);
+    let unique = [];
+
+    dirs.forEach((dir) => {
+        if (unique.indexOf(dir) == -1) {
```
IH-WebDev-TA-Remote, line 16: indexOf works perfect here but since you don't the index if it exists, then using the method .includes would work best! that one already returns a boolean if the string exists in the array or not.

src/movies.js
```diff
@@ -1,16 +1,212 @@
...
+    return unique;
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(arr) {
+    let spielberg = arr.filter((elem) => {
+        return (elem.director == "Steven Spielberg" && elem.genre.indexOf("Drama") != -1);
+    })
+
+    return spielberg.length;
```
IH-WebDev-TA-Remote, line 30: great solution! very few lines of code, awesome :)

src/movies.js
```diff
@@ -1,16 +1,212 @@
...
+}
+
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(arr) {
+    if (arr != '') {
```
IH-WebDev-TA-Remote, line 37: although this condition works it might not be easy to read, I would suggest using something more readable when you try to see if an array is empty. like (arr.length === 0) or (!arr.length)

src/movies.js
```diff
@@ -1,16 +1,212 @@
...
+            if (elem.rate != undefined) {
+                return acc + elem.rate;
+            }
+            else {
+                return acc;
+            }
+
+        }, 0)
+
+        return Math.round((ratings / arr.length) * 100) / 100;
```
IH-WebDev-TA-Remote, line 48: Good approach for getting two decimals, nice!

src/movies.js
```diff
@@ -1,16 +1,212 @@
...
+            if (elem.genre.includes("Drama")) {
+                movies++;
+                return acc + elem.rate
+            }
+            else {
+                return acc;
+            }
+        }, 0)
+
+        if (movies != 0) {
```
IH-WebDev-TA-Remote, line 72: as best practices remember to use "===" or "!==" when comparing values, unless you are specifically trying to convert data types.

src/movies.js
```diff
@@ -1,16 +1,212 @@
...
+        else {
+            return 0;
+        }
+    }
+}
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(arr) {
+    if (arr != '') {
```
IH-WebDev-TA-Remote, line 84: this condition is not fully implemented, it will return undefined if it is not met. no else at the end.

src/movies.js
```diff
@@ -1,16 +1,212 @@
...
+                    }
+                }
+            } 
+        })
+
+        collection.forEach((elem) => {
+            elem.push(elem[1] / elem[2]);
+        })
+
+        collection.sort((a,b)=>{
```
IH-WebDev-TA-Remote, line 186: very clever implementing a sort and throwing the first value, very nice!

IH-WebDev-TA-Remote:
Yay JavaScript indeed! Amazing solution, all tests passing, code is clean and easy to understand!

I see you are already understanding everything well so my suggestion would be to start using reduced versions of methods and try to chain them together to reduce the syntax a little bit. Also to work around reusing function. For example iteration 4 was essentially reusing the function in iteration 3 only with a tweak. Take a look at the solution we will send later for some different approaches.

But again, very impressive job. Really well done! :)

PR #1633, opened 2020-09-17
[REM-2009] Sandra Mendez

scmendez:
Iteration` 4 is driving me slightly up the wall
if I leave the code as is in the file, jasmine's anger is +1
but if I try and address the red test with the code below, jasmine's anger is +4
tried debugging it with the remaining classmates on Zoom and it was a no-go

`function dramaMoviesRate(array) {
    if (!array.includes("Drama")) {
        return 0
    } else {
        let allDramaRates = array.reduce((acc, movie) => {
                return acc + movie.rate;
            } else {
                return acc + 0
            }
        }, 0);

        let allDramaMovies = array.reduce((acc, movie) => {
                return acc + 1;
            } else {
                return acc + 0
            }
        }, 0);

        let dramaAvg = allDramaRates / allDramaMovies
        dramaAvg = dramaAvg.toFixed(2);
        return parseFloat(dramaAvg)
    }
}`

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,117 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(array) {
+    let directorsArr = array.map((movie) => {
+        return movie.director
+    })
+    return directorsArr
+}
+console.log(getAllDirectors(movies));
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(array) {
+    let dramaSpielberg = array.filter((movie) => {
+        return (movie.director === "Steven Spielberg") && (movie.genre.includes("Drama"))
+    })
+    return dramaSpielberg.length
+}
+console.log(howManyMovies(movies));
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(array) {
+    if (!array.length) {
+        return 0
+    } else {
+        let allRates = array.reduce((acc, movie) => {
+                if (movie.rate) {
+                    return acc + movie.rate;
+                } else {
+                    return acc + 0
+                }
+            },
+            0);
+        let rateAvg = (allRates / array.length).toFixed(2);
+        rateAvg = Number(rateAvg);
+        return rateAvg
+    }
+}
+console.log(ratesAverage(movies));
+
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(array) {
+    let allDramaRates = array.reduce((acc, movie) => {
+        if (movie.genre.includes("Drama")) {
+            return acc + movie.rate;
+        } else {
+            return acc + 0
+        }
+    }, 0);
+
+    let allDramaMovies = array.reduce((acc, movie) => {
+        if (movie.genre.includes("Drama")) {
+            return acc + 1;
+        } else {
+            return acc + 0
+        }
+    }, 0);
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,118 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(array) {
+    let directorsArr = array.map((movie) => {
+        return movie.director
+    })
```
IH-WebDev-TA-Remote, line 6: Awesome, using arrow functions! :D

src/movies.js
```diff
@@ -1,16 +1,118 @@
...
+}
+console.log(getAllDirectors(movies));
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(array) {
+    let dramaSpielberg = array.filter((movie) => {
+        return (movie.director === "Steven Spielberg") && (movie.genre.includes("Drama"))
+    })
+    return dramaSpielberg.length
```
IH-WebDev-TA-Remote, line 17: perfect solution, both conditions in the same line and using .includes!

src/movies.js
```diff
@@ -1,16 +1,118 @@
...
+}
+console.log(howManyMovies(movies));
```
IH-WebDev-TA-Remote, line 19: no need to leave this console.logs or function invocations here unless you were checking the console of course.

src/movies.js
```diff
@@ -1,16 +1,118 @@
...
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(array) {
+    if (!array.length) {
+        return 0
+    } else {
+        let allRates = array.reduce((acc, movie) => {
+                if (movie.rate) {
```
IH-WebDev-TA-Remote, line 27: check indentation here, there seems to be different indentation levels in this function.

src/movies.js
```diff
@@ -1,16 +1,118 @@
...
+        let rateAvg = (allRates / array.length).toFixed(2);
+        rateAvg = Number(rateAvg);
+        return rateAvg
+    }
+}
+console.log(ratesAverage(movies));
+
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(array) {
```
IH-WebDev-TA-Remote, line 43: the test not passing here would be solved with the same initial code of iteration 3. 
`if (!array.length) {...`
This happends because along the way the result will become NaN instead of 0 for empty arrays. Using above condition will fix that.

src/movies.js
```diff
@@ -1,16 +1,118 @@
...
+        }
+    }, 0);
+
+    let allDramaMovies = array.reduce((acc, movie) => {
+        if (movie.genre.includes("Drama")) {
+            return acc + 1;
+        } else {
+            return acc + 0
+        }
+    }, 0);
```
IH-WebDev-TA-Remote, line 58: running another reduce to get the amount of movies works fine, however it is not best performance wise because you are running two full loops. Better approach would be to use the above reduce for this purpose. Maybe create a counter before the reduce and increase the value inside the reduce.

IH-WebDev-TA-Remote:
Very good job! All mandatory done except a single test, code is easy to read and understand. Give those bonuses a try when you have a chance, they are really hard but will definitely challenge you! Also, check how indentation should be, I see you were using different levels throughout the code. Sometimes one tab, sometimes two tabs, sometimes more.

As per your question: !array.includes("Drama") won't work because the .includes method doesn't work like a full search that will look through all the atributes of each object. The way it works is it will try to compare the string "Drama" against each element of the array (in this case, each object). And those will always be false. Inside the reduce it works because it is comparing the string "Drama" against each element of the genre array (and those are strings).

But still very minor things, overall very well done, keep it up!

PR #1634, opened 2020-09-17
[REM-2009]zuzana

ZuzanaAK:


src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+    let directorsArr = array.map((movie) => {
+        return movie.director
+    })
+    return directorsArr
+  }
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(array) {
+
+let filterArray=array.filter ((movie)=>{
```
IH-WebDev-TA-Remote, line 17: watch indentation, remember elements inside a function, conditionals and loops should be one tab to the side. This helps a lot with readability of code.

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+    return movie.director === "Steven Spielberg" && movie.genre.includes("Drama")
```
IH-WebDev-TA-Remote, line 18: very good job putting both conditions in the same line and even using .includes!

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(movies) {  
+
+    if (movies.length === 0) {
+        return 0
+    }
+
+    let total = movies.reduce ((acc,elem) => {  
+        return acc+elem.rate
```
IH-WebDev-TA-Remote, line 31: the test that is not passing could be handled here. Sometimes a movie may not have a rate, so trying to add that value (undefined) to the acc will result on an issue. Creating a condition here would solve it!

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+
+     return Number(avgRate.toFixed(2))
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+    
+    function dramaMoviesRate(array){
+
+        if(array.length===0 && array.movie.genre !="Drama")
```
IH-WebDev-TA-Remote, line 44: array is an array, so you cannot access array.movie.genre directly. This condition will always return true, because it will always be undefined and thus, not "Drama"

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+
+        {return 0};
+
+        let totalRate=array.reduce((acc,elem)=>{
+            return acc+elem.rate
+        },0)
```
IH-WebDev-TA-Remote, line 50: this reduce is adding the elements of all the movies, not the "Drama" ones, that condition could be included here since elem would be each movie. So you can always access elem.genre

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+        }
+      })
+      return secondArr
+} 
+
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically (movies) {
+
+    let secondArr = JSON.parse(JSON.stringify(arr))
```
IH-WebDev-TA-Remote, line 86: since you are using .map to create a new array below, there is not need to use this deep copy. In fact you are creating this, only to overwrite the value in the below code :)

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+
+        secondArr = movies.map ((movies) => {
+        return movies.title;
+    })
+
+    secondArr.sort ((a,b) =>{
+        if (a.secondArr > b.secondArr) {
```
IH-WebDev-TA-Remote, line 93: you are trying to sort by the title, so the attributes to compare should be a.title and b.title.

IH-WebDev-TA-Remote:
Very good job, there were some small things breaking the code but nothing too big. The approach and code logic was very good. Try to work a little bit on improving indentation. And set the default VS code spacing to be 2 instead of 4 as this is more common. To do that select where it says Spaces: 4 at the bottom of the vs code and change the default to be 2.
Still, overall, good job! keep working hard :)

PR #1635, opened 2020-09-17
[REM2009] lovro hodak

LovroHodak:
…th this

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,94 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(array){
+    let directorsArr = array.map((movie)=>{
+    return movie.director
+    })
+    return directorsArr
+    }
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(array){
+    
+    let filterArray = array.filter((elem)=>{
+        return (elem.director==="Steven Spielberg" && elem.genre.indexOf("Drama") != -1)
+      })
+      return filterArray.length
+}
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
-
+function ratesAverage(array){
+    if(array != ''){
+    let total = array.reduce((acc, elem)=>{
+        if(elem.rate){
+            return acc + elem.rate / array.length
+        } else {
+            return acc
+        }
+    }, 0)
+    return total
+}
+}
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(array){
+    if (array != ''){
+        let movies =0;
 
+        let ratings = array.reduce((acc, elem)=>{
+            if(elem.genre.includes("Drama")){
+                movies++;
+                return acc + elem.rate
+            } else{
+                return acc
+            }
+        }, 0)
+    }
+    
+}
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,79 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(array){
+    let directorsArr = array.map((elem)=>{
+    return elem.director
```
IH-WebDev-TA-Remote, line 4: check indentation, remember to add elements inside a function one tab to the right

src/movies.js
```diff
@@ -1,16 +1,79 @@
...
+    return directorsArr
+    }
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(array){
+    
+    let filterArray = array.filter((elem)=>{
+        return (elem.director==="Steven Spielberg" && elem.genre.indexOf("Drama") != -1)
```
IH-WebDev-TA-Remote, line 14: Good job on putting the two conditions on the same line. indexOf works good here, however this method is mostly used when you actually want to know the index in case it exists. To know only if an element exists or not inside an array you can also use .includes which returns a boolean.

src/movies.js
```diff
@@ -1,16 +1,79 @@
...
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
-
+function ratesAverage(array){
+    if(array.length === 0){
+        return 0
+    }
+    let average = array.reduce((acc, elem)=>{
+      return acc + elem.rate
+    }, 0)
+    return average / array.length
```
IH-WebDev-TA-Remote, line 26: to pass the test that ask for the number to have only to decimals you can use the method toFixed(2) and then convert the result to a number.

src/movies.js
```diff
@@ -1,16 +1,79 @@
...
-
+function dramaMoviesRate(array){
+    let dramaMovies = array.reduce((acc, elem)=>{
+      if(elem.genre.includes("Drama")){
+        return acc + elem.rate
+      } else{
+        return acc
+      }
+    }, 0)
+    return dramaMovies / dramaMovies.length
```
IH-WebDev-TA-Remote, line 37: you cannot do dramaMovies.length, as dramaMovies is the sum of rates, so, it is a number, not the array. To get this number you will need to add a new variable before the reduce, and inside the condition, increase the value everytime a new rate is added.

src/movies.js
```diff
@@ -1,16 +1,79 @@
...
+      } else if(a.title < b.title){
+        return 1
+      }else {
+        return 0
+      }
+    })
+    if(array.length > 20){
+      array.splice(20)
+    }
+    return oredered
```
IH-WebDev-TA-Remote, line 75: the approach is good, however you only needed to return an array only wit the titles. You can do this with a map.

src/movies.js
```diff
@@ -1,16 +1,79 @@
...
+                  else {
+                      return 0;
+                  }
+              }
+          })
+    return yearSort
+  }
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-
+function orderAlphabetically(array){
```
IH-WebDev-TA-Remote, line 62: again, not to mutate the array you need to deep clone it, or alternately do the map to get only the titles before you apply the sort.

IH-WebDev-TA-Remote:
Good job! the code logic was very good but there were some small issues preventing all tests from passing. Give those bonuses a try during the weekend if you have a chance!
Also, make sure you watch indentation and spacing a little bit more and work on having an even cleaner code. Check the solution so see the differences from your code and that one.
Still, overall very well done. Keep working hard!

PR #1636, opened 2020-09-17
[REM - 2009] Vykintas Petrikas

VykiPetr:
Did all except for the last bonus, i cant wrap my head around on which methods to use. I have an idea what is to be done but i do not understand how to do it. I want to creat an array holding the year with the rate, then take each year with its rate and make all of them into a single element { year: 1999, rate: 8.4},...] but how... sorting afterwards would seem doable.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,179 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+
+function getAllDirectors(movies) {
+    let dirArr = movies.map((i) => {
+        return i.director
+    })
+    return dirArr
+}
+
+function cleanedArray(dirArr) {
+    let cleanArr
+
+}
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies) {
+    let steveDrama = movies.filter((i) => {
+        if (i.director === "Steven Spielberg" && i.genre.includes("Drama")) {
+            return i
+        }
+    }).length
+    return steveDrama
+}
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+
+function ratesAverage(movies) {
+    if (movies.length === 0) {
+        return 0
+    }
+    let numOfMovies = 0
+    let avgTot = movies.reduce((tot, e) => {
+        if (e.rate == undefined) {
+            numOfMovies += 1
+            return tot
+        } else {
+            numOfMovies += 1
+            return tot + e.rate
+        }
+    }, 0)
+
+    let avg = avgTot / numOfMovies
+    avg = avg.toFixed(2)
+    avg = Number(avg)
+    return avg
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(movies) {
+    if (movies.length === 0) {
+        return 0
+    }
+    let numOfMovies = 0
+    let avgTot = movies.reduce((tot, e) => {
+        if (e.genre.includes("Drama")) {
+            numOfMovies += 1
+            return tot + e.rate
+        } else {
+            return tot
+        }
+    }, 0)
+    if (numOfMovies === 0) {
+        return 0
+    }
+    let avg = avgTot / numOfMovies
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,179 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+
+function getAllDirectors(movies) {
+    let dirArr = movies.map((i) => {
```
IH-WebDev-TA-Remote, line 6: the variable i is most used in loops, however, here, instead of i try to work around better names for the parameters of the methods, it will help a lot on bigger functions to understand what is going on. In this case something like movie or even eachMovie would be better.

src/movies.js
```diff
@@ -1,16 +1,179 @@
...
+}
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies) {
+    let steveDrama = movies.filter((i) => {
+        if (i.director === "Steven Spielberg" && i.genre.includes("Drama")) {
+            return i
+        }
+    }).length
+    return steveDrama
```
IH-WebDev-TA-Remote, line 24: good use of .includes, putting both conditions in the same line and even chaining the .length to the filter.
By returning the result of the filter instead of saving that inside a variable you could even make the code shorter ;)

src/movies.js
```diff
@@ -1,16 +1,179 @@
...
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(movies) {
+    if (movies.length === 0) {
+        return 0
+    }
+    let numOfMovies = 0
+    let avgTot = movies.reduce((tot, e) => {
+        if (e.rate == undefined) {
+            numOfMovies += 1
```
IH-WebDev-TA-Remote, line 35: if you are adding this line for every condition you could even do it only once outside of the conditional.

src/movies.js
```diff
@@ -1,16 +1,179 @@
...
+        } else {
+            numOfMovies += 1
+            return tot + e.rate
+        }
+    }, 0)
+
+    let avg = avgTot / numOfMovies
+    avg = avg.toFixed(2)
+    avg = Number(avg)
+    return avg
```
IH-WebDev-TA-Remote, line 46: the last 4 lines could be done in a single line, you are a pro, you can do this! :D

src/movies.js
```diff
@@ -1,16 +1,179 @@
...
+            let minutes = i.duration.slice(0, 3)
+            return i.duration = minutes
+        } else if (i.duration.includes('min')) {
+            let minutes = Number(i.duration.slice(0, 2))
+            return i.duration = minutes
+        } else {
+            return 0
+        }
+    })
+    return copyMovies
```
IH-WebDev-TA-Remote, line 148: Very good solution! There might be a way to reduce a little bit the amount of conditions (for example merge 1 and 2, 4 and 5) However, still very impressive. Keep it up! :)

src/movies.js
```diff
@@ -1,16 +1,179 @@
...
+    }
+    let bestYear = ''
+    let bestRate = 0
+    let yearArr = movies.map((i) => {
+        return i.year   
+    })
+    let rateArr = movies.map((i) =>{
+        return i.rate
+    })
+
```
IH-WebDev-TA-Remote, line 168: having two different arrays might not be the best approach because the you won't be able to match the data (year-rate), that is why the most common approach here is working around objects that hold both the year and the rate. That way using some methods you can extract the required data.

src/movies.js
```diff
@@ -1,16 +1,179 @@
...
+        } else {
+            if (a.title > b.title) {
+                return 1
+            } else if (a.title < b.title) {
+                return -1
+            } else {
+
+                return 0
+            }
+            return 0
```
IH-WebDev-TA-Remote, line 97: this return will never run because the else above is handling the default condition so you can remove it, probably in vs code you can even see it is greyed out for this same reason.

IH-WebDev-TA-Remote:
Very well done, all mandatory tests done and even one bonus iteration! Kudos! also very clean code, easy to understand and the logic is spot on. Best suggestion is work on better naming for your parameters, like I mentioned before, you will thank you used some descriptive names on really long functions.
Keep up the amazing work!

PR #1637, opened 2020-09-17
[REM-2009] Sumiya Ushiro

sumi0820:
I got stuck at iteration 2 and 3, but understood thanks to others' support. If I read a valid code, I understand what is happening and why it works, but the biggest challenge I have now is how to come up with the valid and readable logic as quick as I can. 

As for BONUS - Iteration 7, it became a redundant code and Jasmine also did not pass followings.
- Should return a new array, not update the original one
- The duration of the movie should be a number
I guessed something was wrong with my code in terms of new array creation, and tried to use a JSON trick, but it did not work.
Hence, I suspect the structure probably is wrong...

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,150 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(array) {
+  let directors = array.map((movie) => {
+    return movie.director;
+  });
+  return directors;
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
-// Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(array) {
+  let spielbergMovie = array.filter((movie) => {
+    return (
+      movie.director === "Steven Spielberg" &&
+      movie.genre.indexOf("Drama") != -1
+    );
+  });
+  return spielbergMovie.length;
+}
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(array) {
+  if (array.length !== 0) {
+    let total = array.reduce((acc, movie) => {
+      if (movie.rate != "" && movie.rate != undefined) {
+        return acc + movie.rate;
+      } else {
+        return acc;
+      }
+    }, 0);
+    total = total / array.length;
+    return Number(total.toFixed(2));
+  } else {
+    return 0;
+  }
+}
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
-// Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function dramaMoviesRate(array) {
+  let dramaMovies = array.filter((movie) => {
+    return (
+      movie.rate != "" &&
+      movie.rate != undefined &&
+      movie.genre.indexOf("Drama") != -1
+    );
+  });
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,150 @@
...
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
-// Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(array) {
+  let spielbergMovie = array.filter((movie) => {
+    return (
+      movie.director === "Steven Spielberg" &&
+      movie.genre.indexOf("Drama") != -1
```
IH-WebDev-TA-Remote, line 15: indexOf works fine here, but this method is mostly used when you want to know the index in case it exists. If you only need to know if the element is included, then the method .includes works best as it already returns a boolean.

src/movies.js
```diff
@@ -1,16 +1,150 @@
...
+  });
+  return spielbergMovie.length;
+}
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(array) {
+  if (array.length !== 0) {
+    let total = array.reduce((acc, movie) => {
+      if (movie.rate != "" && movie.rate != undefined) {
```
IH-WebDev-TA-Remote, line 26: you could reduce this conditionals to a single one (movie.rate) this will essentially check if movie.rate is not equal to ANY falsy value (0, "", undefined, null, NaN). and this way is commonly read as... "if movie.rate exists"
`if (movie.rate) {...}`

src/movies.js
```diff
@@ -1,16 +1,150 @@
...
+    return (
+      movie.rate != "" &&
+      movie.rate != undefined &&
+      movie.genre.indexOf("Drama") != -1
+    );
+  });
+
+  if (dramaMovies.length !== 0) {
+    let total = dramaMovies.reduce((acc, movie) => {
+      if (dramaMovies.length > 0) {
```
IH-WebDev-TA-Remote, line 52: this condition is already being checked in line 50 so there is no need to run it again.

src/movies.js
```diff
@@ -1,16 +1,150 @@
...
+
+// // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
+function turnHoursToMinutes(array) {
+  let newMovies = array.map((movie) => {
+    if (
+      movie.duration.indexOf("h") !== -1 &&
+      movie.duration.indexOf("min") === -1
+    ) {
+      let hour = Number(movie.duration[0] * 60);
+      movie.duration = hour;
```
IH-WebDev-TA-Remote, line 123: even though each movie is being passed through the parameter of the map, this still points to the original reference, so doing movie.duration will modify the original data. Doing a deep clone either of the array at the start of the function or of each movie obj at the start of the .map will solve this.

IH-WebDev-TA-Remote:
It takes a lot of practice to come up with this approaches quickly to solve this kind of situations, later during module 1 we will start doing katas, which are specially target at this. So don't worry right now, even if it takes you some time and asking for help, solving these exercises and understanding everything that is happening in your code says that you are in the perfect spot of the learning process :)

As per your question, try adding this line at the start of the function turhHoursToMinutes, and change the array that is being mapped to newArr. It will solve the problem. This is caused because inside the code the original value is being mutated. check the line by lobe feedback to know where.
`let newArr = JSON.parse(JSON.stringify(array))`

Still, overall, really good solutions! code is clean, well indented and the logic is very easy to follow. You even did the first bonus which is quite hard. Keep up the amazing work!

PR #1638, opened 2020-09-17
[REM 2009] Ross Homaidan

ibirossi:
Felt a lot better with this Lab.  Still not understanding why things we do in class, or in previous iterations dont work in other iterations. For example sort in iteration 5 doesn't translate to iteration 6.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,92 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(array){
+    let dirArray=array.map((movie)=>{
+        return movie.director;
+    })
+    return dirArray
+}
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(array){
+    let filterArray=array.filter((movie)=>{
+        return  movie.director==="Steven Spielberg" && movie.genre.includes("Drama")
+
+    })
+    return filterArray.length
+}
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(array){
+    if(array.length===0)
+    {return 0};
+    let totRate=array.reduce((acc,elem)=>{
+        return acc+elem.rate
+    },0)
+
+    let avgRate=totRate/array.length
+    return Number(avgRate.toFixed(2))
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(array){
+    if(array.length===0 && array.genre.length >1 && array.movie.genre !="Drama")
+    {return 0};
+    
+    let totRate=array.reduce((acc,elem)=>{
+        return acc+elem.rate
+    },0)
+
+    let avgRate=totRate/array.length
+    return Number(avgRate.toFixed(2))
+}
+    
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,92 @@
...
+}
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(array){
+    let filterArray=array.filter((movie)=>{
+        return  movie.director==="Steven Spielberg" && movie.genre.includes("Drama")
+
+    })
+    return filterArray.length
```
IH-WebDev-TA-Remote, line 16: really good solution! very short, nice!

src/movies.js
```diff
@@ -1,16 +1,92 @@
...
+}
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(array){
+    if(array.length===0)
+    {return 0};
```
IH-WebDev-TA-Remote, line 22: check the correct syntax for conditionals in multiple lines, the { should start in the if line and the } in a separate line, this is common for better readability.

` if(array.length===0) {
    return 0}
;`

The one below is also accepted and commonly used If you want to write the conditional in a single line (and the code only has one line). Just remember to remove the curly brackets.

`if(array.length===0) return 0;`

src/movies.js
```diff
@@ -1,16 +1,92 @@
...
+    },0)
+
+    let avgRate=totRate/array.length
+    return Number(avgRate.toFixed(2))
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(array){
+    if(array.length===0 && array.genre.length >1 && array.movie.genre !="Drama")
+    {return 0};
```
IH-WebDev-TA-Remote, line 34: Same as above, check correct way of writing the conditionals, also, remember to use always "===" and "!==" unless you specifically want to convert data types during the comparison, which is hardly ever used.

src/movies.js
```diff
@@ -1,16 +1,92 @@
...
+    })
+    return filterArray.length
+}
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(array){
+    if(array.length===0)
+    {return 0};
+    let totRate=array.reduce((acc,elem)=>{
+        return acc+elem.rate
```
IH-WebDev-TA-Remote, line 24: the condition to avoid the movies which do not have a rate to pass this missing test could be applied here. A simple conditional that checks if (elem.rate). Otherwise the value undefined would try to be added to the accumulator and cause the value to end up being NaN.

src/movies.js
```diff
@@ -1,16 +1,92 @@
...
+        if (a.title>b.title){
+            return 1
+        }
+    
+        else if (a.title<b.title){
+            return -1
+        }
+            else return 0
+        }) 
+            return sortNewArray
```
IH-WebDev-TA-Remote, line 87: good job this far! but also needed to get only the titles and only the first 20 elements. You can achieve this working around a .map and a .slice on the sortNewArray :)

IH-WebDev-TA-Remote:
Good job on the lab, the logic is very spot on and the code easy to read. Most of the tests were passing, yet, I would recommend you give it another try and complete the missing tests and even try the bonuses, even following along some of the solutions sent so you understand everything that is happening there.

The sort in iteration 6 is a little bit different because there are other things needed to be done, check the instructions again and the feedback notes. Give that one a try as well and see if you complete it!

Keep up the good work!

PR #1639, opened 2020-09-17
[REM 2009] Aleksandra Tanasijenko

alekstana:


src/movies.js
```diff
@@ -1,15 +1,107 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(array){
+    let directors = array.map((movies) => {
+    return movies.director
+    })
+ return directors
+}
+getAllDirectors(movies)
```
IH-WebDev-TA-Remote, line 8: for these labs, Jasmine is already running all the functions with the tests, so there is no need to invoke the functions yourself.

src/movies.js
```diff
@@ -1,15 +1,107 @@
...
+
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(array) {
+    let dramaMoviesSpielberg = array.filter((movies) =>{
+         return (movies.director === "Steven Spielberg") && (movies.genre.includes("Drama"))       
+    })
+    return dramaMoviesSpielberg.length
+}
```
IH-WebDev-TA-Remote, line 20: Good solution, very short , both conditions in the same line and using .includes, awesome!

src/movies.js
```diff
@@ -1,15 +1,107 @@
...
+howManyMovies(movies)
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(array) {
+    let allRates = array.reduce((acc, elem) => {
+        return acc + elem.rate
+      },0)
+                                
+      let avgRates = allRates/array.length
+      return avgRates.toFixed(2)
```
IH-WebDev-TA-Remote, line 31: the .toFixed() is kinda of an odd method because it takes a number to reduce the amount of decimals, however it will return the number as a string. So you will need to manually change the string back to a number, you can do this in different ways, below there are 3 common ways.
````
Number(avgRates)

parseFloat(avgRates)
+avgRates
````

File: src/movies.js [hunk 4]
```diff
@@ -1,15 +1,107 @@
...
+        if (movies.genre.includes("Drama")){
+        return acc + 1
+    } else {return acc+0}
+      },0)
+
+      let avgDramaRates = allAvgDramaRates/numberOfDramas
+      return vgDramaRates.toFixed(2)
+                                
+}
+
```
IH-WebDev-TA-Remote, line 54: This is a really good approach, however there are some syntax errors causing your code to break, remember to ask more questions in the channel or to us when you think everything should be running correctly and it is not :)

src/movies.js
```diff
@@ -1,15 +1,107 @@
...
+      let avgRates = allRates/array.length
+      return avgRates.toFixed(2)
+}
+console.log(ratesAverage)
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(array) {
+    let allAvgDramaRates = array.reduce((acc, elem) => {
+        if (movies.genre.includes("Drama")){
```
IH-WebDev-TA-Remote, line 39: the two parameters of the .reduce are acc and elem. acc is the accumulator and elem is the movie object, however inside the conditional you are trying to target the genre of a variable called movie, not elem. Change movie for elem in both reduce and they should work property.

src/movies.js
```diff
@@ -1,15 +1,107 @@
...
+      },0)
+
+    let numberOfDramas = array.reduce((acc, elem) => {
+        if (movies.genre.includes("Drama")){
+        return acc + 1
+    } else {return acc+0}
+      },0)
+
+      let avgDramaRates = allAvgDramaRates/numberOfDramas
+      return vgDramaRates.toFixed(2)
```
IH-WebDev-TA-Remote, line 51: typo, the variable is avgDramaRates, not vgDramaRates. Finally, same as above iteration, you will need to convert the string back into a number.

src/movies.js
```diff
@@ -1,15 +1,107 @@
...
+                                
+}
+
+dramaMoviesRate(movies)
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(array) {
+    let newMovies = JSON.parse(JSON.stringify(movies))
```
IH-WebDev-TA-Remote, line 59: remember you are working around the parameter you are getting in the function, which is array, not movies.
`let newMovies = JSON.parse(JSON.stringify(movies))array`
Again, the approach is very good and just that little issue was causing the whole code to break :)

src/movies.js
```diff
@@ -1,15 +1,107 @@
...
+    let sortedArray= newMovies.sort((a, b) => { 
+        if (a.year > b.year) {
+            return 1;
+          }
+         else if (a.year < b.year) {
+           return -1;
+         }
+         else {
+          if (a.title > b.title) {
```
IH-WebDev-TA-Remote, line 68: switch < > symbols for the .title segment to get the in the correct order.

src/movies.js
```diff
@@ -1,15 +1,107 @@
...
+         else if (a.title < b.title) {
+           return -1;
+         }
+         else {
+           return 0
+         }
+         }
+         )
+         return sortedAlphabetically.splice(0,20)
+}
```
IH-WebDev-TA-Remote, line 101: good approach here! you were just missing that instead of returning the whole object you only needed to return the list of titles, this is a similar approach of iteration 1 :)

IH-WebDev-TA-Remote:
Overall good approach for all of the iterations, also code is clean and easy to understand, just check a little bit of the indentations, sometimes there were nested elements visually at the same of the ones outside.

Also, remember to ask more for help when you believe your code should be working and it is not, you can see that some of the issues you had were simple and very common issues but the whole logic was already there.

Keep up the great work, you can do this! :D

PR #1640, opened 2020-09-17
[REM-2009] Santiago Andrés Ardiles Páez

SantAndresP:
Couldn't find a way to solve the last requirement of the second iteration.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,6 +1,6 @@
 /*  Iteration 1: All directors? - Get the array of all directors.  */
 /*  Bonus: It seems some of the directors had directed multiple movies,
-    so they will pop up multiple boths in the array of directors.
+    so they will pop up multiple times in the array of directors.
     How could you "clean" a bit this array and make it unified (without duplicates)? */
 const getAllDirectors = (movies) => {
   const directors = movies.map((movie) => {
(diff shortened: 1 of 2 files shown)


SantAndresP:
I didn't know that it was possible to change the code once it had been submitted.

src/movies.js
```diff
@@ -1,16 +1,117 @@
-// Iteration 1: All directors? - Get the array of all directors.
-// _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+/*  Iteration 1: All directors? - Get the array of all directors.  */
+/*  Bonus: It seems some of the directors had directed multiple movies,
+    so they will pop up multiple minutess in the array of directors.
+    How could you "clean" a bit this array and make it unified (without duplicates)? */
+const getAllDirectors = (movies) => {
+  return movies.map((movie) => movie.director);
```
IH-WebDev-TA-Remote, line 6: simply amazing, one line solution! can't wait to see you solving katas like a pro >:)

IH-WebDev-TA-Remote: if you want to reduce the syntax even more you don't need to use parenthesis in the parameter when you only have one parameter.

src/movies.js
```diff
@@ -1,16 +1,117 @@
...
+};
 
-// Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+/*  Iteration 2: How many drama movies did Steven Spielberg direct?  */
+const howManyMovies = (movies) => {
+  return movies.filter((movie) => {
+    return (
+      movie.director === "Steven Spielberg" && movie.genre.includes("Drama")
+    );
+  }).length;
```
IH-WebDev-TA-Remote, line 15: amazing solution, now try it all in a single line ;)

src/movies.js
```diff
@@ -1,16 +1,117 @@
...
+};
 
-// Iteration 3: All rates average - Get the average of all rates with 2 decimals
+/*  Iteration 3: All rates average - Get the average of all rates with 2 decimals.  */
+const ratesAverage = (movies) => {
+  const avgRating = movies.reduce((acc, movie, _, src) => {
+    return acc + (movie.rate || 0) / src.length;
```
IH-WebDev-TA-Remote, line 21: amazing use of the or operator here. works like a charm.

IH-WebDev-TA-Remote: Also, perfect dividing every item by the length inside the reduce instead of one time the whole sum between the number of movies. some little cool math trick in there.

IH-WebDev-TA-Remote:
Yes! once you submit the pull request any commits you do after will reflect on the same pull request until closed.
I see you kept working on the lab and to be honest the solutions are pretty amazing, very few feedback I could throw here. Code logic is perfect, execution is very good and the code is clean and easy to read. 

I can see why you got a little stuck at the end, probably that approach of creating an array of objects is not the best (it is essentially similar to the original array) instead an object with the years as keys and arrays of rates as values would be better approach. Still, very good you got to that point and maybe even there is a way to work around that.

Still, overall, simply amazing. Keep it up!

PR #1641, opened 2020-09-17
 [REM-2009] Jordane Ruiz

jordaneruiz:


jordaneruiz:
I'll need to deeply review the iterations. I understand the concept but I can't apply it

src/movies.js
```diff
@@ -1,16 +1,170 @@
...
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+const howManyMovies = (array) => {
+    let dramaMovies = array.filter((movie) => {
+        return movie.genre.includes("Drama") && movie.director === "Steven Spielberg";
+        // return movieType.genre.includes("drama");
+    })
+    return dramaMovies.length
```
IH-WebDev-TA-Remote, line 18: good solution! short and using .includes and both conditions in the same line.

src/movies.js
```diff
@@ -1,16 +1,170 @@
...
+}
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+const ratesAverage = (array) => {
+    
+    if (array != "") {
```
IH-WebDev-TA-Remote, line 25: this condition works perfectly fine but it might not be easy to read, specially for a new developer. Since you are trying to check if the array is empty something like (array.length === 0) or (!array.length) would make more sense.

src/movies.js
```diff
@@ -1,16 +1,170 @@
...
+
+        let averageRates = array.reduce((acc, i) => {
+            if (i.rate) {
+            return acc + i.rate / array.length
```
IH-WebDev-TA-Remote, line 29: check indentation level here.

IH-WebDev-TA-Remote: Very smart using the "/array.length" here instead of after the reduce for the whole sum. This is some math magic that allows this but that way you can reduce the code, very clever :)

src/movies.js
```diff
@@ -1,16 +1,170 @@
...
+    }
+
+
+
+
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+const dramaMoviesRate = (array) => {
+    if (array.genre !== "Drama") {
```
IH-WebDev-TA-Remote, line 54: you cannot directly target .genre of array. You can only use dot notation when trying to access attributes of an object. In this case array.genre will be undefined so the condition will always be true.

src/movies.js
```diff
@@ -1,16 +1,170 @@
...
+        return 0;
+    }; 
+
+    let dramaMovies = array.filter((movie) => {
+        return movie.genre.includes("Drama");
+     });
+
+    let dramaAverageRates = array.reduce((acc, i) => {
```
IH-WebDev-TA-Remote, line 62: since you filtered the Drama movies in the above variable dramaMovies then you want to apply reduce to that array (dramaMovies.reduce) instead of the array of full movies.

src/movies.js
```diff
@@ -1,16 +1,170 @@
...
+            return acc + i.rate / dramaMovies.length
+        }, 0)
+
+    return dramaAverageRates
```
IH-WebDev-TA-Remote, line 66: You also needed to return the number with two decimals, so the same approach as last iteration.

src/movies.js
```diff
@@ -1,16 +1,170 @@
...
+                    }
+                    else if (a.title < b.title) {
+                      return -1;
+                    }
+                    else {
+                      return 0
+                    }
+                }
+    });
+    return movieByYear;
```
IH-WebDev-TA-Remote, line 120: Good solution! :)

src/movies.js
```diff
@@ -1,16 +1,170 @@
...
+};
+
+
+
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 
+const orderAlphabetically = (array) => {
+
+    let orderedMovies = JSON.parse(JSON.stringify(array))
+    let orderedTitles = orderedMovies
```
IH-WebDev-TA-Remote, line 130: this line does not make much sense since it is pointing to the same reference,  so using the variable orderedMovies will be exactly the same of using the variable orderedTitles. Since they both point to the same memory location.

src/movies.js
```diff
@@ -1,16 +1,170 @@
...
+        //     return orderedTitles.splice(0, 21)
+        // } else {
+        //     return orderedTitles;
+        // }
+
+    });
+
+    let moviesTitle = orderedTitles.map((elem) => {
+        return elem.title
+   })
```
IH-WebDev-TA-Remote, line 161: Good approach this far! the only thing you were missing was getting the first 20 elements, you can do this with the .slice method for example. So in the end of the function, just returning moviesTitle.slice(0, 20)

IH-WebDev-TA-Remote:
Good job on the lab, there were some final touches preventing the tests from passing but overall good approach to solving all the mandatory ones. Also the code is easy to read and understand. One suggestion would be to change indentation levels from 4 to 2. To do this click on the lower part of VS code where it says spaces: 4, click on indent using spaces and set it to 2. It is most common to use that one.

Overall well done! keep practicing and even try those bonuses when you have a chance :)

PR #1642, opened 2020-09-18
[REM - 2009] Marta Obregon

MartaObregon:


src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+    })
+    return directorsArr
+}
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies (array){
+    let dramaMovies = array.filter ((movie) => {
+        return movie.director ==="Steven Spielberg" && movie.genre.includes("Drama")
+    })
+        return dramaMovies.length
```
IH-WebDev-TA-Remote, line 15: watch indentation here, the return should be at the same level of the let above.

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+}
```
IH-WebDev-TA-Remote, line 16: Good solution here! using .includes and putting the both conditions in the same line. Very good!

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage (movies){
+    if(movies.length===0){
```
IH-WebDev-TA-Remote, line 19: remember to use more spaces for readability. Like this:
`if (movies.length === ) {`

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+        return 0
+
+    }else{
+    let total = movies.reduce ((acc, movies) =>{
+        if ( movies.rate == undefined){
```
IH-WebDev-TA-Remote, line 24: for best practices, remember to use triple equal unless you specifically need to convert data types. Also for cases like this where you want to check if an element exists or not it is common to just put the element and use the ! operator, like (!movies.rate), this can translate as "if movies.rate does not exists". Or (movies.rate) which translates as "if movies.rate exists"

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+    average= average.toFixed(2)
+    average = Number(average)
+    return average
+}
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate (movies) {
+    if (!movies.includes("Drama")){
```
IH-WebDev-TA-Remote, line 39: since movies is an array of objects, then this condition will try to compare the string "Drama" to each object, and so, the condition will always come out as true (!false). So the function will return 0 for every case. It seems you are already checking the "Drama" part later so no need for this par of the code.

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+        return 0
+    }else { 
+        let TotalDramaRates = movies.reduce ( (acc, movies)=>{
+            if (movies.genre.includes("Drama")){
+                return acc + movie.rate;
```
IH-WebDev-TA-Remote, line 44: the parameter name is movies, not movie. Sometimes can be tricky using similar names for variables (movies, movie) so it is common to avoid this to name them a little bit different, you could try moviesArray for the array and eachMovie for when working array each object inside the methods.

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+            if (movies.genre.includes("Drama")){
+                return acc + movies.length;
+            }else{
+                return acc + 0
+            }
+        }, 0)
+    
+    let avgDrama = TotalDramaRates/totalDramaMovies
+    avgDrama = avgDrama.toFixed(2)
+    avgDrama = Number(average)
```
IH-WebDev-TA-Remote, line 59: the variable name is avgDrama, not average.

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+        let TotalDramaRates = movies.reduce ( (acc, movies)=>{
+            if (movies.genre.includes("Drama")){
+                return acc + movie.rate;
+            }else{
+                return acc + 0
+            }
+        }, 0);
+        let totalDramaMovies = movies.reduce ( (acc, movies) =>{
+            if (movies.genre.includes("Drama")){
+                return acc + movies.length;
```
IH-WebDev-TA-Remote, line 51: finally, in this reduce you are trying to obtain the total amount of movies. and this is done simpling adding 1 to the accumulator everytime you find a Drama movie. So: return acc + 1

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+    avgDrama = Number(average)
+    return parseFloat(avgDrama)
+    }
+    }   
+ 
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear (movies){
+    if (movies.includes("year")){
```
IH-WebDev-TA-Remote, line 68: Same as above, you cannot apply .includes here to check for strings, because every element of the array is an objects, and you cannot compare strings and objects.

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+            }else if (a.movies.year < b.movies.year){
+                return -1;
+            }else {
+                return 0
+            }
+        })
+    })
+    let newSort = JSON.parse(JSON.stringify(movies.year));
+    return newSort
+}}
```
IH-WebDev-TA-Remote, line 82: I see you used the correct methods and tools, however the logic approach to solving this was a little bit off. The idea is first to deep copy the array (using JSON.parse and JSON.stringify), then simply apply the sort method to the copy and return the copy. No need to use .includes or .filter :)

src/movies.js
```diff
@@ -1,15 +1,105 @@
...
+                }else if (a.movie.title < b.movie.title){
+                    return -1;
+                }else {
+                    return 0
+                }
+            })
+        })
+        let titleSort = JSON.parse(JSON.stringify(movie.title));
+        return titleSort
+    }}
```
IH-WebDev-TA-Remote, line 100: Same as above, the tools were good, it is just the logical approach that was a little bit off.

IH-WebDev-TA-Remote:
Good effort on the lab, the first iterations were perfect or at least some small issues that where causing the code to break. However I would suggest you try again iterations 5 and 6 from scratch. Try to follow the logic I mentioned in the comments above.

Try as well the pokemon JS exercise Manish sent if you haven't, it is really good practice for these methods. Keep working hard! :)

PR #1643, opened 2020-09-18
[REM-2009] Juan Jose Carreño

JJexcelsior:


src/movies.js
```diff
@@ -1,16 +1,150 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+
+
+function getAllDirectors(movies){
+    return movies.map((movie) => movie.director).filter((value, index, self) => self.indexOf(value) === index);
```
IH-WebDev-TA-Remote, line 7: Very nice, all in one line, however the filter in this case is not trying to solve any Jasmine Test. Removing that part will throw the same result.

src/movies.js
```diff
@@ -1,16 +1,150 @@
...
+}
+
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies){
+    return movies.filter(function(movie){
+        return (movie.director == "Steven Spielberg" && movie.genre.indexOf("Drama") !== -1);
+    }).length;
+}
```
IH-WebDev-TA-Remote, line 16: Very good solution, again, everything is the same return, one suggestion would be using .includes instead of .indexOf.

src/movies.js
```diff
@@ -1,16 +1,150 @@
...
+
+// 
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+function ratesAverage(movies){
+    return  movies.length ? Number((movies.reduce(function(acc, movie){
+        return acc + (movie.rate ? movie.rate: 0);
+    }, 0)/movies.length).toFixed(2)) : 0;
+}
```
IH-WebDev-TA-Remote, line 25: Great solution! really advanced, using ternaries twice. Simply amazing.

src/movies.js
```diff
@@ -1,16 +1,150 @@
...
+        else {
+            return Number(((dramaMovies.reduce((acc, filteredMovie) => {
+                return acc + filteredMovie.rate;
+             }, 0))/dramaMovies.length).toFixed(2));
+        }
+    }         
+     else {
+         return 0;
+     }
+}
```
IH-WebDev-TA-Remote, line 45: Again, really good solution. A lot of nested conditionals that could be simplified but still, for this levels is really good.

src/movies.js
```diff
@@ -1,16 +1,150 @@
...
+
 
-// Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies){
+    if (movies.length) {
+        return movies.sort(function(movie1, movie2){
```
IH-WebDev-TA-Remote, line 50: sort method mutates the original array, so avoid this remember to deep copy it before applying the sort method.

src/movies.js
```diff
@@ -1,16 +1,150 @@
...
+    if (clonedMovies.length) {
+        return clonedMovies.sort(function(movie1, movie2){
+            if (movie1.title < movie2.title){
+                return -1;
+            }
+            else if (movie1.title > movie2.title){
+                return 1;
+            }
+            return 0;
+        }).slice(0, 20).map(function(filteredMovies){
```
IH-WebDev-TA-Remote, line 84: Very advanced chaining the methods together. Kudos!

IH-WebDev-TA-Remote:
Very few feedback to give, simply amazing solutions and very advance way of writing the code. Again, if you are able to code like this and understand everything that is going on you should not have any issues. Keep it up!

PR #1687, opened 2020-10-02
[FT-BCN-092020 Erifyli Evangelou]

erifylo:


src/movies.js
```diff
@@ -1,16 +1,93 @@
...
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies){
+let dramaSteven = movies.filter (function (check){
+        return (check.genre.includes("Drama") && check.director === "Steven Spielberg");
+});
+return dramaSteven.length};
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
```
ross-u, line 23: It would be god to make it a bit more descriptive

PR #1705, opened 2020-10-15
[FT-AMS-102020] Emad Ghnim

Emad-doz:


src/movies.js
```diff
@@ -1,16 +1,168 @@
...
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies) {
+    const dramaSpielbergMovies = movies.filter((movie) => {
+       return movie.genre.includes('Drama') && movie.director === 'Steven Spielberg';
+    })
+    return dramaSpielbergMovies.length;
+}
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
```
LloydC, line 29: Well done Emad!

Here's a different but similar solution, just to give you another perspective
`function ratesAverage(lotsOfMovies) {

if (!lotsOfMovies.length) {
return 0;
}

if (b.rate) {
return a + b.rate;
} else {
return a;
}
}, 0);

}`

src/movies.js
```diff
@@ -1,16 +1,168 @@
...
+    if (a.year === b.year) {
+       if (a.title.toLowerCase() > b.title.toLowerCase()) return 1;
+       if (a.title.toLowerCase() < b.title.toLowerCase()) return -1;
+    }
+}
+
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(movies) {
+    let moviesTitle = movies.map(function (title) {
```
LloydC, line 78: I think you should have named your parameter movie here instead of title, it makes your code easier to understand/read when you name things in a logical manner. It's not a big deal, but in the long run it's a good habit as well as practice.

PR #1706, opened 2020-10-15
[FT-AMS-102020] Elia Chiarucci

eliachiarucci:


src/movies.js
```diff
@@ -1,16 +1,140 @@
...
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(movies) {
+    const dramaSpielbergMovies = movies.filter((movie) => {
+       return movie.genre.includes("Drama") && movie.director === "Steven Spielberg";
+    })
+    return dramaSpielbergMovies.length;
+}
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
-
```
LloydC, line 7: Nice Elia!

Here's a different but similar solution, just to give you another perspective
`function ratesAverage(lotsOfMovies) {

if (!lotsOfMovies.length) {
return 0;
}

if (b.rate) {
return a + b.rate;
} else {
return a;
}
}, 0);

}`

PR #1707, opened 2020-10-15
[FTAMS102020] Nina van den Braak

ninsajala:


src/movies.js
```diff
@@ -1,16 +1,89 @@
...
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(array){
+    const allTitles = array.map(movie => { 
+        const container = movie.title
+        return container})
+
+    const sortedTitles =  allTitles.sort()
+    let top20 = sortedTitles.slice(0, 20)
+    return top20
+}
```
LloydC, line 78: Good work Nina!

Here's a solution that resembles yours but uses a slightly different approach
  return [...lotsOfMovies]
    .sort((a, b) => {
      if (a.title > b.title) {
        return 1;
        return -1;
      } else {
        return 0;
      }
    })
    .slice(0, 20);
}`

src/movies.js
```diff
@@ -1,16 +1,89 @@
...
+    const allRatings = drama.map(movie => { 
+    const container = movie.rate
+    
+    return container})
+
+    let total = allRatings.reduce(function(sum, item){
+        return sum + item
+    })
+    return Math.round((total/drama.length)*100)/100
+}
```
LloydC, line 58: It would have been more efficient to make use of your `ratesAverage` function

`function dramaMoviesRate(array){
    const drama = array.filter( item =>
    item.genre.includes('Drama')
    )
    return ratesAverage(drama);
}`

PR #1708, opened 2020-10-15
[FTAMS102020] Sophie Tichelaar

Sophietic:
1-6 done

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,82 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(array){
+    return array.map(movie=>movie.director);
+}
+
+getAllDirectors(movies)
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,15 +1,82 @@
...
+       else if(a > b){
+        return 1;
+      }
+      else{
+        return 0;
+      }
+    })
+    return alphaList.slice(0,20)
+    }
+
```
LloydC, line 76: Nice work Sophie!

This isn't really a different approach but here's another way to write your solution
  
return [...lotsOfMovies]
    .sort((a, b) => {
      if (a.title > b.title) {
        return 1;
        return -1;
      } else {
        return 0;
      }
    })
    .slice(0, 20);
}`

src/movies.js
```diff
@@ -1,15 +1,82 @@
...
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(array){
 
+    let orderList = array.sort(function(a,b){ 
+   
+     return a.year - b.year;
+   })
+   return orderList
+   }
+
```
LloydC, line 55: Another way to write `orderByYear`

`function orderByYear(array){
  return array.sort( (a, b) => a.year - b.year); 
}`

PR #1715, opened 2020-10-15
[FTAMS102020] Rich Alossi

alossix:


src/movies.js
```diff
@@ -1,16 +1,93 @@
...
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(arr) {
+  let dramaMovies = arr.filter((movie) => {
+    return movie.genre.includes("Drama");
+  });
+  let spielbergDramas = dramaMovies.filter((movie) => {
+    return movie.director.includes("Steven Spielberg");
+  });
+  return spielbergDramas.length;
+}
```
LloydC, line 20: Good work Rich!

Here's an alternative way to write function `howManyMovies`

`function howManyMovies(someMovies) {
 
 return someMovies.filter(
    eachMovie =>
      eachMovie.director === 'Steven Spielberg' &&
      eachMovie.genre.includes('Drama')
  ).length;
}`

src/movies.js
```diff
@@ -1,16 +1,93 @@
...
+  }
+  let dramaRatesArr = dramaMoviesArr.map((r) => {
+    return r.rate;
+  });
+  let dramaRatesTotal = dramaRatesArr.reduce((acc, currVal) => {
+    return acc + currVal;
+  });
+  return Number((dramaRatesTotal / dramaRatesArr.length).toFixed(2));
+}
+
```
LloydC, line 58: You can replace lines 50-56 with `return ratesAverage(arr)`. Also I think lines 47-49 would be better placed at the top before line 44 (it makes more sense to check the length of the array first)

src/movies.js
```diff
@@ -1,16 +1,93 @@
...
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(arr) {
+  if (arr.length === 1) {
+    return arr;
+  }
+  let moviesCopy = arr.map((movie) => {
```
LloydC, line 65: I don't think you need `.map()`, you could just have it `let moviesCopy = arr`

src/movies.js
```diff
@@ -1,16 +1,93 @@
...
+  let titlesList = arr.map(function (t) {
+    return t.title;
+  });
+  titlesList.sort();
+  if (titlesList.length < 20) {
+    return titlesList;
+  }
+  return titlesList.slice(0, 20);
+}
+
```
LloydC, line 90: Nice, I just want to show you a solution where everything is chained together 

return [...lotsOfMovies]
    .sort((a, b) => {
      if (a.title > b.title) {
        return 1;
        return -1;
      } else {
        return 0;
      }
    })
    .slice(0, 20);
}`

PR #1723, opened 2020-10-16
[FTAMS102020] Emma Northridge

emmanorthridge:
Compulsory iterations completed - all green, apart from iteration 5.
Will work on bonuses today.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,75 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(movies) {
+    const arr = movies.map((movies) => movies.director);
+    return arr;
+  }
+
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,75 @@
...
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(movies){
+    const titles = movies.map(movie => {
+        return movie.title})
+
+   const alphaSort = titles.sort();
+    return alphaSort.slice(0, 20);
+    }
+
```
LloydC, line 68: Good work Emma!

Here's a different solution that makes use of the spread operator that we talked about yesterday

    return [...lotsOfMovies]
    .sort((a, b) => {
      if (a.title > b.title) {
        return 1;
        return -1;
      } else {
        return 0;
      }
    })
    .slice(0, 20);
}`

PR #1726, opened 2020-10-16
[FTAMS2020]issaGruber

GruberI:


src/movies.js
```diff
@@ -1,16 +1,102 @@
...
+    const dramaRates = dramaArr.map(films => films.rate);
+
+    console.log(dramaRates);
+
+    const sum = dramaRates.reduce((totalRate, singleRate) => totalRate + singleRate)
+    return +(sum / dramaRates.length).toFixed(2);
+  
+    console.log(sum);
+  }
+
```
LloydC, line 51: You can replace line 42-49 with `return ratesAverage(dramaArr)`

src/movies.js
```diff
@@ -1,16 +1,102 @@
...
+        return 0
+      }
+    })
+  
+    let onlyTitleList = sortedTitleList.map (items => items.title);
+      moviesTitleList = onlyTitleList.slice(0,20);
+  
+    console.log(moviesTitleList)
+  })
+
```
LloydC, line 97: Good work, just so you know you can chain everything together 
`function orderAlphabetically(movies) {
  return [...movies]
    .sort((a, b) => {
      if (a.title > b.title) {
        return 1;
        return -1;
      } else {
        return 0;
      }
    })
    .slice(0, 20);
}`

PR #1728, opened 2020-10-16
[FTAMS102020] Alex Kantsadis

teof95:


src/movies.js
```diff
@@ -1,16 +1,88 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(array){
+    const allDirectors = array.map(movie => { 
+    const container = movie.director
+    
+    return container})
+    return allDirectors
+    } 
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
```
LloydC, line 11: Good work Alex, here's a shorter way to write the same function

```function getAllDirectors(movies) {
  const directors = movies.map(oneMovie => oneMovie.director);
  return directors;
}```

File: src/movies.js [hunk 2]
```diff
@@ -1,16 +1,88 @@
...
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(array){
+    const drama = array.filter( item =>
+    item.genre.includes('Drama')
+    )
+    if (!drama.length){
+    return 0
+    }
+
```
LloydC, line 48: You should interchange line 45-47 with line 42-44, it makes more sense to check first the length of the array before you try to filter it

src/movies.js
```diff
@@ -1,16 +1,88 @@
...
+    const container = movie.rate
+    
+    return container})
+    
+    let total = allRatings.reduce(function(sum, item){
+    return sum + item
+    })
+    return Math.round((total/drama.length)*100)/100
+    } 
+
```
LloydC, line 59: You can replace line 49-57 with `return ratesAverage(drama);`

src/movies.js
```diff
@@ -1,16 +1,88 @@
...
+function orderAlphabetically(array){
+    const allTitles = array.map(movie => { 
+    const container = movie.title
+    return container})
+    
+    const sortedTitles = allTitles.sort()
+    let top20 = sortedTitles.slice(0, 20)
+    return top20
+    } 
+
```
LloydC, line 82: Nice, you can actually chain everything instead of using multiple variables if you want

```
function orderAlphabetically(array){
return [...array].map(movie => movie.title).sort().slice(0,20)
}

PR #1729, opened 2020-10-16
[AMS-FT-102020]delia.drel

deframed:
I've had lots of help..find js in general difficult

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,111 @@
 // Iteration 1: All directors? - Get the array of all directors.
+// create a function = getAllDirectors with parameters movies
+
+
+     const getAllDirectors = (moviesList) => {
+         return moviesList.map((movie) => {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,111 @@
...
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 
+function orderAlphabetically(movies) {
+    let moviesTitle = movies.map(function (title) {
+        return title.title;
+      });
+    let moviesTitleSort = moviesTitle.sort();
+    return moviesTitleSort.slice(0, 20);
+}
+
```
LloydC, line 87: Nice, here's another way you could have written orderAlphabetically without the variables

```
function orderAlphabetically(movies){
return [...movies].map(movie => movie.title).sort().slice(0, 20);
}

PR #1730, opened 2020-10-16
[FTAMS102020] Ahmet Demirlicakmak

codebrokerad:
Iteration 5 has defined functions but it is not complete

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,13 +1,49 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+const getAllDirectors =(movies) => {
+    const directorlist= movies.map((getAllDirectors)=>{
+    return getAllDirectors.director;
+    })
+    return directorlist
+    }
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+const howManyMovies = (movies) => {
+const listOfSteveDramaMovies = movies.filter((directorNameList) => {
+return directorNameList.director === "Steven Spielberg" && directorNameList.genre.includes("Drama")
+});
+return listOfSteveDramaMovies.length;
+}
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+const ratesAverage =(movies) => {
 
+    var average = movies.reduce(function (avg, item) {
+        return avg + ((item.rate) / movies.length);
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,72 @@
...
-// Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+const dramaMoviesRate =(movies) => {
+    const dramaMovieList = movies.filter((movie) => {
+        return movie.genre.includes("Drama");
+        
+    });
+    if (!Array.isArray(dramaMovieList) || !dramaMovieList.length) {
+        return 0;
+    }
+
```
LloydC, line 45: You should interchange line 42-44 with line 38-41, it makes more sense to write your function in that sequence

src/movies.js
```diff
@@ -1,16 +1,72 @@
...
+    });
+    if (!Array.isArray(dramaMovieList) || !dramaMovieList.length) {
+        return 0;
+    }
+ 
+ var average = dramaMovieList.reduce(function (avg, item) {
+   return avg + ((item.rate) / dramaMovieList.length);
+}, 0);
+return parseFloat(parseFloat(average).toFixed(2));
+}
```
LloydC, line 51: You can replace line 46-50 with `return ratesAverage(dramaMovieList);`

src/movies.js
```diff
@@ -1,16 +1,72 @@
...
 
-// Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
```
LloydC, line 52: Here's an example of orderByYear, let me know if you have questions about it
```
function orderByYear(lotsOfMovies) {
  let newArray = [...lotsOfMovies];
  newArray.sort((a, b) => {
    if (a.year > b.year) {
      return 1;
    } else if (b.year > a.year) {
      return -1;
    } else {
      if (a.title > b.title) {
        return 1;
        return -1;
      }
      return 0;
    }
  });
  return newArray;
}

File: src/movies.js [hunk 4]
```diff
@@ -1,16 +1,72 @@
...
+}
+// Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies) {
+    let moviesTitle = movies.map(function (title) {
+        return title.title;
+      });
+    let moviesTitleSort = moviesTitle.sort();
+    return moviesTitleSort.slice(0, 20);
+}
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
```
LloydC, line 71: Nice, you can write orderAlphabetically without the variables
```
 return movies.map(movie => movie.title).sort().slice(0, 20);
}

PR #1731, opened 2020-10-16
[FTAMS102020] CeydaUlubas- lab-javascript-greatest-movies

ceydaulubas:
I will fix Iteration-4.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,115 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+const getAllDirectors = (moviesList) => {
+  const movieDirectorsList = moviesList.map((x) => x.director);
+  
+  return movieDirectorsList;
+}
+
+//just for debugging purposes
+let movies = [
+  {
+    "title": "The Shawshank Redemption",
+    "year": 1994,
+    "director": "Frank Darabont",
+    "duration": "2h 22min",
+    "genre": [
+      "Crime",
+      "Drama"
+    ],
+    "rate": 9.3
+  },
+  {
+    "title": "The Godfather",
+    "year": 1972,
+    "director": "Steven Spielberg",
+    "duration": "2h 55min",
+    "genre": [
+      "Crime",
+      "Drama"
+    ],
+    "rate": 9.2
+  },
+  {
+    "title": "The Godfather: Part II",
+    "year": 1974,
+    "director": "Francis Ford Coppola",
+    "duration": "3h 22min",
+    "genre": [
+      "Crime",
+      "Drama"
+    ],
+    "rate": 9
+  }]
+
+getAllDirectors(movies);
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,140 @@
...
+howManyMovies(movies);
 
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+const ratesAverage = arr =>{
+  if(arr.length ===0) {return 0};
+  let collectRate= arr.map(num=> num["rate"]);
+  let sumRate= collectRate.reduce((acc, curr) => acc + curr);
+  let averageRate= ((sumRate/arr.length).toFixed(2));
+  console.log(averageRate);
```
LloydC, line 100: You should return what you get at the end of your function (best practice)
 `return averageRate`

src/movies.js
```diff
@@ -1,16 +1,140 @@
...
+const dramaMoviesRate = arr =>{
+  let dramaMovies= arr.map(film =>film["genre"].includes("Drama"));
+  if (dramaMovies.length === 0){return 0};
+  let collectDramaMovies= dramaMovies.map(film =>film["rate"]);
+  let average= collectDramaMovies.reduce((a,b)=> a+b);
+  let dramaAverageRate=((average/arr.length).toFixed(2));
+  console.log(dramaAverageRate);
+}
+ratesAverage(movies);
+
```
LloydC, line 115: You can replace line 109-112 with `return ratesAverage(arr);`

src/movies.js
```diff
@@ -1,16 +1,140 @@
...
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+const orderByYear = arr =>{
+  let onlyYear= arr.map(date =>date.year);
+  let sortYear= onlyYear.sort((a,b)=> b-a);
+  return sortYear;
+}
+orderByYear(movies);
+
```
LloydC, line 125: Here's how you can write orderByYear, let me know if you have questions about it
```
function orderByYear(lotsOfMovies) {
  let newArray = [...lotsOfMovies];
  // spread operator is to make sure not to mutate the original array because .sort() does change/mutate the original array so always make sure you create a safe copy before sorting
  newArray.sort((a, b) => {
    if (a.year > b.year) {
      return 1;
    } else if (b.year > a.year) {
      return -1;
    } else {
      if (a.title > b.title) {
        return 1;
        return -1;
      }
      return 0;
    }
  });
  return newArray;
}

File: src/movies.js [hunk 4]
```diff
@@ -1,16 +1,140 @@
...
 
+const orderAlphabetically = arr =>{
+  let moviesTitle = arr.map(name => name.title)
+  let titleArray = moviesTitle.sort()
+  //from 0 to 20
+  let slicedTitleArray = titleArray.slice(0, 20)
+  return slicedTitleArray;
+}
+orderAlphabetically(movies);
+
```
LloydC, line 136: Nice, you can also write orderAlphabetically without the variables
```
const orderAlphabetically = arr => {
return arr.map(name => name.title).sort().slice(0, 20)
}

PR #1732, opened 2020-10-16
[FTAMS102020] Mahalakshmi

mahavinay:
Completed without bonus, will update with bonus later in weekend

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,95 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+
+const getAllDirectors = (movies) => {
+const directorList = movies.map((names) => {
+    return names.director;
+  });
+return directorList
+}
+ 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,15 +1,95 @@
...
+
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+function orderAlphabetically(movies){
+    const titleOrderList = movies.map((titleList) => {
+      return titleList.title;
+    })
+    let orderOfList = titleOrderList.sort();
+   return orderOfList.slice(0,20)
+}
+
```
LloydC, line 92: Nice, you can also write orderAlphabetically without the variables
```
function orderAlphabetically(movies){
return movies.map(movie => movie.title).sort().slice(0, 20);
}

PR #1733, opened 2020-10-19
[PTMIA102020] Rachel Feldman

RFeldman84:


RFeldman84:
is there an app or setting for VSCode to use all javascript versions? mine will not recognize startsWith() method which says its a new method so i'm guessing not es6 which my editor is set up to read

PR #1782, opened 2020-10-23
Berlin WDFT Oct2020 Anna Waszak

annawaszak:
still working on the bonus


src/movies.js
```diff
@@ -1,16 +1,122 @@
...
+const getAllDirectorsUnified = (arr) => {
+    const unified = [];
+    for (let i = 0; i < arr.length; i++) {
+        if (!unified.includes(arr[i])) {
+            unified.push(arr[i]);
+        }
+    } 
+    return unified;
+}
+
```
timbotimber, line 20: how could you do this with a filter method?

timbotimber (commented):
Real nice work, good job. Let us know if you finish the final bonus!

PR #1783, opened 2020-10-23
[Berlin- Oct 2020- FT] Mahmut

lighteronn:


src/movies.js
```diff
@@ -1,16 +1,94 @@
...
+        }
+    })
+    return sortByYear.slice()
+ }
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 
+  function orderAlphabetically(arr) {
+   
+    function compare( a, b ) {
```
pkevinklein, line 75: Hey, great job with the lab! Only a couple of things to mention: Try using expressive words for the variables instead of a single letter. Also, I noticed that you created a "compare" function. This one can be created outside the scope of the mandatory functions for the lab and then be reused inside of them.

src/movies.js
```diff
@@ -1,16 +1,94 @@
...
+    function dramaMoviesRate(arr) {
+        let dramas = arr.filter(function(f) {
+            return f.genre.includes("Drama")
+        })
+        let sum = dramas.reduce(function(acc, val) {
+             return (acc+val.rate)   
+        },0)
+        if(!sum) {return 0}
+        return +(sum/dramas.length).toFixed(2)
+    }
```
pkevinklein, line 40: In this one you could have reused the ratesAverage function inside the dramaMoviesRate function. Instead of having to write it again from scratch. Other than that, excellent! Keep up the hard work!

PR #1784, opened 2020-10-23
Berlin Oct 2020 WD FT Facundo Gomez

FacundoGo:


src/movies.js
```diff
@@ -1,16 +1,83 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(array){
+    const directors = array.map(function(el){
+      return el.director
```
timbotimber, line 5: try to use more specific variable names so you can easily see what you're looking for in the future.

src/movies.js
```diff
@@ -1,16 +1,83 @@
...
+    })
+    return directors
+}
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(array){
+    let counter = 0;
+    let steve = array.filter(function (el){
```
timbotimber, line 15: You don't need to assign this filter function to a variable. It works with but it would be cleaner without.

timbotimber (commented):
super nice, great effort man. Nice to see that you used JSON. methods in the last iterations.

PR #1785, opened 2020-10-23
BER-WEBDEV-FT-Ale Rodriguez

prullansky:
Done with the three first exercises

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,10 +1,34 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(arr) {
+    const allDirectors = arr.map(function(movie){
+    return movie.director;
+    })
+    return allDirectors;
+    }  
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,60 @@
...
+    })
+    const dramaAvr = ratesAverage(dramaMovies);
+    return dramaAvr;
+    }
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-
+function orderByYear(arr) {
+    const sorted = arr.slice().sort((a, b) => b.year - a.year);
+    return sorted.reverse();
+    }
```
timbotimber, line 48: this works but you could switch your sort around and not use the reverse()

timbotimber (commented):
Looks good! If you'd like to attempt the bonuses in a group or later in the week let us know

PR #1790, opened 2020-10-24
Berlin-WDFT-Oct20-Katie Gracey

kgracey93:
All iterations complete!

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,119 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(array) {
+  const allDirectors = array.map(function (movie) {
+    return movie.director;
+  });
+  return allDirectors;
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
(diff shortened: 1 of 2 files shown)


src/movies.js
```diff
@@ -1,15 +1,119 @@
...
+    if (array.length === 0) {
+      return 0;
+    }
+  
+    let valuesOnly = array.map(function (movies) {
+      return movies.rate;
+    });
+    console.log(valuesOnly);
+  
+    function bouncer(arr) {
```
Tyouxik, line 37: I think you made your life harder here, a simple reduce method with an if statement inside for the edge cases would have been enough.
But it is nice that you experimented :)

src/movies.js
```diff
@@ -1,15 +1,119 @@
...
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(array) {  
+  let dramas = array.filter(function (movie) {
+    if (movie.genre.includes('Drama')) {
+      return true;
+    } 
+  });
+
+  let dramasSum = dramas.reduce(function (sum, review) {
```
Tyouxik, line 58: You create a rateAverage function right before, you can use it here :)

Tyouxik (commented):
Good job, you have a good understanding of array methods

src/movies.js
kgracey93: Thanks for the feedback I really appreciate it!

PR #1791, opened 2020-10-25
WDFT Oct 2020 - Berlin - Ineke

methner:


src/movies.js
```diff
@@ -1,16 +1,83 @@
...
+        return sortedMovies
+    }
+    
+    orderByYear(movies);
+    
+    // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+    
+    function orderAlphabetically(movieArray) {
+        let sortedByTitle = movieArray.title.sort(function(a, b) {
+        if movieArray.title < 20 {
```
Tyouxik, line 74: movieArray.title is not a number, it is a string. It doesn't make sense to use <20 here

src/movies.js
```diff
@@ -1,16 +1,83 @@
...
+        })
+        return sortedMovies
+    }
+    
+    orderByYear(movies);
+    
+    // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
+    
+    function orderAlphabetically(movieArray) {
+        let sortedByTitle = movieArray.title.sort(function(a, b) {
```
Tyouxik, line 73: inside your sort method, you need to use your parameters a and b. That's how sort knows how to sort.

Tyouxik (commented):
You understand most Javascript concepts, with more practice, it is gonna become clearer.
Continue like that

PR #1794, opened 2020-10-25
[WDFT BER OCT2020] Jan Völker

jan-webdev:


src/movies.js
```diff
@@ -1,15 +1,78 @@
...
+      if (movies.length === 0) {
+        return 0;
+      }
+    // Create new array with only Dramas  
+      let onlyDramaMovies = movies.filter(function (movie) {
+        if (movie.genre.includes("Drama")) {
+          return true;
+        }
+      });
+    // Reduce the new Array to sum
```
Tyouxik, line 49: If only you had just created a function that calculates the average rate of an array of movies :)

Tyouxik (commented):
Great job, you clearly understand how to use map, filter and sort.

src/movies.js
jan-webdev: :D Well that was not very DRY of me. Thanks for catching that.

PR #1796, opened 2020-10-25
Berlin - FT - Juliana Caruso

juliana-caruso:


src/movies.js
```diff
@@ -1,16 +1,121 @@
-// Iteration 1: All directors? - Get the array of all directors.
+
+  
+  // Iteration 1: All directors? - Get the array of all directors.
+
+  function getAllDirectors(directors) {
```
Tyouxik, line 5: Your parameter "directors", this function will receive an array of movies.
Also in your map function, "person" refers to each movie in the array.

Naming is one of the hardest thing in coding, always name your variable or function for what they are.

src/movies.js
```diff
@@ -1,16 +1,121 @@
...
+function ratesAverage(movies) {
+  if (movies.length === 0) {
+    return 0;
+  } else {
+    const totalRates = movies.map(function(value) {
+      return value.rate;
+    });
+    let emptyOnes = 0;
+    let averageRate = totalRates.reduce(function(acc, val) {
+      if (val) {
```
Tyouxik, line 40: Here "val" represents a "movie", what you want to check is if the movie(here named val) has a rate.

src/movies.js
```diff
@@ -1,16 +1,121 @@
...
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+const numberOfDramas = movies.filter(function(drama) {
+  if (drama.genre.includes('Drama')) {
+   return drama;
+  }
+})
+
+function dramaMoviesRate(movies) {
```
Tyouxik, line 66: This function doesn't filter the drama movies, your variable numberOfDramas should be inside this function not before :)

src/movies.js
```diff
@@ -1,16 +1,121 @@
...
+  return parseFloat(averageRate.toFixed(2));
+}
+}
+dramaMoviesRate(numberOfDramas);
+
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+  function orderByYear(movies) {
+    const sortYear = movies.map(function(getYear) {
```
Tyouxik, line 93: getYear represents each movie in the movies array, a better name would be movie.
if you don't know how to name it, you can console.log the variable inside map and see what it really is. Then you can modify the name of the variable.

src/movies.js
```diff
@@ -1,16 +1,121 @@
...
+      return getYear.year
+    })
+    const sorted = sortYear.sort(function (a, b) {
+    return a - b;
```
Tyouxik, line 97: Nice :)

Tyouxik (commented):
Good job, you understand how to use map, reduce, filter and sort.
Take your time to name your variable and function :)

PR #1797, opened 2020-10-25
Berlin-WDFT-Oct2020-Davide-Andreani

dialmojordan:


timbotimber (commented):
looks good! You didn't need to move the movies list into the same file but it doesn't really matter.

dialmojordan:
Sure, thank you Tim

PR #1803, opened 2020-10-26
BER WD FT OCT 2020 - Pedro Blumenschein - lab-javascript-greatest-movies

PedroBlumenschein:


src/movies.js
```diff
@@ -1,16 +1,111 @@
...
+function howManyMovies(arr) {
+
+    return arr.filter(function(item) {
+      if (item.director === 'Steven Spielberg' && item.genre.includes('Drama')) {
+        return true;
+      }
+    }).length
+    
+  };
+
```
pkevinklein, line 22: Great work Pedro! I just want to remind you that filter returns something as long as it is true. Therefore, you don't necessarily need an if statement. Your if statement here could also been written like this:
return movie.director === 'Steven Spielberg' && movie.genre.includes('Drama').
If anything, i recommend you to check the MDN documentation.

src/movies.js
```diff
@@ -1,16 +1,111 @@
...
+            return 1;
+        }
+        else if (a.title < b.title) {
+            return -1
+        }
+    });
+    return ordered;
+  }
+  
+
```
pkevinklein, line 84: I noticed that you created two variables: sorted and ordered. Instead of using two different functions that do the same, you could reuse one several times. Also in the dramaMoviesRate you can reuse the ratesAverage function after mapping over the Drama movies for the same result. Try that out!

PR #1805, opened 2020-10-26
lots of errors

WilliamCampbellie:


src/movies.js
```diff
@@ -1,16 +1,158 @@
...
+
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 
+function howManyMovies(movies) {
+    let moviesArray = movies.filter(function(movie) {
+      if (movie.director == "Steven Spielberg" && movie.genre.includes ('Drama')) {
```
pkevinklein, line 22: Great work William! I just want to remind you that filter returns something as long as it is true. Therefore, you don't necessarily need an if statement. Your if statement here could also been written like this:
return movie.director === 'Steven Spielberg' && movie.genre.includes('Drama').
If anything, i recommend you to check the MDN documentation

src/movies.js
```diff
@@ -1,16 +1,158 @@
...
+    let moviesArray = movies.map(function(movie) {
+    if (movie.genre.includes ('Drama')) {        
+        totalRating = totalRating + movie.rate
+        return totalRating;}
+    });
+  if (moviesArray.length == 0) {
+      return 0 };
+return  Math.round(((totalRating/moviesArray.length) + Number.EPSILON) * 100) / 100
+}
+
```
pkevinklein, line 56: I noticed that you created two "compare" functions. Instead of using two different functions that do the same, you could reuse one several times. Also in the dramaMoviesRate you can reuse the ratesAverage function after mapping over the Drama movies. Try that out!

PR #1856, opened 2020-11-13
[PAR-WDFT-911] Malek Fouhal - Mathias Valverde

Denak39:


VProum:
Hi Malek and Mathias,
nice job overall, i feel like you could do the bonus ;)
can you share this feedback with your partner? i don't think he will receive an email

Good job 👍

Good job again 👍

Good job 👍 and you used reduce 💯
tiny detail, a length of an array will never be negative, so you can use 
if (movies.length === 0)

Good job 👍
what is the line 49 purpose ?
`const averageRound = averageRate;`
you can stick with averageRate

Good job 👍
you are using a lots of "if else" with a single operation inside, you can try to use ternary operator to write simpler and faster code :
```
if (a.title > b.title) {
        return 1;
        return -1;
      } 
```
`a.title > b.title ? 1 : -1`
those two are the same

Good job 👍 short and nice 

🥇

PR #1857, opened 2020-11-14
[PAR-WDFT-0911 -EDOUARD-PFAUWADEL] 

lepfau:
Still have to figure out everything but I think i understood the main concept of HOF


Fanny-Sbn:
Hello Edouard, thanks for submitting your lab. It's a really good job!

Here are a few comments : 

Iteration 1 : You could write your function in one line 

    return movies.map(movie => movie.director)
}

Iteration 2 :

Fanny-Sbn:
Hello Edouard, thanks for submitting your lab. It's a really good job!

Here are a few comments :

Iteration 1 & 2 : 

You could write your functions in one line

return movies.map(movie => movie.director)
}

function howManyMovies(movies) {
    return movies.filter(movie => movie.director === "Steven Spielberg" && movie.genre.includes("Drama")).length
}

Iteration 3 : 

Good but you actually don't need to do a first map as you using a reduce, meaning you can access the movies' rates directly in it 

    if (!movies.length) return 0;

    }, 0)

}

Iteration 4 :

Almost there but you don't check if the initial array don't have any drama movies, to make it clearer you should reuse the previous function ratesAverage()

function dramaMoviesRate(movies) {
    const dramaMovies = movies.filter(movie => movie.genre.includes("Drama"));
}

Iteration 5 & 6 : 👍 

Very good job overall!

PR #1858, opened 2020-11-14
[PAR-WDFT-911] René Koné & Victorien Provenzano

vicpzn:
lab done, no bonuses for now.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -78,4 +78,41 @@ function orderAlphabetically(array) {
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 
+function getHoursinMinutes(array) {
+  let hours = [];
+  for (let i = 0; i < array.length; i++) {
+    if (array[i].duration.includes("h")) {
+      hours.push(Number(array[i].duration.slice(0, 1) * 60));
+    } else {
+      hours.push(Number(array[i].duration.slice(0, 2)));
+    }
+  }
+  return hours;
+}
+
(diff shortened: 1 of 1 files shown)


vicpzn:
Added bonus 1.

Everything is okay but Jasmine doesn't seem to validate the following for the first bonus:
- Should return a new array, not update the original one
- The duration of the movie should be a number

But, strangely enough, whenever I click on "The duration of the movie should be a number" I get no failure and the condition seems validated.

Fanny-Sbn:
Hello René & Victorien, thanks for submitting your lab. 

Here are a few comments :

Iteration 1 : 👍 

Iteration 2 : 
It's working but you could use only once the filter function and simplify your function 

    const SpielbergDramaMovies = array.filter(movie => movie.director === "Steven Spielberg" && movie.genre.includes('Drama'))
    return SpielbergDramaMovies.length
}

Iteration 3 : 👍 

Iteration 4 : 
It's working but you overcomplicated the whole thing, you have just cretad before a ratesAverage function, what's about reusing it? 

function dramaMoviesRate(array) {
    let amountMoviesDrama = array.filter((n) => n.genre.includes("Drama"));
    return ratesAverage(amountMoviesDrama)
}

Iteration 5 : 👍

vicpzn:
Thank you Fanny for the feedback, we have taken your comments into account! 😃
Regarding the iteration 3 & 4, we coded them separately, this why we forgot to reuse the average function.

PR #1860, opened 2020-11-14
[PAR-WDFT-911] TRAN Thi Thanh Xuan

tranttxuan:
I have 2 questions:
- How to import variable movies in file data.js into movies.js
  In file data.js, I added: export let movies = {};
  In file movies.js, I coded: import * as movies from "./data";
 => App crashed

- In iteration 7, I couldn't clone the initial array by using method [..array] map() but when using JSON.parse, it worked.


N10T:
Hello Xuan !

Have you did it alone ? 🥇 Jasmine is really happy ! full green

 to export:
```
in data.js
  module.exports = { movies }

in movies.js
  const data = require('./data')
```


Iteration 1: All directors - Get the array of all directors.
-
Good job 👍 
A shorter way 
```
const getAllDirectors = array => array.map((item) => item.director)
```


Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
Good job again 👍, 

A shorter way ? :)
```
const howManyMovies = array => array.filter(
  ({director,genre}) => director === "Steven Spielberg" && genre.includes("Drama")).length;
```


Iteration 3: All rates average - Get the average of all rates with 2 decimals
-

Good job 👍  and you used reduce 💯 


Iteration 4: Drama movies - Get the average of Drama Movies
-

Good job 👍 but prefer to use filter, which give you back a new array
a shorter way : 
```
const dramaMoviesRate = (array) => ratesAverage(array.filter(movie => movie.genre.includes("Drama")))
```


Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-
```
a.title < b.title
```
This is not the best way to compare a string, prefer to use a method like locale compare
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/String/localeCompare
sort is not sending a new array but modifying the original one, in this case you can do it like this 
```
const orderByYear = array => [...array].sort(function (a, b) {
    //sorted year
    else if (a.year > b.year) return 1;

    //same year, sort title
    if (a.title.localeCompare(b.title)) return -1;
    return 1;
  });
```


Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-
As I said in the previous iteration prefer the locale compare


BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
-

Good job ! 👍  I'm not sure you understand the regex, this one do the job `/\W/g`


BONUS - Iteration 8: Best yearly rate average - Best yearly rate average
-
Good job ! You used a reduce to create an object 🥇 

**🥇**

tranttxuan:
I did it with my older sister Google 😀. Problem is that I must spend a lot
of time searching on the Internet and that I can't remember syntax
functions.
Thanks for your review.

PR #1862, opened 2020-11-15
[PAR-WDFT-911] Lucas Souleyrasse - Etouma Bela

LucasSlrs:
Almost all done, no bonus.


VProum:
Hi Etouma and Lucas,
Nice job overall! you even used ternary operator!

Good job 👍
A shorter way
`const getAllDirectors = array => array.map((item) => item.director)`

Good job again 👍

Good job 👍 and you used reduce 💯 but you are not managing the fact that movies may have not have a rate : 
simple fix, line 32 :
`acc += isNaN(val.rate) ? 0 : val.rate;`

Same as above

you have to clone your array before the sort, not after because you are modifing the initial array when doing that

perfect

🥇

PR #1864, opened 2020-11-15
[PAR-WDFT-911] MENHAJ SAFI

Menhaj01:


Fanny-Sbn:
Hello Menhaj, thanks for submitting your lab. 

Here are a few comments :

Iteration 1: 👍 No need to copy the original array, using map creates a new array

Iteration 2: no need to use a if, you're using a filter so you can directly return all the films when the condition is met 

    let totalMovies = array.filter((drama) => {
        return drama.director === "Steven Spielberg" && drama.genre.includes("Drama")
    });
    return totalMovies.length;
}

Iteration 3: 👍 

Iteration 4: It's working but you could simplify it by using the ratesAverage function you previously created. 

function dramaMoviesRate(arr) {
    const dramaFilms = arr.filter(film => film.genre == "Drama");
    return ratesAverage(dramaFilms)
}

Iteration 5: 👍 

Iteration 6: You could simplify a bit your function by chaining the HOF 

    return arr.map((film) => {
        return film.title;
    }).sort().slice(0, 20)
}

Good job overall Menhaj! Keep on going 💪

PR #1865, opened 2020-11-15
[PAR-WDFT-911] Anojan Mylvaganam - Ann Victoria Duclaux - Ann Lampa

anojanmyl:
iteration 3 failed and bonuses not done


Fanny-Sbn:
Hello Anojan, Anne-Victoria and Ann, thanks for submitting your lab

Here are a few comments :

Iteration 1: 👍 

Iteration 2: 👍 

Iteration 3 :

Almost there, you don't validate the last spec because you're not checking if the movie has a rate. The rest of the logic is good. But you actually don't need to do a first map as you using a reduce, meaning you can access the movies' rates directly in it

if (!movies.length) return 0;
}, 0)
}

Iteration 4: 👍 Nice reuse of ratesAverage function

Iteration 5: Nice but you could use localeCompare function instead 

function orderByYear(array){
  const newSortedArray = array.sort((movie1, movie2) =>{
    if(movie1.year === movie2.year){
    }
  })
  return newSortedArray;
}

Iteration 6: Nice, but same, you could use localeCompare ;) 

Nice job overall :) Wax on, wax off!

PR #1866, opened 2020-11-15
[PAR_FTWD-911] Gwerborc'h Mahéo

Billybillycode:


N10T:
Hello Gwerborc'h  !

Wow you're a console logger addict

Iteration 1: All directors - Get the array of all directors.
-
Good job 👍 
A shorter way 
```
const getAllDirectors = array => array.map((item) => item.director)
```


Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
Good job again 👍 

A shorter way ? :)
```
const howManyMovies = array => array.filter(
  ({director,genre}) => director === "Steven Spielberg" && genre.includes("Drama")).length;
```


Iteration 3: All rates average - Get the average of all rates with 2 decimals
-

Good job 👍  and you used reduce 💯 for the last spec check this to verify if movie.rate exist
```
movies.reduce((ac, movie) => ac + (movie.rate ? movie.rate : 0) , 0);
```


Iteration 4: Drama movies - Get the average of Drama Movies
-

Good job 👍, to check if you have drama movie a way to do it

```
if(!movies.some(movie=>movie.genre.includes("Drama"))) return 0
```

Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-

Not finish but good start 👍 

```
a.title < b.title
```
This is not the best way to compare a string, prefer to use a method like locale compare
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/String/localeCompare

you forget this part to change the object to only a title:
```
let orderArray = newArray.sort((a, b) => (a.title > b.title) ? 1 : -1 ).map(v=>v.title);
```

**🥇**

PR #1867, opened 2020-11-15
[PAR - WDFT - 911] Nathan JEHU, Katalin STOLCZ

The-Child:


Fanny-Sbn:
Hello Katalin & Nathan, thanks for submitting your lab. 

Here are a few comments 

Iteration 1: 👍 
Shortest way to write it : 
const getAllDirectors = (movies) => movies.map(movie => movie.director);

Iteration 2 : 👍 

Iteration 3: 👍 

Iteration 4: One of your spec is not validated... The first part (filter on drama movies) is correct after this try to reuse your ratesAverage function, it's way easier. 

Iteration 5: That's a great beginning but you forgot a part of the iteration, you need to order alphabetically by their titles two movies which have the same release year. 

Iteration 6: Several points here, first you can chain HOF, no need to create a new variable each time you're using a HOF. Seccondly, you don't validate all your specs because the order of the functions you're running matters. Change the order properly to have the expected result. 

You need to change a bit iteration 4 & 6 with the comments I've made and work again on iteration 5, it's not totally finished. The rest is great. Please let me know when you're done ;) 

Wax on, wax off!

PR #1868, opened 2020-11-16
Not finished at all

katepavlova1:
I am only commiting to let you know that I have started working on it but it's too challenging. I am planning to reviez the class and work on this lab a little more.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2967,3 +2967,9 @@ let movies = [
       "rate": 8
     }
   ]
+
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Please name your pull request with the following pattern :  [PAR-FT-0911] Your full name
Otherwise, we won't be able to correct your exercise (students from others campuses are also submitting their labs and we filter with PAR)

PR #1869, opened 2020-11-16
(PARIS-FT-911)Jacqueline Jumaga & Marine Sanjuan

jjumaga:
We didn't get 3 lines - we both tried multiple things that actually ended up taking away other checks in Jasmine.

- Found a useful way to use multiple HOF's in the same function to make it easier to read which we used in iterations 4, 5 and 6.

- There is an order is which to use each HOF depending on what you want to do - this took some time to figure out and was the result of many error message (map before you sort etc.).

Let us know your thoughts! Thanks TA's <3

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,343 @@
+/*let movies = [
+        {
+          "title": "The Shawshank Redemption",
+          "year": 1994,
+          "director": "Frank Darabont",
+          "duration": "2h 22min",
+          "genre": [
+            "Crime",
+            "Drama"
+          ],
+          "rate": 9.3
+        },
+        {
+          "title": "The Godfather",
+          "year": 1972,
+          "director": "Francis Ford Coppola",
+          "duration": "2h 55min",
+          "genre": [
+            "Crime",
+            "Drama"
+          ],
+          "rate": 9.2
+        },
+        {
+          "title": "The Godfather: Part II",
+          "year": 1974,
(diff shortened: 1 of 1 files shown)


VProum:
Hi Jacqueline and Marine!
Good job overall! the tests you didn't pass on jasmine were specific case you should handle (and that's why there is jasmine to remind you of those cases)

Good job 👍 but you don't have to clone the intial array (and your argument name is quite funny, try to stay explicit)
A shorter way 
`const getAllDirectors = array => array.map((item) => item.director)`

Good job again 👍
why use "horse" as a name for your arguments? XD

Good job 👍 and you used reduce 💯
but you are not managing the fact that a movie may not have a rate.
if at one time, a movie doesn't have a rate, you "ratesTotal" will be equal to "NaN"
so "average" will be equal to NaN and Math.round(average * 100) / 100 will be equal to NaN too! 
(Nan is toxic, as undefined is also)
and your average2 will return 0, as NaN is falsy

To solve that, you have to check that your movie has a rate first, by doing a check with an if statement
`if( !isNaN(cur.rate)) `
for instance

same issue as above

You are almost good in this iteration, but what if there is movies with the same year? you are currently not managing this issue, and jasmine want it in alphabetic order!
you have suceeded in iteration 6 so I know you know how to sort in alphabetical order :D

in your sort function, you have to check if the 2 years are the same with an if statement and order them in alphabetic order like in iteration 6 

Nice one! 👍

🥇

PR #1872, opened 2020-11-16
[PAR-WDFT-911] Adam Ghaoul & Mina Samwail

MinaSamwail:
The bonuses are not done yet but will be :)


Fanny-Sbn:
Hello Mina & Adam. Thanks for submitting your lab.

Here are a few comments :

Iteration 1: 👍

Shortest way to write it : 
    return movies.map(array => array.director)
}

Iteration 2: 👍

Iteration 3 :

Almost there, you don't validate the last spec because you're not checking if the movie has a rate. The rest of the logic is good.

if (!movies.length) return 0;
}, 0)
}

Iteration 4: 👍 Nice but you could reuse your ratesAverage function

Iteration 5: Nice but you could use localeCompare function instead

function orderByYear(array){
const sortYear = array.sort((movie1, movie2) =>{
if(movie1.year === movie2.year){
}
})
return sortYear;
}

Iteration 6: Nice, but same, you can chain HOF 

Nice job overall :) Wax on, wax off!

PR #1873, opened 2020-11-16
[PAR-WDFT-911]YUMIAO

yuuumiao:
some bonus parts i didnt manage to pass the Jasmin test, due to the limited time

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2966,4 +2966,10 @@ let movies = [
       ],
       "rate": 8
     }
-  ]
+  ];
+
+
+
+
+
+  module.exports = { movies };
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,202 @@
-// Iteration 1: All directors? - Get the array of all directors.
-// _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+// Goal : Practice JavaScript advanced methods (map, reduce, filter and sort to manipulate arrays
+// var { movies } = require('./data.js');
 
-// Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+
+
+// Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(arr){
+        
+        let arrDirectors = arr.map((movie) => {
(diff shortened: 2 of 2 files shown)


VProum:
Hi Yu,
Excellent work 👍
i see that you are doing lots of console.log :D this is very good, never try to guess the value of your variable and console logging them is a good way to understand what is going on on your programm

Good job with bonus 👍
A shorter way
`const getAllDirectors = array => [...new Set(array.map((item) => item.director))]`

Good job 👍

Good job 👍 
but you are overthinking it :D
in javascript, there are truthy and falsy value so 
if(Boolean(currentValue) == 1) and if(currentValue) are the same
basically you don't need Boolean(toto) and your ternary can even be shorter and cleaner :D

same as above, you don't need Boolean(smthg) and also, you have just created a function returning the average in iteration 3, why not reuse it?

Excellent 👍 

here again some extra work that is not needed all your " || [..arr] or || 'zzzzzz' " will never be reached, as the thing before the || are always truthy

### BONUS - Iteration 7 and 8: Time Format - Turn duration of the movies from hours to minutes
here are an exemple of  solutions : 

```
    var result = [];
    var copy = [...movies];
    result = copy.map(hoursToMinutes);
    console.log(result);
    return result;
}

function hoursToMinutes(movie) {
    var hoursStr = movie.duration;
    if (hoursStr.includes("h")) {
        var strArray = hoursStr.split('h');
        var hours = Number(strArray[0]);
        var minStr = strArray[1].split("min");
        var min = minStr[0];
        min = Number(min.slice(1));
        movie.duration = hours * 60 + min;
        return movie
    } else {
        var minStr = hoursStr.split("min");
        var min = minStr[0];
        min = Number(min);
        movie.duration = min;
        return movie
    }
}


  const yeardObj = movies.reduce((acc, movie) => {
    if (acc[movie.year]) acc[movie.year].push({ rate: movie.rate });
    else acc[movie.year] = [{ rate: movie.rate }];
    return acc;
  }, {});

  let best = { bestYear: 0, avg: 0 };

  for (let key in yeardObj) {
    let avg = ratesAverage(yeardObj[key]);
    if (avg > best.avg) best = { bestYear: key, avg: avg };
  }

  return `The best year was ${best.bestYear} with an average rate of ${best.avg}`;
}

```


🥇

PR #1935, opened 2021-01-25
[RMT-PT-012021]KhrystynaLeshko &LucaManzon

KhrystynaLeshko:
done 6 iterations

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,114 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(movies) {
+  let directors = movies.map((movie) => movie.director);
+  return directors;
+
+  /*let directors = movies.map (function(movie){
+    return movie.director;
+  });
+  return directors; */
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,115 @@
...
+  });
+  if (rates.length === 0) {
+    return 0;
+  }
+  let avg = rates.reduce(function (acc, num) {
+    return acc + num;
+  });
+  let result = avg / rates.length;
+  return Math.round(result * 100) / 100;
+}
```
IH-WebDev-TA-Remote, line 69: Could also have reused the previous ratesAverage() function to calculate the average here! 😄

src/movies.js
```diff
@@ -1,16 +1,115 @@
...
+    } else if (value2.title < value1.title) {
+      return 1;
+    }
+    return 0;
+  });
+  let top20 = sort.filter(function (element) {
+    if (sort.indexOf(element) < 20) {
+      return element.title;
+    }
+  });
```
IH-WebDev-TA-Remote, line 106: Interesting use of the filter method here! Good job! 🏅 Another possible solution would be to use the [_.slice_ method](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice). 👍

IH-WebDev-TA-Remote (commented):
Great job overall! Another lab well done!  🥇

PR #1937, opened 2021-01-25
[WD-PT-JAN2021] Gizem Wanka && Aleksandra Porada

gizeminthecity:


src/movies.js
```diff
@@ -1,3 +1,6 @@
+
+// Gizem Wanka && Aleksandra Porada
+
 // Iteration 1: All directors? - Get the array of all directors.
 
 function getAllDirectors(arrayOfMovies) {
```
IH-WebDev-TA-Remote, line 6: Although a perfectly valid solution, the purpose of this iteration was to try the .map() array method to get you used to it! 💡

PR #1940, opened 2021-01-26
Not yet finished.

lucajans:


src/movies.js
```diff
@@ -1,16 +1,84 @@
...
+      return accumulator + movie.rate;
+    }, 0);
+    return Math.round((averageRate / arr.length) * 100) / 100;
+  }
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(arr) {
+    let dramaMovies = arr.filter(function (movie) {
+        return movie.genre.indexOf("Drama") !== -1})
```
IH-WebDev-TA-Remote, line 40: Another possible solution here would be to use the array method [.includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes)! 💡

PR #1955, opened 2021-02-06
[PAR-WDFT-102] giorgia galasso, paul parant, victoria virlan, naomi yamanaka

giorgiagalasso:
We tried our best, was hard. will try and work on it separately this weekend.


Vika-Vi:
I didn't manage to figure out the last iteration 
struggling with more complex functions
spending a lot of time on one iteration/exercise 
my brain is fried hahah  :D

ta-web-paris:
Hey guys ! 

### ⭐ Iteration 1 :

Good job - for the bonus part, you can remove duplicates with several methods : filter, a forEach loop, or a method called "Set" in Javascript which can do the work for you. Check this link where everything is explained clearly :

### ⭐ Iteration 2 :

Good job - you can use a methode called "includes()" to avoid having to use a loop (which will turn a few lines of code into one line, always cooler) check this link to see how it works : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

That way you can do you filter with two conditions to meet : director is spielberg AND genre includes drama 🧠 

### ⭐ Iteration 3 :

Nicely done ! You used "Math.round()" to get 2 decimals in the rating, and that's working well, but there is one quicker way to do it with the method toFixed(value) and that makes even more sense when you read the code. Both are fine, it's just so you know your options ! Check this link to know how to use it : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

### ⭐ Iteration 4 :

Great re-use of the function of last iteration, that's exactly the good way ! No re-writing the same code, congrats !
Same thing about using "includes" to find the Drama movies in the array

### ⭐ Iteration 5 :

Great job here too ! Duplicating the array, nice to treat the alphabetical sort beforehand, clean code. 

### ⭐ Iteration 6 :

Nice !

### ⭐ Iteration 7 : 

Nicely done ! This one was tricky and you managed to break the problem in small steps, good job. 

### ⭐ Iteration 8 : 

For this one, you forgot to return a string with a sentence instead of just raw data. You also have a small mistake in the formula because the numbers aren't the ones expected. What you have to do : 
1 - break the list by year
2 - calculate the average rating for each year
3 - find the year with the biggest rating
4 - return that (if there are more than one with the biggest rating, return the oldest year)

Don't worry Guillaume will send the correction !

If you have any question on this feedback, something you don't understand / disagree with, or whatever, please come find me on Slack and i'll be happy to answer ! I don't know if you all receive the email with the feedback, so please tell your partner about it so you all can check 

Nina

PR #1956, opened 2021-02-06
[PAR-WDFT-102] Mélusine Rey & Matthieu Cheynut

zelumine:
Done iterations 1 to 6. 
We left out a few errors because we can't find a solution :(
And we didn't do the bonuses either, sorry!

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Jasmine Spec Runner v2.8.0</title>
+
+    <link rel="shortcut icon" type="image/png" href="jasmine/jasmine-2.8.0/jasmine_favicon.png" />
+    <link rel="stylesheet" href="jasmine/jasmine-2.8.0/jasmine.css" />
+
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Melusine and Matthieu, thanks for submitting your lab. 

Here are a few comments: 

Iteration 1: Nice! You could have written it in one line ;)
const getAllDirectors =(movies) => movies.map(movie => movie.director)

Iteration 2: Nice

Iteration 3: Almost there! You just missed one spec, you need to check if the rate exists and if not return 0 for each film (maybe a ternary in your map function ?)

Iteration 4: Nice you reuse your ratesAverage() functtion

Iteration 5: You're overcomplicating the iteration, the yearsOrdered array is enough you just need to make a copy of the original array (spread operator?) before using your sort HOF (its mutating the original array)

Iteration 6: Nice, FYI you can chain HOF
 
function orderAlphabetically(arrMovies) {
  let newArray = arrMovies.map(movie => movie.title)
  .sort((a, b) => a.localeCompare(b))
  .splice(0,20);
    return newArray;
  }

Nice job overall, try the bonuses ;)

PR #1958, opened 2021-02-07
[PAR-WDFT-102] Anaïs ENGLER Myriam MIRA

myracodes:
We did all the iterations but not the bonuses.
Although, we tried several options (all working, according to Jasmine) for some iterations to practice map/reduce/filter/sort

Thanks for reviewing it ☺


VProum:
Hi Anaïs and Myriam!

Super nice job, Jasmine is all green! you used high order functions and fat arrow functions :D

Small improvements you can do : 
```
if (movieA.year < movieB.year) {
            return -1;
        } else if (movieA.year > movieB.year) {
            return 1;
        }

```
you can directly return movieA.year < movieB.year, as it is numbers, 
same goes for order aphabetically

Also, when dealing with simple IF statement, you can try to use ternary :
```
if (movieA.year < movieB.year) {
            return -1;
        } else if (movieA.year > movieB.year) {
            return 1;
        }


===

movieA.year < movieB.year ? -1 : 1
```


Maybe you could try some of the bonuses to challenge yourselves?

keep up the good work!

PR #1959, opened 2021-02-07
[PAR-WDFT-102] Maud Fabrello, Aurélie Lopez-Vicente & Mathieu Jacob

mathieujacob:
Hi,

We had difficulties to manage it but we did all the iterations and one bonus.

Thanks,

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,176 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+function getAllDirectors(movieList) {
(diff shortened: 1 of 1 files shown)


VProum:
Hi Maud and Aurelie and Mathieu J!

Nice job overall! you used high order functions, fat arrow functions and even ternary operator :D

Nice reuse of iteration 3 to make the iteration 4

I can see that you have jasmine test that you didn't pass.
Did you use jasmine to verify your code? I don't see that you renamed the specrunner file.

Iteration 4 is not working as intended, because you have to filter if a movie has a rate or not, instead of checking if the average is a number at the and of the computation.

Nice try with the bonus,
keep up the good work!

mathieujacob:
Hi,

Thank you for your help and great advice.
We were always verify on jasmine but sorry, I miss to push the index.html file on GitHub.
Iteration 4 : I will work on it after the pizza lab, thank you.

PR #1960, opened 2021-02-07
[PAR-WDFT- 102] Benjamin Pinard & Thaïs Hauguel

thaishauguel:


ta-web-paris:
Hey guys ! 
Great job on this lab, clean and smart !

A few points : 
- Steps 2 & 4: you can use "includes()" instead of "indexOf()" for an even cleaner code
- Steps 5 & 6 : you can compare strings by doing "a - b" to sort them alphabetically
- Global : try to organise your ternaries so they're easier to read, because right now i'm crying blood haha

```
condition
  ? if true
  : else
```

and if you have nested ternaries : 

```
condition 1
  ? if true
  : else condition 2
    ? if true
    : else
```

Best practice : when you'll have to write several lines in each ternary step, prefer not to use the ternary syntax ; i know it's cooler but it's very hard to read when you have to review code, so seniors might ask you to put regular if / else or even break your code with guards (condition && if true) so it's faster to read and more human-friendly

Otherwise very good job !


Nina

PR #1962, opened 2021-02-07
[PAR-WDFT-102] Pauline Bellaud, Anna Lefour, Antoine Stouff

Iron-Popi:
We couldn't finish the iterations 3 and 6 of the lab, but we did our best. :(


VProum:
Hi Pauline, Anna and Antoine!

Nice job overall! you used high order functions and callbacks :D
clean commits aslo!

regarding iteration 3 : 
your functions has to work even if some of the movies in the list doens't have rates, which means you have to check if a movie has a rate or not

Regarding iteration 6 : 
your sort method is not finished :(
you have :
`if (a.title < b.title) {
      return -1;
    } `  
but what if a.title > b.title ?
you can use split to limit the length of an array to 20

Small improvements/advices :
in a if statement, you don't need to explicitly compare a boolean to true :
`if (movies.genre.includes("Drama") === true) ` is equal to `if (movies.genre.includes("Drama"))`

When you feel comfortable, try to use fat arrow in your callback! you just have to remove the keyword "function" and put a " =>" between the parameters and the first curly brackets!

keep up the good work!

PR #1963, opened 2021-02-07
[PAR WDFT-102] Yassine Abouzaid & Adrien Mallie

ysabzd:
We had some difficulties trying to solve the iterations, we still cannot understand fully how to properly use .map, .filter, .sort functions. We would love to have reinforcement in these issues.


Fanny-Sbn:
Hello Adrien adn dYassine, thanks for submitting your lab. 

HOF are not easy at the beginning but you'll use them a lot and it will become easier. Not sure if the reinforcement is going to be on HOF or DOM manipulation yet but you can still look at funfunfunction (a youtube channel) and look for HOF explanations (it super well explained). 

Here are a few comments : 

Iteration 1: Ok 

Iteration 2: You're missing a part of the condition, you need to select only the drama movies and not all the movies (AND condition and includes function ???) 

Iteration 3: Almost there, what Jasmine tests are expecting from you : 

- if there is no movie return 0, in other words, if the length of the array is 0 return 0 (you can add this condition to your function) 
- if there is no movie rate you need to take this movie into account in the avarage anyway, in other words if there is no movie rate return 0 (add this conidtion in your map function

Iteration 4: Nice you reuse ratesAverage function

Iteration 5: Nice 

Iteration 6: You forgot to return only the title of the movies (you're returning all the information), try to first return all the titles of the film, then sort and slice. Keep in mind you can chain your HOF. 

You need to work again on the iterations you didn't pass all the Jamsine tests, try to do it whith the comments I've mde! 

Wax on wax, off! 

PS: Share the feedback with your partner ;)

PR #1964, opened 2021-02-07
PAR-WFDT-102-FJAM-RBENARD

Ooki-Dundun:
Hi all, please see our code!


ta-web-paris:
Hey guys ! 

### ⭐ Iteration 1 :

I don't understand why you wrote the line 14... If you have (director, i) then indexOf(director) will be "i" obviously haha because "i" IS the index of director, so you don't need that line the code works just fine without it ! You can just write

```
return arr.map(n => n.director);
```

For the bonus part, you can remove duplicates with several methods : filter, a forEach loop, or a method called "Set" in Javascript which can do the work for you. Check this link where everything is explained clearly :

### ⭐ Iteration 2 :

Good job, you can also use the method "includes()" instead of "indexOf" which is shorter and more global, check it here : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes

### ⭐ Iteration 3 :

Here you have an error with "NaN" (meaning "not a number" in Javascript) because at some point, one movie didn't have a rate, so it couldn't work with a numeric value for it, and you got a non numeric end value. You have to add a condition to guard this situation. You already placed one (if the array is empty, return 0) so it's the same logic : if the movie doesn't have a rating value, put "0" as a value ! That way you'll always have numbers to work with and won't have an error with "NaN".

The line where you wrote 
```
(!isNaN(obj.rate) || (obj.rate) !== undefined || typeof(obj.rate) !== "string")
```

you can simply check for NaN because undefined and string, are not numbers so it's the same
You than have to specify what to do in the "else" of that condition

### ⭐ Iteration 4 :

Nice !

### ⭐ Iteration 5 :

You went deep in the compare function, and you forgot to treat the case of the same year, which makes you have to sort them alphabetically ! You can make the code a little bit cleaner and complete by using something like this :
```
  if (a.year !== b.year) {
    return a.year - b.year;
  } else {
    return a.title - b.title;
  }
};
```

### ⭐ Iteration 6 :

Nice !

and good job on the bonuses as well !


If you have any question on this feedback, something you don't understand / disagree with, or whatever, please come find me on Slack and i'll be happy to answer ! I don't know if you both receive the email with the feedback, so please tell your partner about it so you both can check !

Nina

Ooki-Dundun:
Hi Nina, 
Iteration 1 - I believe line 14 is for the bonus. Is this not ok?
Iteration 3 - ok, is there another way? I find it quite arbitrary to say that no rate = 0. I've tried to add: else { return 0 } in the map function but I still have an error: expected 20 to be 2. Any advice?
Iteration 5 - noted, thanks.

PR #1965, opened 2021-02-07
[PAR - WBFT - 102] Juan Falconi / Fatoumata Diaby

JuanFalconi86:
Dear TA's:
We struggled a lot doing this excercice but were able to almost finish all 6 iterations (not the bonus...).

- Iteration 1 ==> we were happy to get Fanny's help that made us understand the syntax and how to logic?

- On iteration 3: I struggled using the .toFixed() function, because even if it rounded the result to 2 decimals, Jasmine was creating problems. With some help of some of the camarades, I used the parseFloat function and then it worked. 

- Also, in iteration 4, when I used the line "return movie.genre.includes("Drama")" I thought this was supposed to give a list of boolean results. Then I did another test and realized it gave actually a new array with only the drama movies, so it helped me to take the average. But still, I would like to understand because i thought that the .includes() function returned just boolean results.

- On iteration 6, were only able to order alphabetically the movies, but not to print the first 20 movies. I would love to see the answer in order to understand!

- For sure that even if was able to use it correctly, I still struggle understanding the .reduce() function

Thanks dear TA's!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
--- a/src/data.js
+++ b/src/data.js
@@ -1,2969 +1,2003 @@
 let movies = [
-    {
-      "title": "The Shawshank Redemption",
-      "year": 1994,
-      "director": "Frank Darabont",
-      "duration": "2h 22min",
-      "genre": [
-        "Crime",
-        "Drama"
-      ],
-      "rate": 9.3
-    },
-    {
-      "title": "The Godfather",
-      "year": 1972,
-      "director": "Francis Ford Coppola",
-      "duration": "2h 55min",
-      "genre": [
-        "Crime",
-        "Drama"
-      ],
-      "rate": 9.2
-    },
-    {
-      "title": "The Godfather: Part II",
-      "year": 1974,
-      "director": "Francis Ford Coppola",
-      "duration": "3h 22min",
-      "genre": [
-        "Crime",
-        "Drama"
-      ],
-      "rate": 9
-    },
-    {
-      "title": "The Dark Knight",
-      "year": 2008,
-      "director": "Christopher Nolan",
-      "duration": "2h 32min",
-      "genre": [
-        "Action",
-        "Crime",
-        "Drama",
-        "Thriller"
-      ],
-      "rate": 9
-    },
-    {
-      "title": "12 Angry Men",
-      "year": 1957,
-      "director": "Sidney Lumet",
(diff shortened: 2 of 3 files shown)


Fanny-Sbn:
Hello Fatoumata and Juan, thanks for submitting your lab. 

Here are a few comments: 

Iteration 1&2&5: Nice

Iteration 3: You forgot to check if there is a movie rate for each movie, you need to return 0 if there is no movie rate (add a condition in your reduce function)

Iteration 4: You should reuse your ratesAverage function ;) 
To answer your question, you're totally right movie.genre.includes("Drama") return true or false because it's a condition **BUT** you're in a filter function and the behavior of a filter function is following : a filter function accepts a condtion if the condtion is true the element you're iterating on is added to a new array, if the condition is false the element you're iterating on is not added to the new array. 
In this specific case when the movie is a drama one the condition is true so the the movie is added to the array ;)

Iteration 6: Actually, you're just missing a few steps, you need to only return the movie titles (you can use map) and then only take the first 20 movies (have a look to the slice method)
FYI, you can chain HOF ;) 

Wax on wax off!

PS: Share the feedback with your mate

Fanny

JuanFalconi86:
> Hello Fatoumata and Juan, thanks for submitting your lab.
> 
> Here are a few comments:
> 
> Iteration 1&2&5: Nice
> 
> Iteration 3: You forgot to check if there is a movie rate for each movie, you need to return 0 if there is no movie rate (add a condition in your reduce function)
> 
> Iteration 4: You should reuse your ratesAverage function ;)
> To answer your question, you're totally right movie.genre.includes("Drama") return true or false because it's a condition **BUT** you're in a filter function and the behavior of a filter function is following : a filter function accepts a condtion if the condtion is true the element you're iterating on is added to a new array, if the condition is false the element you're iterating on is not added to the new array.
> In this specific case when the movie is a drama one the condition is true so the the movie is added to the array ;)
> 
> Iteration 6: Actually, you're just missing a few steps, you need to only return the movie titles (you can use map) and then only take the first 20 movies (have a look to the slice method)
> FYI, you can chain HOF ;)
> 
> Try to modify some of your iterations with the comments I've made and push your changes again ;)
> 
> Wax on wax off!
> 
> PS: Share the feedback with your mate
> 
> Fanny


Thanks a lot @Fanny-Sbn !!! Really interesting the info about the .includes() behavior in filter function. Happy to know that!

Regarding iteration 6, you are right!!! I thought the .slice() only worked with strings, since those were the examples I saw in the prework! Now I know it can be used with more things!!! Thanks!

PR #1967, opened 2021-02-07
[PAR-WDFT-102] Boubacar SIDIBE & Augustin LEMAN

august159:
Lab completed


VProum:
Hi Boubacar and Augustin!

Super nice job, Jasmine is all green! you used high order functions, fat arrow functions and ternary operator :D

All the bonus is done too, i don't have any much comment to say :)
maybe you can optimise a little more better your code using ternary when dealing with simple if else statement : 

```
if (b.rate !== a.rate) {
        return b.rate - a.rate;
      } else {
        return a.year - b.year;
      } 
```
same as `return b.rate !== a.rate ? b.rate - a.rate : a.year - b.year`

```
movies2.map((toto) => {
    return toto.title;
  });
```
the return is not mandatory when you have a single line execution !

keep up the good job!

VProum:
I forgot, nice commits too! :D

PR #1968, opened 2021-02-08
[PAR-WDFT-102] Alexis Verthuy & Yuka TANAKA

seieeeen:
could not turned all into green...


ta-web-paris:
Hey guys ! 

### ⭐ Iteration 1 :

Good job - for the bonus part, you can remove duplicates with several methods : filter, a forEach loop, or a method called "Set" in Javascript which can do the work for you. Check this link where everything is explained clearly :

### ⭐ Iteration 2 :

Nicely done !

### ⭐ Iteration 3 :

Here you have an error with "NaN" (meaning "not a number" in Javascript) because at some point, one movie didn't have a rate, so it couldn't work with a numeric value for it, and you got a non numeric end value. You have to add a condition to guard this situation. You already placed one (if the array is empty, return 0) so it's the same logic : if the movie doesn't have a rating value, put "0" as a value ! That way you'll always have numbers to work with and won't have an error with "NaN".

You also used "Math.round()" to get 2 decimals in the rating, and that's working just fine, but there is one quick way to do it with the method toFixed(value) check this link to know how to use it : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

### ⭐ Iteration 4 :

Careful here : you call for "moviesList.genre.includes("Drama")" but moviesList containes movies, you have to use the method on a single movie, not on the array, because moviesList.genre doesn't exist ! (that's why you have "undefined" as an error") so a good way would be to filter your array to find movies including drama as a genre, that way you can call the includes method on each movie in the filter ! 

A smart way to work to not duplicate code would be to filter your array to only have the Drama movies, then call the function you created in the step before to get the average by passing the filtered version as a parameter 🧠 

### ⭐ Iteration 5 :

Here you have a confusing error : "expected [] not to be []" which means you have to create a copy to work on, because you don't want to modify the original array (which would be pretty messy if you worked on a real database !). When using methods like filter, or map, etc... the copy is automatically created, that's why you didn't get this error before now. It was the part of the instructions saying "do not mutate the original array"

You went deep in the compare function, and you forgot to treat the case of the same year, which makes you have to sort them alphabetically ! You can make the code a little bit cleaner and complete by using something like this :
```
  if (a.year !== b.year) {
    return a.year - b.year;
  } else {
  }
};
```
### ⭐ Iteration 6 :

Here the problem is you forgot to return only the title, and to return only 20 movies maximum. You also didn't work on the title you worked on the whole movie object, which is impossible to sort because it's an object, not a value ! You have to :
1 - sort by comparing the titles
2 - limit the result to 20 movies maximum
3 - return only the titles



And no worries about the bonuses at the end, you'll get the code in Guillaume's correction.
If you have any question on this feedback, something you don't understand / disagree with, or whatever, please come find me on Slack and i'll be happy to answer ! I suggest you try to really understand how you could have done (if you have time even try to work on it but it's not mandatory) because those are concepts that you will use in real situations !

I don't know if you both receive the email with the feedback, so please tell your partner about it so you both can check !

Nina

PR #1973, opened 2021-03-21
[PT-AMS-MAR21] Zsoka Montvai

ZsokaM:


ZsokaM:
Could not figure out the last Bonus, if we could look at it during class or in one of the lab "breaks" that would be great! thanks

PR #1993, opened 2021-03-25
[FTAMS032021] Jeroen, Sylvia, Jamal

jeroenvanbart:
Enough for today

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -90,4 +90,25 @@ console.log(orderAlphabetically(movies))
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 
+let calculateHoursToMin = (arr) =>{  
+    arr.forEach(element => {
+    return element.duration.replace(/h/g, '*60+') && (/min/g, '');
+    });
+}
+
+let turnHoursToMinutes = (arr) => {
+    let newArray= arr.map(arr => ({
+    "title": arr.title,
+    "year": arr.year,
+    "director": arr.director,
+    "duration": calculateHoursToMin(movies),
+    "genre": arr.genre,
+    "rate": arr.rate
+
+    }));
+    return newArray
+}
+
+
+
 // BONUS - Iteration 8: Best yearly rate average - Best yearly rate average
--- a/src/movies.js
+++ b/src/movies.js
@@ -3,8 +3,6 @@ const getAllDirectors = (movies) => {
     return movies.map(mappedArray => mappedArray.director); 
 };
 
-// getAllDirectors(movies);
-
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,114 @@
...
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+// const orderByYear = (array) => {
+//     if (array.length === 0){return [];}
+//     newOrder = array.sort((a,b)=> { 
+//     return a.year - b.year 
+// })
+// return newOrder
+// };
```
reMRKableDev, line 57: This block of commented code can be deleted as it serves no purpose to the rest of the code

src/movies.js
```diff
@@ -1,16 +1,114 @@
...
+    if (array.length === 0){return [];}
+    let orderByAlpha = array.sort(function(a, b) {
+      if (a.title === b.title) {return 0 }
+      else if (a.title < b.title){ return -1 }
+      else {return 1};
+      });
+    let newOrder = array.sort((a,b)=> { 
+    return a.year - b.year 
+  })
+  return newOrder
```
reMRKableDev, line 69: It would be good to make use of the `orderByAlpha` variable which you created in the first sort method. Then your function will return either the results of `newOrder` or `orderByAlpha`

```suggestion
  return newOrder || orderByAlpha
```

src/movies.js
```diff
@@ -1,16 +1,114 @@
 // Iteration 1: All directors? - Get the array of all directors.
+const getAllDirectors = (movies) => {
+    return movies.map(mappedArray => mappedArray.director); 
+};
+
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
+const set = new Set([...getAllDirectors(movies)])
+const cleanDirectorArray =[...set];
```
reMRKableDev, line 10: Move this into `getAllDirectors` function

src/movies.js
```diff
@@ -1,16 +1,114 @@
...
+console.log(orderAlphabetically(movies))
+
+
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 
+let calculateHoursToMin = (arr) =>{  
+    arr.forEach(element => {
+    return element.duration.replace(/h/g, '*60+') && (/min/g, '');
+    });
+}
```
reMRKableDev, line 97: The use of `forEach` here doesn't return any value. This will result in getting `undefined` back when called later.
You want to iterate but return some value back

reMRKableDev (commented):
Nice attempt @jeroenvanbart 🚀 

I've left some comments that can help you get passing tests. Check the comments

PR #2004, opened 2021-03-26
lab-greastest-movies

cintiagarcia:
Iteration 1- 6

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,96 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors (movies){
+    let mapped = movies.map(function(movie){
+        return movie.director;
+    });
+    return mapped; 
+}
+
+//console.log(getAllDirectors(movies))
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+function howManyMovies(movies){
+    const dramaMovies = movies.filter(function(dramaMovie){
+        if (dramaMovie.director === "Steven Spielberg" && dramaMovie.genre.includes("Drama")){
+            return true;
+        }
+    }) 
+    return dramaMovies.length
+}
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,97 @@
...
+
+function orderByYear(movies){
+  const mapped = movies.map(function(movie){
+    return [movie.title ,movie.year]
+  });
+  const sorted = mapped.sort(function(a,b){
+    return b-a
+  });
+  return sorted;
+}
```
ahmedhany82, line 78: The array `mapped` here will be an array of arrays. Each element will be an array of two elements title and year. This will get complex.

A simpler way is to just use `sort()`. In order to not mutate the original array, use `slice()` to make a copy of the array before sorting.

```
function orderByYear(array) {
  const sorted = array.slice().sort((a, b) => {
    if (a.year !== b.year) {
      return a.year - b.year;
    } else {
    }
  });
  return sorted;
}
```

Remember you need to sort alphabetically in case the years are equal. `localeCompare` can do that for you.

src/movies.js
```diff
@@ -1,16 +1,97 @@
...
 
+function orderAlphabetically(movies) {
+    const titlesMovie = movies.map(function(movie) {
+        return movie.title;
+    });
+    const titlesSorted = titlesMovie.sort(function(title1, title2){
+        return title2 - title1;
+    });
+    return titlesSorted.slice(0,20);
+}
```
ahmedhany82, line 92: Here we could use map first as you did because we only care about titles. So it is better if we get an array of just the titles.

For the sort, since we are sorting strings. I would use `localCompare` like so: `sort(function(a, b) { return a.localeCompare(b)) }`

src/movies.js
```diff
@@ -1,16 +1,97 @@
...
+
+
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 
 // BONUS - Iteration 8: Best yearly rate average - Best yearly rate average
```
ahmedhany82, line 97: Take a look at the bonus part in the lab solution and if you have any questions, feel free to reach out.

ahmedhany82 (commented):
Hey Cintia, good job for this lab! Please find my review comments. Feel free to reach out in case you have any questions. Regards!

PR #2005, opened 2021-03-26
[BER-FT-032021] Hélène Trommelen

helenetrommelen:
Progress on iteration 1-5.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -40,15 +40,51 @@ function howManyMovies(movies){
     return filteredMovies.length;
 } 
 
-
-
-
-
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
+ function ratesAverage(movies){
+     if (movies.length === 0){return 0
+    } else {
+        let avg = movies.reduce(function(acc, val){
+            //conditioanl statement, if not return acc.
+            if (val ===''){
+                return 0;
+            } else {
+                return acc + val.rate;
+            }
+        },0)/movies.length;
+        return parseFloat(avg.toFixed(2));
+    }
+}
+
+console.log(ratesAverage(movies))
+
+
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(movies){
+    let dramaMovies = movies.filter(function(movie){
+        if(movie.genre.includes('Drama')){
+            return true
+        }
+    })
+    return ratesAverage(dramaMovies);
+}
+
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
+function orderByYear(movies){
+    return movies.map(movie => movies.sort(function (a,b){
+        return a.year-b.year
+    }));
+    
+    //return moviesByYear
+}
+
+// const moviesByYear = movies.sort(function (a,b){
+//     return a.year-b.year
+// })
+
+
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,7 +1,48 @@
 // Iteration 1: All directors? - Get the array of all directors.
+//console.log(movies);
+
+// function getAllDirectors(movies){
+//     return movies.map(function (movie)
+//     {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,13 +1,90 @@
...
+//     return movies.map(function (movie)
+//     {
+//     return movie.director;
+// })
+// }
+
+//easier ES6 notation
+function getAllDirectors(movies){
+    return movies.map(movie => movie.director);
+}
```
ahmedhany82, line 14: This is an arrow function expression. We will cover this soon :)

src/movies.js
```diff
@@ -1,13 +1,90 @@
...
+//     console.log(howManyMovies(movies));
+function howManyMovies(movies){
+
+    let filteredMovies = movies.filter(function(movie){
+        if ((movie.director === 'Steven Spielberg') && (movie.genre.includes('Drama'))){
+            return true}
+            
+    })
+    return filteredMovies.length;
+}
```
ahmedhany82, line 41: Filter expects a function that returns true or false, so I would explicitly return false as well. You could either add an else to the "if" and return false or you could directly return the expression since it evaluates to either true or false.

```
    let filteredMovies = movies.filter(function(movie) {
        return ((movie.director === 'Steven Spielberg') && (movie.genre.includes('Drama')));  
    })
```

src/movies.js
```diff
@@ -1,13 +1,90 @@
...
+            //conditioanl statement, if not return acc.
+            if (val ===''){
+                return 0;
+            } else {
+                return acc + val.rate;
+            }
+        },0)/movies.length;
+        return parseFloat(avg.toFixed(2));
+    }
+}
```
ahmedhany82, line 57: "val" here is a movie object, it could be totally empty or it could be only missing the rate. If you want to check whether the rate field is empty, you could do val.rate === "" but a better way that would work even if the whole movie is empty is to check `if (!val.rate) `. It is like checking if val.rate does not exist. In this case, you would still need to return the `acc` but without adding anything to it.

 ```
       let avg = movies.reduce(function(acc, val){
            //conditional statement, if not return acc.
            if (!val.rate){
                return acc;
            } else {
                return acc + val.rate;
            }
        },0)/movies.length;

src/movies.js
```diff
@@ -1,13 +1,90 @@
...
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
+function dramaMoviesRate(movies){
+    let dramaMovies = movies.filter(function(movie){
+        if(movie.genre.includes('Drama')){
+            return true
+        }
+    })
+    return ratesAverage(dramaMovies);
+}
```
ahmedhany82, line 71: Similar to the filter function above. I would add an else or return the boolean expression.

src/movies.js
```diff
@@ -1,13 +1,90 @@
...
+    return movies.map(movie => movies.sort(function (a,b){
+        return a.year-b.year
+    }));
+    
+    //return moviesByYear
+}
+
+// const moviesByYear = movies.sort(function (a,b){
+//     return a.year-b.year
+// })
```
ahmedhany82, line 85: Here using `map()` will not be so useful. You could only use `sort()` but in order not to mutate the original array, you could first copy it using `slice()`. You need also to handle the case where the years are equal and sort in alphabetical order in this case. `localeCompare` can do that for you.

```
function orderByYear(array) {
    if (a.year !== b.year) {
      return a.year - b.year;
    } else {
    }
  });
  return sortedArray;
}
```

src/movies.js
```diff
@@ -1,13 +1,90 @@
...
+
+
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
```
ahmedhany82, line 89: Give this one a try. Here you could use map() to get an array with just the titles, then sort this array and use slice to get the first 20 titles.

ahmedhany82 (commented):
Hi Hélène, good job with this lab. Please find my review comments. Feel free to reach out in case you have any questions. Regards!

PR #2006, opened 2021-03-26
Berlin - WDFT MAR2021 - Alexander Rathbone 

atrathbone:
Completed tasks. 

Bonus Material: 
Cannot work out why failing test for duration to be a number.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -13,7 +13,7 @@ function getAllDirectors(movies) {
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
 function howManyMovies(movies) {
     const spielbergDramas = movies.filter(function (movie) {
-        if (movie.director === 'Steven Spielberg' && movie.genre === 'Drama') return true;
+        if (movie.director === 'Steven Spielberg' && movie.genre.includes('Drama')) return true;
     })
     return spielbergDramas.length;
 }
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,132 @@
...
+function turnHoursToMinutes(movies){
+    let moviesCopy = [...movies];
+    for (let i=0; i<moviesCopy.length; i++){
+     
+      let justNumbers = moviesCopy[i].duration.replace(/\D/g,'');
+      let separated = justNumbers.slice(' ');
+      moviesCopy[i].duration = (separated[0]*60) + (parseInt(separated[1])*10) + (parseInt(separated[2])); 
+    }
+    return moviesCopy; 
+  }
```
ahmedhany82, line 97: I think there is an issue here with the logic in line 94. Try to console.log `separated` and you will find that sometimes it is a 1 digit, sometimes 2 and other times 3 digits. So trying to access separated[1] or separated[2] will give you undefined. 

Also if you click on the test case in Jasmine, you will come across this error `TypeError: moviesCopy[i].duration.replace is not a function`. I would say try looking into the point above first and then we can discuss it.

ahmedhany82 (commented):
Hey Alexander, good job with this lab. Please find my review comments. Feel free to reach out if you have any questions or to discuss the bonus iteration. Regards!

PR #2007, opened 2021-03-26
[BER-FT-032021] Philipp Dam

Philipphub:


src/movies.js
```diff
@@ -1,16 +1,68 @@
...
+  return arr.filter(
+    (item) =>
+      item.director === "Steven Spielberg" && item.genre.includes("Drama")
+  ).length;
+}
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
+function ratesAverage(arr) {
+  const arrRate = arr.map((item) => item.rate);
```
annawaszak, line 19: When you map over an array which does not have a rate - it returns - undefined. That's why when you reduce it later it returns NaN.

src/movies.js
```diff
@@ -1,16 +1,68 @@
...
+  const arrAvg = arrRate.reduce((a, b) => a + b / arrRate.length, 0);
+  return Math.round(arrAvg * 100) / 100;
+}
+
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate(arr) {
+  let arrDrama = arr.filter((item) => item.genre.includes("Drama"));
+  let arrRate = arrDrama.map((item) => item.rate);
+  let arrAvg = arrRate.reduce((a, b) => a + b / arrDrama.length, 0);
```
annawaszak, line 29: To make it even more efficient, you could skip those two lines and instead make use of the ratesAverage() function you created in iteration 3.

src/movies.js
```diff
@@ -1,16 +1,68 @@
...
+  return Math.round(arrAvg * 100) / 100;
+}
+
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
 
+function orderByYear(arr) {
+  let arrYear = arr.sort((a, b) => {
```
annawaszak, line 36: The sorted method, unlike map or filter, sorts the elements of an array **in place** and returns the sorted array. In order to pass the tests, you need to return a **new** array.

src/movies.js
```diff
@@ -1,16 +1,68 @@
...
+}
+
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 
+function turnHoursToMinutes(arr) {
+  let newArray = [...arr];
+  for (let i = 0; i < newArray.length; i++) {
+    let duration = newArray[i].duration;
+    let hour = Number(duration.split(" ")[0].replace("h", ""));
+    let minutes = Number(duration.split(" ")[1].replace("min", ""));
```
annawaszak, line 62: It does not work because some movies could have only hours, others could have only minutes, so need a way to figure out what each duration contains and then proceed based on this.

annawaszak (commented):
Hi Philipp,

Great work on this lab! You clearly have a good understanding of the topics covered in week 1. In the comments below, you will find a few suggestions on how to improve your code even more. If you have any questions, don't hesitate to let me know.

Thanks,
Anna

PR #2010, opened 2021-03-26
[Berlin-WDFT-MAR2021] Marcia Faria Netter

mfndnc:
for BONUS - Iteration 8,  I created 2 solutions.
one rather convoluted called function bestYearAvgConvoluted
and one that uses more Array methods called function bestYearAvg

is the later really better in term or readability or speed or neither?

any comments would help me. thanks in advance

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "liveServer.settings.port": 5501
-}
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,154 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+function getAllDirectors(movies) {
+  let allDirectors = movies.map(function (cur) {
+    return cur.director;
+  });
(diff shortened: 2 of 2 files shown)


annawaszak (commented):
Hi Marcia,

Yet again, fantastic job! Not only you have done it all iterations including bonuses, but you also did it remarkably quickly. Your code is readable and very efficient, which I really admire. 

In terms of the last bonus, the first option does seem a bit easier to understand, but both are well executed. 

I can share with you how I solved this one, I think it's equally convoluted. 

```
const bestYearAvg = (movies) => {
    let ratingsByYear = {};
  
    movies.forEach(movie => {
        const existingYear = ratingsByYear[movie.year];
        if (!existingYear) {
            ratingsByYear[movie.year] = [movie.rate]
        }
        else ratingsByYear[movie.year].push(movie.rate)
    })  
  
    for (let year in ratingsByYear) {
      ratingsByYear[year] = ratingsByYear[year].reduce((acc, rate) => acc + rate, 0 ) / ratingsByYear[year].length;
    }
      
    let highestRate = Math.max.apply(null, Object.keys(ratingsByYear).map(x => ratingsByYear[x]))
  
    let yearWithHighestRate = Object.keys(ratingsByYear).reduce((result, key) => { 
      if (ratingsByYear[key] === highestRate){ 
        result.push(key); 
      } 
      return result; 
    }, []);
  
    return `The best year was ${highestRate} with an average rate of ${yearWithHighestRate}`
  }
```

As always, I remain available if you had any questions or suggestions!

Thanks,
Anna

mfndnc:
Thanks for sharing your version. I really appreciate learning how coding can be used in so many different ways.

I am going to paste it into the VS code to see if I understand the last part let yearWithHighestRat ... .keys .reduce

thanks again
Marcia

PR #2012, opened 2021-03-26
BER-WDFT2021-R.Stein

Chocogrenouille:


src/movies.js
```diff
@@ -1,16 +1,113 @@
...
 // Iteration 4: Drama movies - Get the average of Drama Movies
 
+function dramaMoviesRate (movies) {
+    const filterDramaMovies = movies.filter(function (movie) {
+        if (movie.genre.includes('Drama')){
+            return true;
+        } else {
+            return false
+        }
+    })
```
ahmedhany82, line 52: You got the filter right this time. You could directly return the boolean expression.

`const filterDramaMovies = movies.filter(function (movie) {
        return (movie.genre.includes('Drama')) })`

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+        if (a.title < b.title){
+          return -1
+        } else if (a.title > b.title) {
+          return 1
+        }
+      }
+  });
+    return sortedYear;
+}
+
```
ahmedhany82, line 82: You could use slice to copy the array before sorting. Another way to sort alphabetically is using localeCompare.

The shorter version would be :
```
function orderByYear(array) {
    if (a.year !== b.year) {
      return a.year - b.year;
    } else {
    }
  });
  return sortedArray;
}
```

src/movies.js
```diff
@@ -1,16 +1,113 @@
...
+      const alphabeticalTitles = movieTitlesOnly.sort(function (a,b) {
+          if (a < b) {
+            return -1
+          } else if (a > b) {
+            return 1
+          }
+      })
+
+    return alphabeticalTitles.slice(0,20);
+  }
```
ahmedhany82, line 106: Map always returns a new array so here there is no need to create a new array your own.

You could use map to return the titles of the movies in a new array like so:
```
      const onlyTitles = newArray.map(function (movie) {
        return movie.title;
      })
```
      
Then you can run sort on this onlyTitles array.

ahmedhany82 (commented):
Hey Roxana, good job with the greatest-movies lab. Please find my comments on your code. Feel free to reach out in case you have any questions.

PR #2018, opened 2021-03-26
Berlin-WDFT-Bryan Sanderson

weltenbuerger:


src/movies.js
```diff
@@ -1,16 +1,143 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(movies) {
+  return movies.map((movie) => movie.director)
```
wax-bear, line 4: Awesome Javascript.
A semicolon would not hurt at the end of the line.

wax-bear (commented):
Great stuff Bryan!

A few general remarks:
- a let is used when a variable will be reassigned at some point. Some lets in your code will not be reassigned, so could be turned into consts.
- There is discussion in the JS community about the use of a semicolon after each declaration. I like non, but seem to remember a source that advises to use them, for readability.

Best,

Thomas

PR #2019, opened 2021-03-26
[FTAMS032021] Igor Godinho Figueiredo Movies of all time

Igor-GF:
In iterations 5 and 6 I based on Andrea's code to try to understand better the logic and relations but even so I could not succeed yet, according to Jasmine.


reMRKableDev (commented):
Nice attempt @Igor-GF 🚀 

I've left some comments for you to check out

PR #2024, opened 2021-03-28
BER-WDFT-MAR2021 - Alkin 

SirDev97:
I have some difficulties with this lab, thats why it took long to submit. Would love it if we can do a recap of this.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,64 @@
 // Iteration 1: All directors? - Get the array of all directors.
+function getAllDirectors(arr) {
+    const result = arr.map(function (movie) {
+        return movie.director;
+    })
+    return result;
+}
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(arr) {
+    let counter = 0;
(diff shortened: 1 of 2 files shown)


src/movies.js
```diff
@@ -1,16 +1,77 @@
...
+        return acc + val.rate;
+    }, 0)
+    return Number((dramaAvgRate / dramaMovies.length).toFixed(2));
+}
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-
+function orderByYear(arr) {
+    const sorted = arr.sort(function (a, b) {
+        if (a.year === b.year) {
+            return a.title.length - b.title.length;
```
annawaszak, line 57: You want to return a string that has a higher alphabetical order, it does have to do anything with the length of that word. When sorting strings you should return "1" (b is less than a), "0" (both are equal) or "-1" (a is less than b). You don't need to do it manually, you could use the localCompare() method.

src/movies.js
```diff
@@ -1,16 +1,77 @@
...
+        }
+    }).slice().sort(function (a, b) {
+        return a.year - b.year;
+    })
```
annawaszak, line 61: You don't need the slice() and another sort(). You can put all sorting conditions inside the first one. Just ass line 60 as an "else".

annawaszak (commented):
Hi Alkin,

Great job on the lab. I can tell you have an understanding of what was covered in the lectures last week. The only problem was related to sorting strings.

I left all by comments below. If you have any questions, just let me know, I'm happy to assist.

Thanks,
Anna

PR #2026, opened 2021-03-28
[BER-WDFT-032021] Movies - Pietro Ballarin

pietroballarin:
I still can't solve the issue of an empty array without an initial value not returning 0 and how to make it work with the reduce function

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,15 +1,78 @@
-// Iteration 1: All directors? - Get the array of all directors.
-// _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
+
+
+// Iteration 1: All directors? - Get the moviesay of all directors.
+
+function getAllDirectors(movies){ 
+  const allDirectors = movies.map(function(mov){
+    return mov.director;
+  })
+  return allDirectors;
+}
+
+// _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the moviesay of directors. How could you "clean" a bit this moviesay and make it unified (without duplicates)?
+
+function findUniqueDirectors(movies) {
+  const uniqueDirectors = [...new Set(getAllDirectors(movies))];
+  return uniqueDirectors ;
+}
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,79 @@
...
+  return numberOfDramas.length;  
+}
+
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
 
-// Iteration 4: Drama movies - Get the average of Drama Movies
+function ratesAverage(movies){
+   
+  const sumRating = movies.reduce(function(acc, val){
+    if (movies.length === 0) return 0;
```
annawaszak, line 33: I would suggest to move this part outside the reduce method - to the beginning of the function.

src/movies.js
```diff
@@ -1,16 +1,79 @@
...
+    if (val.rate !== '' && 'rate' in val) {
```
annawaszak, line 34: This if statement should check if a rating **doesn't exist** within the specific movie ("val") object. You could achieve this by simply adding `!val.rate` as the conditional. It's the same as `val.rate === false`. If it doesn't exist, it should just return the sum ("acc"), otherwise add the value of the rating to the sum.

src/movies.js
```diff
@@ -1,16 +1,79 @@
...
+    return movies.genre.includes('Drama');
+  }); 
+  const dramaRates = numberOfDramas.map(function(movies){
+    return movies.rate;
+  }); 
+  const sumOfRates = dramaRates.reduce(function(acc, val){
+  return acc + val;
+  }); 
+  const avg = sumOfRates / dramaRates.length;
+  return Number(avg.toFixed(2));
```
annawaszak, line 54: You could save yourself some code by applying the ratesAverage() here instead or rewriting it's content. If you decide to use this solution, then the other error goes away, because that function already returns 0 when the array is empty. If you decided to keep the code, then you will need to add that additional line.

annawaszak (commented):
Hi Pietro,

Great work on this lab! You clearly have a good understanding of the topics covered in week 1. In the comments below, you will find a few tips and suggestions. If you have any questions, don't hesitate to let me know.

Thanks,
Anna

PR #2044, opened 2021-03-30
BER WDFT MAR2021 - Olesia Kulykova

olakul:


src/movies.js
```diff
@@ -1,15 +1,58 @@
+
+
 // Iteration 1: All directors? - Get the array of all directors.
+const getAllDirectors = movies.map(function(movie){
```
wax-bear, line 4: I would add a space:
`function(movie) {`

src/movies.js
```diff
@@ -1,15 +1,58 @@
...
+    return(movie.director)
+  })
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors. How could you "clean" a bit this array and make it unified (without duplicates)?
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-
+function howManyMovies(movies) {
+    const dramaMovies = movies.filter(function (movie) {
+      if (movie.genre.includes("Drama") === true && movie.director === "Steven Spielberg") {
```
wax-bear, line 12: `movie.genre.includes("Drama")` returns true or false so `=== true` is not necessary.

src/movies.js
```diff
@@ -1,15 +1,58 @@
...
+        return true;
```
wax-bear, line 13: Instead of returning inside the if statement, the condition itself can be returned.

src/movies.js
```diff
@@ -1,15 +1,58 @@
...
+      }
+    });
+    return dramaMovies.length;
+}
 // Iteration 3: All rates average - Get the average of all rates with 2 decimals
-
+function ratesAverage(movies){
+    const sumRates = movies.reduce(function(acc,val){
+      return acc + val.rate
```
wax-bear, line 21: What happens if `val.rate` is not defined? :)

src/movies.js
```diff
@@ -1,15 +1,58 @@
...
+    },0)
+    return +(sumRates / movies.length).toFixed(2)
+    
+  } 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-
+function dramaMoviesRate(movies){
+    let dramas = movies.filter(function(movie) {
```
wax-bear, line 28: Both variables dramas and avg will not be reassigned, so const instead of let can be used.

wax-bear (commented):
Great stuff Ola!

Please check my comments for potential improvements. In general, try to be mindful about semicolons and spacing. Finally, iteration 6 is sorting by year, but not by name when the year is equal, yet.

Please reach out if you have a question.

Best,

Thomas

PR #2115, opened 2021-06-03
Bart Spangenberg: Lab day 4

BartSpangenberg:
That's was quite a challenge.


TA-Remote:
Hey Bart! amazing work again! Good commenting! Very good logic for the bonuses, and I see that you use also the rest operator! Keep going!

PR #2118, opened 2021-06-03
Liliia - WDFT - RMT - MAY2021 Week1Day4

liliiakhr:
Lab submission


TA-Remote:
Very very good job on the lab! Feel free to check the solution to see how you could have completed the last bonus. But very nice work, take rest this weekend :)

PR #2120, opened 2021-06-03
Mariana WDFT-RMT-MAY21

marileitune:


TA-Remote:
Very nice lab :) Try to go over the solution to check how you could have done the bonuses and save some lines of code in the mandatory iterations. But nice job, rest well this weekend :)

PR #2121, opened 2021-06-03
[WDFT-RMT-MAY21] Conchita 

CodeChita:
Hey TA's

It was more confusing because the console was not being my friend :( So this is the monster code I produced. If I have enough energy I will look at  it in the morning to make it more efficient. I would love to get some tips where I can make my code shorter and more efficient. 

Have a lovely day!!


TA-Remote:
Well done Conchita!! You did all the mandatory iterations except for the bonus. Im going to give you a few improvement points 👍 
- On iteration 4 you could have tried to avoid the .map because with the reduce method you can also access to the property score of every element to add to the accumulator.
- I think you did really good. Do not worry too much for now to make your code shorter, that will come out little by little 🥇 

PR #2124, opened 2021-06-03
Halil Ibrahim WDFT-RMTMAY21

halokaya67:


TA-Remote:
Good job Halil! You did all the mandatory iterations and also tried the bonus. Im going to give you a few improvement points 👍 
- You should have tried to avoid using .forEach but I'm very happy to see all tests passing (expect last bonus).

PR #2129, opened 2021-06-04
RITA MAK WDRT MAY 2021

ritamak:


TA-Remote:
Well done Rita! You did all the mandatory iterations except for the bonus. Im going to give you a few improvement points 👍 
- To round the answer to 2 decimals you can use the X.toFixed(2) method but keep in mind it returns it as a string so you also have to convert it to a number with the Number(X).
- Try to read the test cases when they fail becuase the give you the information to make them pass. 

PR #2130, opened 2021-06-04
EMMY RMTFT MAY 2021

emmyacosta:


TA-Remote:
Hey Emmy! We've tried to check you lab and the test cases but is not working because the lab has a lot of sintax errors on iteration 3. Look at the problems tab on VSCode and try to fix them.
Resubmit when you solve that so we can review you lab. 
Thanks!

PR #2136, opened 2021-06-04
BER-WDFT-052021 - Andrew Oppo

andrewoppo:
Not finished yet. Still need to troubleshoot. It just takes so much longer having to run the npm test after every single change, as my my live view doesn't work

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,77 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
+
+const movies = require("./data");
+
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(arr) {
+  const directorArray = arr.map(function(movie) {
+    return movie.director
+  });
+  return directorArray;
+}
+
+console.log(getAllDirectors(movies))
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(arr) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,77 @@
...
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+function orderAlphabetically(arr) {
+  const newArr = arr.slice();
+  const alphMovies = newArr.sort( function(a, b) {
+    for (i=0; i < 20; i++){
+      return a.title - b.title;
+    }
+  });
+    return alphMovies.title;
```
atkat, line 73: You are currently running a loop in your sort, which is not possible as sort is already a kind of loop. I mean, it is technically possible but more in the case when you first do a for loop and then for example a map. Not with sort. If you want to do it this way, you'd need to sort first and then just loop and print:
```
let alphMovies = newArr.sort( function(a, b) {  return a.localeCompare(b) }
let resultArr = [];
for (i=0; i < 20; i++) { 
  resultArr(movies[i].title) 
}
return resultArr
```
BUT

the most efficient way to do it would be the way Jan showed you guys: 

```
  const ordered = movies.map(function (movie) {
    return movie.title
  }).sort(function (a, b) {
    return a.localeCompare(b);
  }).slice(0, 20);

  return ordered;
}
```

Also, on line 73 in your original code- alphMovies is supposed to be an array not just one of the movie objects - we can't access the title before mapping through it.

atkat (commented):
Hey Andrew, 

I've made quite a few comments, but don't worry I know you understand the concepts. Maybe just some of the array methods were still not 100% cemented when you were doing it.  I think you will be able to work out a lot of it on your own once you start to understand the errors/output a bit more.  Can you have a look through my comments and see if everything makes sense?

Best,
Katka

PR #2140, opened 2021-06-04
BER - WDFT - MAY21 - Alan Sanchez

alnwolff:
Finished all the mandatory iterations.
The Bonus iteration are not completed yet.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,23 +1,93 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(arr) {
+  let directors = []
+
+  arr.map (function (movie) {
+    
+    directors.push (movie.director)
+    return directors
+  })
+  
+  return directors
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(arr) {
+  
+  const moviesCount = arr.filter (function (movie) {
+    return movie.director === 'Steven Spielberg' && movie.genre.indexOf('Drama') != -1
+  })
+  return moviesCount.length
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(arr) {
 
+  if (arr.length === 0) {
+    return 0
+  }
+  
+  const sum = arr.reduce (function (acc, val) {
+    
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,23 +1,93 @@
...
+  const onlyDrama = arr.filter (function (movie) {
+    return movie.genre.indexOf('Drama') != -1
+  })
+  return scoresAverage(onlyDrama)
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear(arr) {
+  const copy = JSON.parse(JSON.stringify(arr));
```
atkat, line 51: I can see that you understand the logic of what needs to be done. Have a look at this, it's a little shorter and more efficient: 

```
  const sorted = movie .sort(function (a, b) {
    if (a.year !== b.year) {    //checks if they are different
      return a.year - b.year;   //sorts
    } else {
      return a.title.localeCompare(b.title);   //uses localeCompare to sort the ones with the same year
    }
  });
  return sorted;
}
```

src/movies.js
```diff
@@ -1,23 +1,93 @@
...
+    }
+  })
+  return orderedYear
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+function orderAlphabetically(arr) {
 
+  const copy = JSON.parse(JSON.stringify(arr));
```
atkat, line 73: you can also use arrr.slice() to make a copy of the array

src/movies.js
```diff
@@ -1,23 +1,93 @@
...
+    
+  const orderedTitles = copy.sort (function (a,b) {
+    if (a.title > b.title) {
+        return 1
+      } else if (a.title < b.title) {
+        return -1
+      }
+  })
```
atkat, line 81: have a look at the localeCompare function, it's very useful in this case and would save you three lines

src/movies.js
```diff
@@ -1,23 +1,93 @@
...
+  
+  const first20Titles = orderedTitles.slice(0,20)
+
+  const movies = first20Titles.map( function (movie) {
+    
+    return movie.title
+  })
+  return movies
+}
```
atkat, line 90: You could maybe try chaining the methods. Your code works, but you don't need to save everything into a variable, it's less writing for us and also more efficient: 
```
  const sorted = movies.slice().sort(function (a, b) {
    if (a.year !== b.year) {
      return a.year - b.year;
    } else {
    }
  });
  return sorted;
}
```

atkat (commented):
Hey Alan! 

Well, done! I can see you understand the concepts :) My feedback is mostly about writing shorter, more efficient code :)

Let me know if you have any questions.  Oh and you could have a go at the bonuses, what do you think? ;)

Best, Katka

PR #2144, opened 2021-06-04
 [BER-WDFT-MAY2021] <Irina Falko>

IraAwesomePossum:


src/movies.js
```diff
@@ -1,31 +1,69 @@
...
+  let year = arr
+    .map(function (i) {
+      return {
+        title: i.title,
+        year: i.year
+      };
+    })
+    .sort((a, b) => a.year - b.year || a.title.localeCompare(b.title));
+  return year;
+}
```
atkat, line 52: I assume you're doing the map because you don't want to mutate the original array. I'd say it's probably more efficient to use arr.slice() or even [...arr]. 

Well done on the sort syntax!

src/movies.js
```diff
@@ -1,31 +1,69 @@
...
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+function orderAlphabetically(array) {
+  let titles = array.map((i) => i.title).sort();
+  let sliced = titles.slice(0, 20);
+  return sliced;
+}
```
atkat, line 59: For some reason this is passing the tests, but in reality the sort function does not work very well on strings. That's why we use the localeCompare. 
https://fuzzytolerance.info/blog/2019/07/19/The-better-way-to-do-natural-sort-in-JavaScript/

atkat (commented):
Hey Irina,

Nice work, again! I only have a few comments. 

However, I bet you could do the bonuses too. How come you didn't? Too many things to do?

Best,
Katka

PR #2145, opened 2021-06-04
[BER-WDFT-MAY2021] Alexandre Oliveira

alexandreoliv:
Last bonus can definitely be refactored

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,28 +1,190 @@
+const movies = require("./data");
+
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(movies) {
+  let directorsArray = movies.map(function(movie){
+    return movie.director;
+  });
+  return directorsArray;
+}
+directorsArray = getAllDirectors(movies);
+// console.log(directorsArray.length)
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,28 +1,190 @@
...
+  let entries = [];
+  for (let i = 0; i < uniqueYearsArray.length; i++) {
+    for (let j = 0; j < yearsAndScoresArray.length; j++) {
+      if (yearsAndScoresArray[j].year === uniqueYearsArray[i]) {
+        uniqueYearsArrayWithRatings[i] = (uniqueYearsArrayWithRatings[i] || 0) + yearsAndScoresArray[j].score;
+        entries[i] = (entries[i] || 0) + 1;
+      }
+    }
+  }
+
```
atkat, line 159: You are using a lot of helper variable and arrays. As a result, it's a bit hard to read what you are trying to do and it becomes quite long. You can often just map or filter or reduce through the original object (using the dot notation to retrieve the things you need) and then doing something with them. You can also chain methods (arr.map.filter). And you don't need to clone the original array if you're using the methods that return a new array.

I don't really want to give you a solution, Jan already gave one in the review I believe. There a few, as always. I can, but maybe you can try to rewrite it first, make it more efficient and more concise, and then I can give you a couple of solutions?

atkat (commented):
Hey Alexandro,

Really well done as always! Just the last bonus is not efficient and would probably not pass if it was an online challenge or a real life functionality. Please see my individual comments.

Great stuff though!!! :)

PR #2147, opened 2021-06-06
BER-WDFT-MAY2021-Joanna

joannaberlin:
done without bonuses and not all tests passed

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,90 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(movies) {
+  const directors = movies.map(function (all) 
+  {return all.director
+  }); return directors
+}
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,90 @@
...
+  // const sorted = first20.sort()
+  //   return sorted
+  const sorted = first20.sort(function(a, b) {
+    return a.title.localeCompare(b.title)
+  }) 
+  const onlyTitles = sorted.map(function (titl) {
+    return titl.title
+    })
+  return onlyTitles
+}
```
ahmedhany82, line 87: Here it is better to `sort` first and then `slice` at the end. We need to first to sort the whole list and then take the first 20 movie titles. It is also better we do the map first before the sort because this simplifies the array to be sorted (array having titles instead of array having movies).

I modified your code a little:
```
 
  const onlyTitles = movies.map(function (movie) {
    return movie.title
    })
  const sorted = onlyTitles.sort(function(a, b) {
    return a.localeCompare(b)
  }) 
  return sorted.slice(0,20)
} 
```

Check the solution on github for a shorter version

ahmedhany82 (commented):
Hey Joanna,
Good work for the lab. Please find my comments below. Feel free to reach out if you have any questions.
Best regards,
Ahmed

PR #2148, opened 2021-06-06
Ber-WDFT-May2021-Zhanna

Decdancer:
could not get tests passed, though iterations do work in codepen

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,85 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
+
+const movies = require("./data")
+
+function getAllDirectors() {
+  const directors = movies.map(function (film) {
+    return film.director
+    })
+    const directorsUniq = directors.reduce(function(a,b){
+      if (a.indexOf(b) < 0 ) a.push(b)
+      return a
+    },[])
+  return directorsUniq
+  }
+
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies() {
+  const dramas = movies.filter(function (film) {
+    return film.genre.indexOf('Drama') !== -1
+      })
+  const dramasOfSpielberg = dramas.filter(function (film) {
+    return film.director === 'Steven Spielberg'
+  })
+  return dramasOfSpielberg.length
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage() {
+  const sumOfScores = movies.reduce(function (acc, film) { 
+    return acc + film.score
+  }, 0)
+  return (sumOfScores / movies.length).toFixed(2)
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore() {}
+function dramaMoviesScore() {
+  const dramas = movies.filter(function (film) {
+    return film.genre.indexOf('Drama') !== -1
+  })
+    const scoreDramas = dramas.reduce(function (acc, film) { 
+      return acc + film.score
+    }, 0)
+  return (scoreDramas / movies.length).toFixed(2)
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear() {
+  const moviesSorted = movies.slice()
+  moviesSorted.sort(function(a, b) {
+    const titleA = a.title.toUpperCase()
+    const titleB = b.title.toUpperCase()
+    if (titleA < titleB) {
+      return -1
+    }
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,99 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
+
+const movies = require("./data")
+
+function getAllDirectors() {
```
ahmedhany82, line 6: One common comment on all the functions is that you are missing the argument of the function. This is why some of the test cases are failing. The test cases call your function and pass an array to it. However, the function is not accepting any inputs. Try instead:
`function getAllDirectors(movies) {`

src/movies.js
```diff
@@ -1,22 +1,99 @@
...
+function getAllDirectorsUniq() {
+  const directors = movies.map(function (film) {
+    return film.director
+    }).reduce(function(a,b){
+      if (a.indexOf(b) < 0 ) a.push(b)
+      return a
+    },[])
+  return directors
+  }
+console.log(getAllDirectorsUniq(movies).length) // => 149
```
ahmedhany82, line 24: It is easier if you do it with `filter` or `set`. Check the solution posted by Jan.

src/movies.js
```diff
@@ -1,22 +1,99 @@
...
-function howManyMovies() {}
+function howManyMovies() {
+  const dramas = movies.filter(function (film) {
+    return film.genre.indexOf('Drama') !== -1
+      })
+  const dramasOfSpielberg = dramas.filter(function (film) {
+    return film.director === 'Steven Spielberg'
+  })
+  return dramasOfSpielberg.length
+}
```
ahmedhany82, line 36: You could combine that into one filter function using the `&&` operator.

```
if (film.director === 'Steven Spielberg' &&
      film.genre.indexOf('Drama') !== -1 ) {
      return true
    }
```

src/movies.js
```diff
@@ -1,22 +1,99 @@
...
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage() {
+  const sumOfScores = movies.reduce(function (acc, film) { 
+    return acc + film.score
+  }, 0)
+  return (sumOfScores / movies.length).toFixed(2)
+}
+console.log(scoresAverage(movies)) // => 8.31
```
ahmedhany82, line 46: A few things is missing in this function, first you need to pass the argument as mentioned above. You also need to add a check if the array is empty and return 0 in this case. This you could do by adding a check for that `if (movies.length === 0) return 0;` in the first line in the function.

You also need to check when accumulating the score whether this score is defined.  Check the posted solution by Jan.

Finally, toFixed changes the number to string but the test cases expect a number so we should convert the result back to a number `return Number((sumOfScores / movies.length).toFixed(2));`

src/movies.js
```diff
@@ -1,22 +1,99 @@
...
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore() {}
+function dramaMoviesScore() {
+  const dramas = movies.filter(function (film) {
+    return film.genre.indexOf('Drama') !== -1
+  })
+    const scoreDramas = dramas.reduce(function (acc, film) { 
+      return acc + film.score
+    }, 0)
+  return (scoreDramas / movies.length).toFixed(2)
```
ahmedhany82, line 56: You could instead run the `scoreAverage` function on the `dramas` array. It will do the work for you:
`return scoresAverage(dramas)`

src/movies.js
```diff
@@ -1,22 +1,99 @@
...
+      return -1
+    }
+    if (titleA > titleB) {
+      return 1
+    }
+    return 0
+  }).sort(function(a, b) {
+  return a.year - b.year}) // . sorted by years
+  return moviesSorted
+}
```
ahmedhany82, line 76: If you add the `movies` argument to the function here, it passes all the test cases except one. The problem is that the test case is calling your function with an array like this `[{ year: 2002 }, { year: 1982 }, { year: 1995 }]` which is a bit out of context of the movies data considered in the lab.

In Anyway, I think your approach is correct but you could do it in one `sort` instead of two. Check the solution posted by Jan for a shorter version. Using `localeCompare` is also always simpler and easier

src/movies.js
```diff
@@ -1,22 +1,99 @@
...
+  }).sort(function(a, b) {
+    const titleA = a.toUpperCase()
+    const titleB = b.toUpperCase()
+    if (titleA < titleB) {
+      return -1
+    }
+    if (titleA > titleB) {
+      return 1
+    }
+    return 0
```
ahmedhany82, line 92: You could use `localeCompare` here.
```
.sort(function (a, b) {
    return a.localeCompare(b);
```

ahmedhany82 (commented):
Hey Zhanna,

Good work for the lab. I think the reason the test cases are failing is that you missed to add the parameter to the functions. Remember the functions should expect an array as input. If you try adding a `movies` parameter to each of the functions to represent the passed in array, you will have more test cases passing.

You can find below I added some comments on the code. Feel free to reach out if you have any questions.

Best regards,
Ahmed

PR #2151, opened 2021-06-06
Gergö Eperjesi - Berlin WDFT Maj2021

GergoEper:
done

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,70 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(arr) {
+  const director = arr.map(function (arr) {
+    return arr.director
+  });
+return director
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
-
+function howManyMovies(arr) {
+  let spielbergDramas = arr.filter(function(arr){
+    return (arr.director === 'Steven Spielberg' && arr.genre.includes('Drama'))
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,70 @@
...
+  });
+  if (dramas.length === 0) {
+    return 0;
+  }
+  let dramasScore = dramas.reduce(function (acc, value){
+    return acc + value.score
+  }, 0)
+  let dramasScoreAverage = parseFloat((dramasScore / dramas.length).toFixed(2))
+  return dramasScoreAverage
+};
```
atkat, line 41: You can reuse the scoresAverage function here instead of writing the logic for the average again.

src/movies.js
```diff
@@ -1,22 +1,70 @@
...
+  const ascendingOrder = copyMovies.sort(function (a, b){
+    return a.year - b.year
+  })
+  const ascendingOrderAlph = ascendingOrder.sort(function (a, b){
+    if (a.year === b.year)
+    return a.title.localeCompare(b.title)
+  })
+  return ascendingOrderAlph
+}
+
```
atkat, line 55: You can shorten it by putting if/else into the sort and thus only sorting once:
```
if (a.year !== b.year) {
      return a.year - b.year;
    } else {
    }
```

src/movies.js
```diff
@@ -1,22 +1,70 @@
...
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+function orderAlphabetically(movies) {
+  let copyAlphabetically = JSON.parse(JSON.stringify(movies))
```
atkat, line 59: Here, you don't actually need to copy the array because you are doing a map later. A map always returns a new array so you can  just do the map first and and then the sort.

atkat (commented):
Hi Gergö,

Very nicely done! I only have a few tips, please see the comments below.

Would you like to try one of the bonuses maybe?

Best,
Katka

PR #2152, opened 2021-06-06
BER-WDFT-MAY-2021-Zoriglon Ayush

timbstar58:
last bonus task not finished

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,25 +1,101 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(array) {
+  const allDirectors = array.map(function (arr) {
+      return arr.director
+    });
+  return allDirectors
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(array) {
+  let howManyDramaMovies = 0;
+  const allSpielbergDrama = array.filter(function (arr) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,25 +1,101 @@
...
-function orderAlphabetically() {}
+function orderAlphabetically(array) {
+  
+  let allMovieTitles = array.map(function(arr){
+    return arr.title;
+  }); 
+  let sortAlphabetically= allMovieTitles.sort(function (a,b){
+    return a.localeCompare(b);
+  })
+  return sortAlphabetically.slice(0,20)
```
ahmedhany82, line 77: Just so that you know, you could also call array methods on each other.

```
let allMovieTitles = array.map(function(arr){
    return arr.title;
  }).sort(function (a,b){
    return a.localeCompare(b);
  }).slice(0,20);
  
return sortAlphabetically
```

src/movies.js
```diff
@@ -1,25 +1,101 @@
...
-function turnHoursToMinutes() {}
+function turnHoursToMinutes(arr) {
+  let copyHoursToMinutes =JSON.parse(JSON.stringify(arr));
+  
+  let minutes = 0;
+  let hours = 0;
+  
+  copyHoursToMinutes.map(function (array) {
+  
+  hours = parseInt(array.duration.split(' ') || 0);
```
ahmedhany82, line 90: The `split` function splits a string and returns the result into an array. `parseInt` expects a string. I think the reason the test case passes is that the array here is converted to a string by JavaScript (concatenating all its elements on one string) and `parseInt` picks the number out of this string. 

Another point here is that map always runs on an array and returns a new array. Here you are missing to return an array.

I think a better way is to use slice to pick the number before the 'h' for the minutes and after the space for the minutes. Check the solution Jan posted on Github.

ahmedhany82 (commented):
Hey Zoriglon, 
Good job for the lab. Please find below my comments. Feel free to reach out if you have any questions and maybe give the last bonus one a try if you have some time.
Best regards,
Ahmed

PR #2153, opened 2021-06-06
BER - WDFT - May 2021 - Juan Ramirez

Mangonando:


src/movies.js
```diff
@@ -1,22 +1,85 @@
...
+  let averageDrama = dramaMovies.reduce(function (scoreTotal, movie) {
+    return scoreTotal + movie.score
+  }, 0) / dramaMovies.length
+  return Number(averageDrama.toFixed(2))
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear(movies) {
+  const copy = JSON.parse(JSON.stringify(movies));
```
ahmedhany82, line 53: You can also use` slice()` to copy the array.

`const copy = movies.slice()`

src/movies.js
```diff
@@ -1,22 +1,85 @@
...
+    return a.localeCompare(b)
+  })
+  let movieOrdered = []
+  if (movieAlphabetically.length < 20) {
+    return movieAlphabetically
+  }
+  for (let i = 0; i < 20; i++) {
+    movieOrdered.push(movieAlphabetically[i])
+  }
+  return movieOrdered
```
ahmedhany82, line 80: You could use slice to pick the first 20 elements like so:

`return movieAlphabetically.slice(0,20)`

You could also call the array methods on each others like so:

```
  const ordered = movies.map(function (movie) {
    return movie.title
  }).sort(function (a, b) {
    return a.localeCompare(b);
  }).slice(0, 20);

  return ordered;
}
```

ahmedhany82 (commented):
Hey Juan,

Good work for the lab! Please find my comments below. Feel free to reach out if you have any questions. Maybe give the bonus a try if you have some time.

Best regards,
Ahmed

PR #2175, opened 2021-06-08
BER-WDFT-MAY2021 Elizabeth Hinojos

elihinojos:


src/movies.js
```diff
@@ -1,25 +1,90 @@
...
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(movies) {
+  if (movies.length === 0) return 0;
+  const avgScore = movies.reduce(function (sum, movie){
+    if (movie.score === '' || movie.score === undefined) {
+      movie.score = 0;
+    }
+    return sum + (movie.score || 0)
+  }, 0 / movies.length;
```
atkat, line 41: You're missing a bracket here after the 0.

src/movies.js
```diff
@@ -1,25 +1,90 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(movies) {
+  const directors = movies.map(function (movie){
+    return movie.directors;
```
atkat, line 6: here you have a typo: it should be movie.director

src/movies.js
```diff
@@ -1,25 +1,90 @@
...
+  })
+}
```
atkat, line 8: Your function isn't returning anything: in the tests, you get 'function should return an array'. Were you having trouble finding out why is it giving you that error?

src/movies.js
```diff
@@ -1,25 +1,90 @@
...
+
+const filteredDirectors = new Set(directors)
+const arr = [...new Set(directors)]
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(movies) {
+  const filtered = movies.filter(function (movie) {
+    if (movie.director --- 'Steven Spielberg' &&
+    movie.genre.includes('Drama')){
```
atkat, line 26: this needs to be a triple equals sign movie.director === 'Steven Spielberg'

atkat (commented):
Hey Elizabeth,

Looks good, I only have a few comments, please see below.

Did you understand all of it after you guys went through it with Jan or are there still any spots /methods hat are not 100% clear yet? I am happy to go over them in case you have questions.

Best,
Katka

PR #2208, opened 2021-07-17
[WD PTRM]Cheryl Huang

CherylHuang1988:


IH-WebDev-TA-Remote:
Fantastic work Cheryl, another lab solver! :tada:  Just a small detail: on the `orderAlphabetically()` function, the `const Alphas` should be lowercase according to the naming convention in JS (camelCasing convention). Everything else is amazing, congratulations! :trophy:

PR #2216, opened 2021-07-23
MF LUCE WDPTRM Solved labs Greatest Movies

MFLuce:


IH-WebDev-TA-Remote:
Hello Marie! Congratulations on solving yet another lab! :tada: One quick comment regarding the code:
-> in the `dramaMoviesScore()` function, you're dividing the total score of the drama movies, by the length of **all** the movies! Remember that the array passed as an argument has all the movies. You want to divide it by the length of the array that you filtered above, that includes only the Drama movies :smile: 
Apart from that, excellent job Marie! Keep up the great work! :trophy:

PR #2258, opened 2021-08-14
[BER WDFT Aug2021] Esteban Pettinari

esteban-p:


atkat (commented):
HI Esteban,

I have just a few comments, please see below. I can see that you understand the concepts, which is great! Now it's just time to gain more practice and work on making the code more efficient.

(I haven't reviewed the bonus yet, but there is a solution in Jan's repo).

Best,
Katka

PR #2260, opened 2021-08-14
Berlin-WDFT-Rahaf Abu Alhassan

Farxa:
solved iteration 1 to 6, bonus iterations still under construction.


atkat (commented):
Hey Rahaf,

Well done, I just have a few comments below:)

Best,
Katka

PR #2261, opened 2021-08-15
Caner Ünüs WDFT Berlin AUG2021

kumax0:
Done (without Bonus)


src/movies.js
```diff
@@ -1,22 +1,107 @@
...
+}
+})
+// console.log(sorted); --> logging sorted titles (work)
+// return top 20 of sorted titles
+  if (sorted.length > 20) {
+  return sorted.slice(0, 20);
+}
+// return final result sorted
+return sorted;
+}
```
J-1000, line 105: "allTitles" could be declared using "const" - apart from that and the other detail in iteration one really well done 👍

Try the bonus if you have some time i am sure you will be able to do it.

PR #2266, opened 2021-08-16
iteration1-3

Tanunchai-Dumruttana:


src/movies.js
```diff
@@ -1,13 +1,62 @@
...
+      }
+    }             
+)
+
+ return dramaFromSpielberg.length
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(movie) {
```
atkat, line 43: same thing here as above with the argument

atkat (commented):
Hey Chai,

I see you haven't finished the lab. Is this material more clear to you now or do you still need some help? It is going to be very important for you guys that understand these concepts, would you like to go over any of this again with us?

Best,
Katka

PR #2268, opened 2021-08-16
BERLIN - WDFT - Ludumo Toni solved greatest movies lab

Ludumo:


src/movies.js
```diff
@@ -1,22 +1,74 @@
...
-function scoresAverage() {}
+function scoresAverage(movies) {
+  if (movies.length === 0) return 0;
+  const average = movies.reduce(function(sum, movie) {
+    if (movie.score === '' || movie.score === undefined) {
+      movie.score = 0;
+    }
+    return sum + movie.score
+  }, 0)
+   average / movies.length;
```
atkat, line 36: This line doesn't do anything right now. If you want to reassign the value to average, you need to first declare it with the ```let``` folder and then do ```average /= movies.length ``` which stands for ```average = average / movies.length```/

The other option would be to divide it already in the reduce:
```
const average = movies.reduce(function(sum, movie) {
    if (movie.score === '' || movie.score === undefined) {
      movie.score = 0;
    }
    return sum + movie.score
  }, 0) / movies.length;
```

atkat (commented):
Hey Ludumo, 

Just one comment below :)

Best,
Katka

PR #2271, opened 2021-08-16
[MAD DE] Vanessa Zagone

SybilVane:
Not too bad on the first 6 iterations.
Struggled a lot with number 7, many things still unclear.
Iteration 8, didn't even try. I found some idea on the internet, but I didn't dare trying using it (it's commented out).


ta-web-mad:
Hola Vanesa,

Veo que al final has logrado la iteración 7 y de una manera bastante interesante. En general una solución bastante buena.

Sigue así !

PR #2272, opened 2021-08-16
Tobias.Müller WDFT Aug2021

AveParker:


src/movies.js
```diff
@@ -1,25 +1,68 @@
...
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear() {
+  sortedArr = movies.sort(function(a,b){
+    return a.year - b.year
+  })
+  return sortedArr
+}
+
```
atkat, line 49: The sort function mutates the original array, which is not often desired, so instead we should copy the array first with, for instance, with `JSON.parse(JSON.stringify(movies))` or `array.slice()`. Also, in the case of two movies having the same year, we should sort alphabetically by using localeCompare:
```
function orderByYear(array) {
    if (a.year !== b.year) {
      return a.year - b.year;
    } else {
    }
  });
  return sortedArray;
}
```

src/movies.js
```diff
@@ -1,25 +1,68 @@
...
+    return a+b
+  })
+  return (summedscore/countofscore).toFixed(2)
+  
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear() {
+  sortedArr = movies.sort(function(a,b){
```
atkat, line 44: Here you should be using a let or const. This way of declaring is possible, but should not really be used as it can cause problems.

https://stackoverflow.com/questions/49357522/is-there-a-use-for-not-using-var-let-or-const

src/movies.js
```diff
@@ -1,25 +1,68 @@
...
+}
+
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+function orderAlphabetically() {
+  orderedTitles = movies.sort(function(a,b){
+    return a.title - b.title
+  })
+}
```
atkat, line 56: A duplicate again ☝🏻👇🏼, maybe you forgot to delete one?

src/movies.js
```diff
@@ -1,25 +1,68 @@
...
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
-function turnHoursToMinutes() {}
+function orderAlphabetically() {
+  let onlytitles = movies.map(function(el){
+    return el.title
+  })   
+  orderedtitles = onlytitles.sort()
```
atkat, line 63: Same thing here with the const/let. 

The sort function does work here, but in reality it needs the longer syntax (with a callback function) like this:
```.sort((a, b) => a.localeCompare(b))```

atkat (commented):
Hey Tobi! 

Please see my comments below. Let me know if you have any questions.

Best,
Katka

PR #2289, opened 2021-08-18
[BER-WDFT-0821] Stephen Gallagher

stephen-gallagher:


src/movies.js
```diff
@@ -1,31 +1,112 @@
...
+function orderAlphabetically(array) {
+  let dateOrder = [...array].sort(function (a, b) {
+    if (a.title < b.title) {
+      return -1;
+    } else if (a.title > b.title) {
+      return 1;
+    }
+  });
+  return dateOrder.slice(0, 20);
+}
```
atkat, line 85: We need to use localeCompare again, as this won't sort everything right. 

.sort( (a,b) =>  a.localeCompare(b) )

atkat (commented):
Hey Stephen, 

really not much to say haha. I have one comment and I will look at the bonus later.

Best,
K

PR #2290, opened 2021-08-18
[BER-WDFT-0821] Miguel Acosta

andalucinando:


src/movies.js
```diff
@@ -1,25 +1,139 @@
...
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+
+function orderByYear(array) {
+
+  const sorted = movies.slice().sort(function(a,b){
+    if(a.year !== b.year) {
+      return ayear - b.year;
```
atkat, line 81: typo here: a.year

src/movies.js
```diff
@@ -1,25 +1,139 @@
...
+const average = dramas.reduce((total,next) => total + next.score, 0) / dramas.length; 
+
+console.log(average); 
+
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+
+function orderByYear(array) {
```
atkat, line 77: here again, the argument you're passing is called `array` but you are iterating over `movies`

src/movies.js
```diff
@@ -1,25 +1,139 @@
...
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+
+// users.sort(function(a,b){
+//   return a.firstname.localeCompare(b.firstname);
+// })
+
+
+function orderAlphabetically(array) {
+
```
atkat, line 104: same

src/movies.js
```diff
@@ -1,25 +1,139 @@
...
+    return a.title.localeCompare(b.title)
+
+  })
+
+  const solution = sorted.slice(0,20)
+  
+  return solution;
+  
+}
```
atkat, line 120: here I think you meant to iterate over the `titles` array. Also, the titles array is now just an array of strings so there is no need to do a.title:
```

  const titles = movies.map(function(movie){
    return movie.title
  })


  const sorted = titles.sort(function(a,b) {

    return a.localeCompare(b)

  })

  const solution = sorted.slice(0,20)
  
  return solution;
  
}
```

atkat (commented):
Hey Miguel,

Please see my comments below. If you want we can go over them together.

Best,
Katka

PR #2291, opened 2021-08-18
The Greatest Movies

jflores08:
I will attempt the Bonuses when caught up :)

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,19 +1,94 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(array) {
+  const allDirectors = array.map(function(movie){
+    return movie.director
+  })
+  return allDirectors;
+} 
+
+// allDirctors.filter(function(director, i){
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,107 @@
...
+  allScores = array.map(function(movie){
+    return movie.score
+  })
+  const totalScores = allScores.reduce(function(acc,curr){
+    return  acc+(curr||0)
+    
+  },0)
+  let Average
+  if (allScores.length == 0){
+    Average =0
```
atkat, line 41: As of the naming convention, we should only name classes and React components with a capital letter.

src/movies.js
```diff
@@ -1,22 +1,107 @@
...
+  } else
```
atkat, line 42: This else doesn't do anything here. It's not flagged up as an error but an else needs curly brackets for it to work on multiple lines.  You don't actually need an else here anyway - if the condition is not met, the function will just move on to the next thing.

```
function scoresAverage(array) {
  let allScores = array.map(function(movie) {
    return movie.score
  })
  const totalScores = allScores.reduce(function(acc,curr) {
    return  acc + (curr || 0)  
  },0);

  let average;
  if (allScores.length === 0) {
    average = 0
  } 

  average = totalScores / array.length
}
```

Also, please note the formatting above, try adding some more horizontal spacing for readability.

atkat (commented):
Hey Jon, 

Please see my comments below.

Best,
Katka

PR #2306, opened 2021-09-23
[FT-RMT-092021] Aintzane Morales

AintzaneM:
Hey Karina!

Here it is the LAB 4. I was struggling a lot with the new lessons. I couldn't finish iterations 5 and 6.

Thank you:)

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,69 @@
+
+
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+
+
+function getAllDirectors(moviesArr) {
+  const totalDirectors = movies.map(function(filmObj){
+    return filmObj.director;
+  });
+  console.log(totalDirectors);
+}
+
+
+
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(arr) {
+  const result = movies.filter(function (film) {
+    return film.director === 'Steven Spielberg' && film.genre.includes('Drama');
+      
+  });
+  console.log(result.length);
+}
+  
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(arr) {
+  if(!movies.length){
+    return 0;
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,69 @@
+
+
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+
+
+function getAllDirectors(moviesArr) {
+  const totalDirectors = movies.map(function(filmObj){
```
TA-Remote, line 9: You pass 'moviesArr' as argument, this array contains your movies data that you pass into your function -> OK
Then you use map method on 'movies' variable -> what does this array contain? ;)

Be consistent with the names you give.  If function takes moviesArr as parameter, that is what you will be using.  eg. moviesArr.map()

Review same issue in all the iterations ;)

src/movies.js
```diff
@@ -1,16 +1,69 @@
...
+    return filmObj.director;
+  });
+  console.log(totalDirectors);
```
TA-Remote, line 12: Do not forget:  console.log and return is not the same thing. You usually want to return the result of your function

-> Console.log is a javascript function to print something in the console.
-> While return saves the result of your function in memory, so you can reuse that result or print it later.
     Return by itself does not print anything 

Review same issue in all the iterations ;)

src/movies.js
```diff
@@ -1,16 +1,69 @@
...
+  if(!movies.length){
+    return 0;
+  }
+  const result = movies.map(function (film) {
+    return film.score;
+  });
+  
+  const sum = movies.reduce(function(acc, currentValue){
+    return acc + currentValue;
+  }, 0);
```
TA-Remote, line 40: When working with objects, pay attention if you need to access full object or its property. In this case you only want to be adding score. 

Movies -> it is array of movie objects
CurrentValue -> is single movie object

So what you want to be adding  is: acc + currentValue.score

TA-Remote (commented):
Overall great effort! You have good thinking, but please pay attention to details ;) Read lab instructions clearly -> if it says create a function that returns the average score. It means it literally has to RETURN something - that's your clue already ;)

Pay attention to:
-> consistent variable names
-> console.log vs return in functions

BONUS EXERCISE: try to convert your anonymous functions into arrow functions, for example this could be written in one line ;)

const totalDirectors = movies.map(function (filmObj) {
    return filmObj.director;
  });

PR #2307, opened 2021-09-23
[FT-RMT-092021] - NEKRUZ AVGANI 

nekruza:


src/movies.js
```diff
@@ -1,16 +1,56 @@
...
-function dramaMoviesScore() {}
+
+function scoresAverage(key) {
+  const total = key.reduce(function (acc, currentValue) {
+    if (movies.genre === 'Drama') {
+      return acc + currentValue.score;
+    }
+  }, 0);
+  return total / key.length;
+}
```
TA-Remote, line 51: 1. Is that correct name for the function? ;) You are checking drama movies score here...
2. what is 'movies' here on line 46? This variable is not defined in your function.. Instead of that you should be accessing the currentValue .genre
3. Always analyze the data you are working with, if you look at data.js file, you can see that genre in your object is actually an array. So instead of checking if array is equal to a string, you should be checking if array includes that string

TA-Remote (commented):
Hey! Firstly good job, your code is tidy and easy to read, good syntax. However, please please pay attention to details :) You have good thinking, I can see that you understand the task, but get stuck on details. 

Read lab instructions carefully:
-> if it says create a function that returns the average score. It means it literally has to RETURN something - that's your clue already ;)
-> if it asks for Spielberg dramas, it means you need to filter for both conditions etc

Pay attention to:
-> consistent variable names
-> choose variable name that make sense (that describes the variable)
-> console.log vs return in functions

It is clear you understand the concepts, but need to spend more time practising :muscle:

PR #2308, opened 2021-09-23
[FT-RMT-092021] Geoffre Laville

GeoLvl:
Roller coaster feeling !!

I lost a big amount of time and I only did the first 2 iterations 
I don't understand what I've been missing

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,40 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(movies) {
+    const directorsNames = movies.map((movie) => {
+      return movie.director
+    }); 
+      return directorsNames; 
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
-
+function howManyMovies(movies) {
+  const countMovies = movies.filter((movie) => {
+    if (movie.director === "Steven Spielberg" && movie.genre.indexOf('Drama') !== -1) {
+      return movie;
+    }    
+  });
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,40 @@
...
+ 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(movies) {
+  const average = movies.reduce((acc, curr) => {
+    console.log(`${acc} .... ${curr}`);
+  return acc + curr;
+  });
+  const averageRounded = average/movies.length;
+  return averageRounded.toFixed(2);
```
TA-Remote, line 29: This is a tricky one! toFixed() method actually returns a string, so when you use it, you need to convert your result back to number.

i have mentioned before, bur very useful if you read the error you are getting from the test. Cause this might give you clues how to solve it.

Failed test says:
Function "scoresAverage"
should return a number

Expected: "number"
Received: "string"

One option to convert string to a number:
Number(averageRounded.toFixed(2));

src/movies.js
```diff
@@ -1,16 +1,40 @@
...
+  return countMovies.length;
+}
+    
+ 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(movies) {
+  const average = movies.reduce((acc, curr) => {
+    console.log(`${acc} .... ${curr}`);
+  return acc + curr;
```
TA-Remote, line 26: When working with objects, pay attention if you need to access full object or its property. In this case you only want to be adding score.

Movies -> it is array of movie objects
Curr -> is single movie object

So what you want to be adding is: acc + curr.score

TA-Remote (commented):
Rollercoaster feeling - very common. I've been up and down many times during my own bootcamp:D Everyone works in their own pace, so it is fine if you can't finish it all.

You will see later on -  it is better to complete tasks in small chunks and in good quality rather than a little bit of everything half way. The iterations you have done are :ok_hand:

I suggest to go through the lectures in student portal over the weekend and then try to apply them again in the lab :)

PR #2309, opened 2021-09-24
[FT-RMT-092021] Alice

AliceXix:
I gained some clarity about the difference between variables and objects. I am still struggling when it gets to access properties of nested objects.
I couldn't get the last iteration to work. After trying to debug with help of console.log, I am quite certain the issue is in my if statement and would even assume in my for loop.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -69,7 +69,37 @@ function orderByYear(a, b) {
 //console.log(movies);
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+
+
+function orderAlphabetically(arr) {
+
+  const sortedMovies = movies.sort(function (a, b) {
+    if (a.title < b.title) {
+      return -1;
+    }
+    if (a.title > b.title) {
+      return 1;
+    }
+    return 0;
+  });
+
+console.log(sortedMovies)
+
+  const arrLength = sortedMovies.length;
+
+console.log(arrLength)
+
+  if (arrLength <= 3) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -69,7 +69,37 @@ function orderByYear(a, b) {
...
+
+  if (arrLength <= 3) {
+    return sortedMovies;
+  } else {
+    for (let i = 0; i > sortedMovies[3]; i++) {
+     return sortedMovies.pop();
+    }
+  }
+
+}
```
TA-Remote, line 100: 1. Do not forget to have return at the end of the function
2. Function has to return array of sorted movie titles (not the array of objects)

So you might need to create new array with only movie titles, maybe use map method for that?

src/movies.js
```diff
@@ -69,7 +69,37 @@ function orderByYear(a, b) {
...
+
+  const arrLength = sortedMovies.length;
+
+console.log(arrLength)
+
+  if (arrLength <= 3) {
+    return sortedMovies;
+  } else {
+    for (let i = 0; i > sortedMovies[3]; i++) {
+     return sortedMovies.pop();
```
TA-Remote, line 96: I am not entirely sure what you are trying to do here, what does  '<= 3' mean ? :)

TA-Remote (commented):
Hi Alice, I can see that you put a lot more effort into this lab, well done! Left some comments for what you can improve here. As we spoke, definityely review the recording/or lecture notes over the weekend to recap. It takes time and practice and all concepts will fall into places, so keep on working!

PR #2310, opened 2021-09-24
[FT-RMT-092021] Zeynep Alpaslan

CodingEinhornDeu:
this was the worst lab for me.I just need to clarify usage of this functions.At the weekend i will do that. :(

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,28 +1,59 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
+
+const movies = require("./data");
+
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(mArray) {
+  const allDirectors = mArray.map(function (direct) {
+    return direct.director;
+  });
+  return allDirectors;
+}
+getAllDirectors(movies);
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(arrMov) {
+  let numbOfMov = arrMov.filter(function (movie) {
+    if (movie.director === 'Steven Spielberg' && movie.genre.includes('Drama'))
+      return movie;
+
+  });
+  return (numbOfMov.length);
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,28 +1,59 @@
...
+function dramaMoviesScore() { }
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear(movieYear) {
+  const result = movieYear.sort((firstScore, secondScore) => {
+    if (firstScore.year > secondScore.year) {
+      return -1;
+    } else {
+      return 1;
```
TA-Remote, line 39: All logic is correct, you just mixed up the 1 with -1. 

function compare(a, b) {
  if (a is more than b) {
    return 1;
  }
  if (a is less than b) {
    return -1;
  }
}

src/movies.js
```diff
@@ -1,28 +1,59 @@
...
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore() {}
+function dramaMoviesScore() { }
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear(movieYear) {
+  const result = movieYear.sort((firstScore, secondScore) => {
```
TA-Remote, line 35: This iteration was tricky, because the task asks to return a new array.
Sort method changes the existing array, so it means firstly you have to make a copy of your original array, and only then sort() that new array:

function orderByYear(movieYear) {
    // clone your array:
       const result = 'you make copy of your movieYear array here'
   
    // sort your new array
       result.sort((firstScore, secondScore) => {})


couple of options to clone an array:
-> spread operator:
     const newArray = [...originalArray]
-> slice():
     const newArray = originalArray.slice()

TA-Remote (commented):
Hi Zeynep, great effort overall ! I can see you are putting work to solve the labs and follow the lectures. 'Clue' lab you did great, so well done! 

I see you have more difficulty with functions (movies lab and functions&arrays lab), so my advise is to review recordings or lectures on student portal  about functions.

src/movies.js
CodingEinhornDeu: Thank you for your feedback:) it is really important for me.I will do it:)

PR #2311, opened 2021-09-24
[FT-RMT-092021] Jeannette Valencia

JeannetteValencia:
I've managed to do iterations 1, 5, 6.
Iteration 2: the conditions in the return seemed not to be ok
Iteration 3: I received a msg that the score of the 'accValue.score' was not defined...
Iteration 4: I was not able to understand this task

I will keep on practising later!

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,55 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(movies) {
+  const mappedArray = movies.map(function (singleMovie){
+    return singleMovie.director;
+   })
+   return mappedArray;
+}
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(movies){
+  const movieDirector = movies.filter( (directorName) => {
+    return directorName.director === 'Steven Spielberg';
+  })
+  for (let i=0; movieDirector === 'Steven Spielberg'; i++){
+    const directorDrama = movieDirector.filter( (dramaMovie) => {
+      return dramaMovie.genre === 'Drama';
+    })
+  }
+  return directorDrama;
 
+}
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(movies) {
+  let movieScores = movies.reduce(function (accValue, currentValue){
+    return accValue.score + currentValue.score
+  })
+  return movieScores
 
+}
 // Iteration 4: Drama movies - Get the average of Drama Movies
 function dramaMoviesScore() {}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear (movies) {
+  const newArray = movies.map(function (movieYear){
+    return movieYear.year
+   })
+  return newArray.sort();
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+function orderAlphabetically(movies) {
+  for (let i=0; i<20;i++){
+    const newArr = movies.map(function(movieTitle){
+     return movieTitle.title
+    })
+    return newArr.sort();
+  }
+}
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 function turnHoursToMinutes() {}


src/movies.js
```diff
@@ -1,22 +1,55 @@
...
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(movies){
+  const movieDirector = movies.filter( (directorName) => {
+    return directorName.director === 'Steven Spielberg';
+  })
+  for (let i=0; movieDirector === 'Steven Spielberg'; i++){
+    const directorDrama = movieDirector.filter( (dramaMovie) => {
+      return dramaMovie.genre === 'Drama';
```
TA-Remote, line 19: Pay attention to the data you are given, movie genre is actually and array, not a string, for example:
genre: ['Crime', 'Drama'],

So you won't be able to compare if dramaMovie.genre === 'Drama'. Instead you need to check if genre array includes 'Drama' string ;)

dramaMovie.genre --> is an array of genres
Drama --> is a single genre

src/movies.js
```diff
@@ -1,22 +1,55 @@
...
+    })
+  }
+  return directorDrama;
```
TA-Remote, line 22: This might be tricky, but the task asks how many movies there are (not asking for the array of movies), so function is expected to return an integer, not an array. Any ideas how to get that? :)

src/movies.js
```diff
@@ -1,22 +1,55 @@
...
 
+}
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(movies) {
+  let movieScores = movies.reduce(function (accValue, currentValue){
+    return accValue.score + currentValue.score
+  })
```
TA-Remote, line 29: My suggestion is to go back to lectures in student portal, 'Arrays & Objects'  specifically, you have detailed example there on how to use reduce method on array of objects. 
1.  when using reduce with objects, you always need to define initial value, add zero usually:

  let movieScores = movies.reduce(function (accValue, currentValue){
    return accValue.score + currentValue.score
  }, 0)

2. Since accValue is an initial value that you set yourself, it is an integer (not an object), and has no properties That's why you get that error, ''accValue.score' was not defined'.

src/movies.js
```diff
@@ -1,22 +1,55 @@
...
 function dramaMoviesScore() {}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear (movies) {
+  const newArray = movies.map(function (movieYear){
+    return movieYear.year
+   })
+  return newArray.sort();
+}
```
TA-Remote, line 42: 1. Pay attention that task asks to return a new sorted array. We mentioned yesterday that sort() method changes original array, so before you use it you need to make a copy of original array.

couple of options to clone an array:
-> spread operator:
const newArray = [...originalArray]
-> slice():
const newArray = originalArray.slice()

2. The task asks you to sort the array, and return new sorted array of movie objects. Does map method really work here? Maybe we can use sort method instead of map... Check sort method syntax in student portal lecture ''Arrays Sort & Reverse" 

Sort method iterates through your arrays and compares two values:

function compare(a, b) {
    if (a is more than b) {
       return 1;
    }
    if (a is less than b) {
       return -1;
    }
}

src/movies.js
```diff
@@ -1,22 +1,55 @@
...
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+function orderAlphabetically(movies) {
+  for (let i=0; i<20;i++){
+    const newArr = movies.map(function(movieTitle){
+     return movieTitle.title
+    })
+    return newArr.sort();
+  }
+}
```
TA-Remote, line 52: Since the task asks you to print first 20values from sorted array, maybe we can first sort and then print? So we actually get the alphabetically first 20 titles ;)

Same as in iteration5, please review how sort method works.

src/movies.js
```diff
@@ -1,22 +1,55 @@
...
-function scoresAverage() {}
+function scoresAverage(movies) {
+  let movieScores = movies.reduce(function (accValue, currentValue){
+    return accValue.score + currentValue.score
+  })
+  return movieScores
 
+}
 // Iteration 4: Drama movies - Get the average of Drama Movies
 function dramaMoviesScore() {}
```
TA-Remote, line 34: Steps for iteration 4:
1. create a function that receives array of movies
2. filter that array so we have only drama movies. --> do nto forget genre is an array, so you might want to use .include() method
3. use reduce method to add all movie scores
4. calculate and return the average score of drama movies

TA-Remote (commented):
HI Jeannette, great effort! Happy to see you tried to work through all iterations. My advise for you is to review lectures on student portal, specifically: 'Arrays & Objects', 'Arrays - Sort & Reverse' and try to implement those examples in your lab :) 

Keep up great work! practice practice practice  :muscle:

src/movies.js
```diff
@@ -1,22 +1,55 @@
...
+}
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(movies){
+  const movieDirector = movies.filter( (directorName) => {
+    return directorName.director === 'Steven Spielberg';
+  })
+  for (let i=0; movieDirector === 'Steven Spielberg'; i++){
```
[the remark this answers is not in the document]
JeannetteValencia: I've filter it again, but I think I still need a counter to count how many movies contain Drama...maybe a while?

  const movieDirector = movies.filter( (directorName) => {
    return directorName.director === 'Steven Spielberg';
  })
  const directorDrama = movieDirector.filter ((dramaGenre)=> {
    return dramaGenre.genre === 'Drama';
  })
  return directorDrama;
};

src/movies.js
JeannetteValencia: that's true...
I would need the position to access the strings or a loop to go through each string

src/movies.js
JeannetteValencia: I think we need a counter somewhere...

src/movies.js
JeannetteValencia: Ohh I see, I've read it again

PR #2415, opened 2021-11-13
Diogo & Vitor WD_PT_Lisbon

diogoascarneiro:
Did all the exercises except the bonus

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,19 +1,61 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(movies) {  
+  const directors = movies.map(filmes => filmes.director);
+  return directors;
+}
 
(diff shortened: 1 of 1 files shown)


carlarsmendes:
Iteration 1 - How to get rid of duplicates "efficiently"
Check these 2 advanced features of JS: Sets and Spread operator.
You can create a new Set with the directors and then turn it into an array by using the spread operator
```
var s = new Set(directors);
let arr = [...s];
```
Iteration 3: All scores average - Get the average of all scores with 2 decimals
Using only reduce:

```
function ratesAverage(lotsOfMovies) {
  //  return !lotsOfMovies.length? 0 : Number((lotsOfMovies.reduce((a,b) => b.rate? a + b.rate : a, 0)/lotsOfMovies.length).toFixed(2));
  // this is the one line version

  if (!lotsOfMovies.length) {
    return 0;
  }

    if (b.rate) {
      return a + b.rate;
    } else {
      return a;
    }
  }, 0);

}
```


Iteration 6 - A different solution for OrderAlphabetically: 

```
function orderAlphabetically(lotsOfMovies) {
  return [...lotsOfMovies]
    .sort((a, b) => {
      if (a.title > b.title) {
        return 1;
        return -1;
      } else {
        return 0;
      }
    })
    .slice(0, 20);
}
```


Nicely done!!!!

PR #2423, opened 2021-11-18
WDPT NOV2021 André && Ana

andregn26:
We had some difficulty during the exercise. The biggest difficulties occurred in the use of //locale.compare and in the execution of the last iteration

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,58 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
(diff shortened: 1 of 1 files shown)


carlarsmendes:
Hi guys, I like your solutions, congrats! You did a good job - This is a tough lab, it's normal for students to struggle with it, it's a new way of looking at: objects; arrays; functions; And you're just barely getting confortable with some of these concepts
You have time to get used to map, filter, reduce. 
Keep up the good work!

On your last iteration:
You didn't really need to use localeCompare, here's an alternative solution if you'd like to check it out:
```
newArray.sort((a, b) => {
    if (a.year > b.year) {
      return 1;
    } else if (b.year > a.year) {
      return -1;
    } else {
      if (a.title > b.title) {
        return 1;
        return -1;
      }
      return 0;
    }
  });
  return newArray;
```

PR #2456, opened 2022-01-13
[REMOTE FT 012022] Denise Sengül

denisesenguel:


src/movies.js
```diff
@@ -1,28 +1,111 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(allMovies) {
+  
+  const allDirectors = allMovies.map(movie => movie.director);
+  const uniqueDirectors = allDirectors.filter((element, index, array) => array.indexOf(element) === index)
+  
+  return uniqueDirectors
```
TA-Remote, line 9: Since you clearly understand how these methods work, a note for you. You can always combine the methods, so for example this would be correct expression 

`return allMovies.map(<....>).filter(<......>)`

```
return allMovies.map(movie => movie.director).allDirectors.filter((element, index, array) => array.indexOf(element) === index)
```

TA-Remote (commented):
Well done Denise, really great work!

TA-Remote:
Testing testing.

PR #2457, opened 2022-01-13
[WDFT - Jan 22] Day 4 Lab - Doan Tran

Dorrrn:
Iteration 1-4 done

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,51 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
+
+const movies = require('./data');
+
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(movies) {
+  const allDirectors = movies.map(function (film) {
+    return film.director;
+  });
+  return allDirectors;
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(movies) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,51 @@
...
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
+
+const movies = require('./data');
+
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(movies) {
+  const allDirectors = movies.map(function (film) {
+    return film.director;
+  });
```
TA-Remote, line 10: If you clearly understand this syntax, start practising arrow functions, shorter syntax and you'll use it more in the future ;)
```
 const allDirectors = movies.map((film) => film.director);
```

If it is only one line of code after arrow, you do not need curly brace and return.

src/movies.js
```diff
@@ -1,16 +1,51 @@
...
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore() {}
+function dramaMoviesScore(movies) {
+  const dramaMovies = movies.filter(function (film) {
+    return film.genre.includes('Drama');
+  });
+  return scoresAverage(dramaMovies);
+}
```
TA-Remote, line 48: The task is also asking to check if your movie object has score property. So first you need to check that , and if it exists you go on with your logic

TA-Remote (commented):
Great work on the lab! I recommend practising arrow functions as well, so as an exercise you can try converting the ones you have :) 

Also I advise to work on remaining interations.

src/movies.js
Dorrrn: thanks for your feedback, appreciate it <3

PR #2459, opened 2022-01-13
Marian Gispert- WDFT-JAN2022

mgispert:
Kinda struggled with reduce() in scoresAverage() but finally managed. However I couldn't even try the labs, so exhausted!!


TA-Remote:
Hi Marian! The array methods can be very complex, especially in the beginning, and you've done a great job with that. I don't have much to say about your code, because you code very clearly and well.
There are shorter ways to code, some of which we will learn, but tastes differ. I and also some more experienced developers think the more readable the better and some think the shorter the better. What is super and important though: reuse code, which you did for example in dramaMoviesScore by reusing the scoresAverage function. Pretty good, I feel like you've understood the concepts very well so far. Keep up the good work! :)

PR #2460, opened 2022-01-13
[Remote-FT-012022] Zena Bos

zenabos:


src/movies.js
```diff
@@ -1,19 +1,60 @@
...
+  const moviesDrama = movies.filter((movie) => movie.genre.includes('Drama'));
+
+  if (moviesDrama.length === 0) return 0;
+  const calcSum = moviesDrama.reduce((sum, movie) => sum + movie.score, 0);
+
+  let average = Math.round((calcSum / moviesDrama.length) * 100) / 100;
+  return average;
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
```
TA-Remote, line 50: really good work, Zena!

src/movies.js
```diff
@@ -1,19 +1,60 @@
...
-function orderByYear() {}
+function orderByYear(movies) {
+  let sortedMovies = movies.sort((a, b) => {
```
TA-Remote, line 52: one thing you need to do here is the deep copy of the movies array. 
For example like this:
  const moviesArray = JSON.parse(JSON.stringify(movies));
  let sortedMovies = moviesArray.sort((a, b) => {
    return a.year - b.year;
  });

  return sortedMovies;
}

TA-Remote: to make sure, that you sort by title if two movies have the same year you need to do it for example like this:
```
  const moviesArray = JSON.parse(JSON.stringify(movies));
  let sortedMovies = moviesArray.sort((a, b) => {
    if (a.year > b.year) {
      return 1;
    } 
    else if (b.year > a.year) {
      return -1;
    } 
    else {
      if (a.title > b.title) {
        return 1;
        return -1;
      }
      return 0;
    }
  });

  return sortedMovies; 
}

```

PR #2463, opened 2022-01-13
WDFT GJ Day 4 labs

gjschoutsen:
This was a fun one! 

But i couldn't figure out how to solve the last test at iteration 3 
"should return average even if one of the movies does not have score"

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,58 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
-
+function getAllDirectors(arr) {
+  const pickDirector = arr.map((x) => x.director);
+  return pickDirector;
+}
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(arr) {
+  const sortedMovie = arr.filter((movie) => {
+    if (
+      movie.director === 'Steven Spielberg' &&
+      movie.genre.indexOf('Drama') > -1
+    ) {
+      return true;
+    }
+  });
+  return sortedMovie.length;
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(arr) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,58 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
-
+function getAllDirectors(arr) {
+  const pickDirector = arr.map((x) => x.director);
+  return pickDirector;
```
TA-Remote, line 6: If the code is this short, you can alway return result directly - no need to create variable

```
return arr.map((x) => x.director)
```

src/movies.js
```diff
@@ -1,16 +1,58 @@
...
-function howManyMovies() {}
+function howManyMovies(arr) {
+  const sortedMovie = arr.filter((movie) => {
+    if (
+      movie.director === 'Steven Spielberg' &&
+      movie.genre.indexOf('Drama') > -1
+    ) {
+      return true;
+    }
+  });
```
TA-Remote, line 17: with `filter()` method, you do not have to actually return true. If the condition passes -> the element will be moved to new array. So for example, the below would be equivalent to what you have:

```
  const sortedMovie = arr.filter((movie) => {
    return (
      movie.director === 'Steven Spielberg' &&
      movie.genre.indexOf('Drama') > -1
    ) 
  });
```

To make it more shorter, if you code is only one line - when using arrow function you can skip curly brace and return 

```
  const sortedMovie = arr.filter((movie) => movie.director === 'Steven Spielberg' &&
      movie.genre.indexOf('Drama') > -1);
```

But please only use this shorter syntax if you understand it! ;)

src/movies.js
```diff
@@ -1,16 +1,58 @@
...
+  return sortedMovie.length;
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(arr) {
+  const sumScore = arr.reduce((prevVal, curVal) => {
+    return prevVal + curVal.score;
+  }, 0);
```
TA-Remote, line 25: The failing test you mentioned. Inside reduce method, before you sum those two values, you need to check if your movie object has property `score`, in other words if this property exists

TA-Remote (commented):
Great work GJ! I advise working on the remaining iterations to get in more practice

PR #2465, opened 2022-01-13
WDFT Remote Helena Saborit

Helsinky91:
completed only 1-4, struggled a bit with it. 2. Tomorrow morning I'll try iteration 5 & 6.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,64 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(movies) {
+  const directorList = movies.map(element => {return element.director});
+  return directorList
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+
+function howManyMovies(movies) {
+  if ( movies.length === 0) {
+    return 0;
+  } 
+  const spielbergDrama = movies.filter(element => { 
+    return element.director === 'Steven Spielberg' && element.genre.includes('Drama');
+  });
+return spielbergDrama.length
+} 
+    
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,23 +1,65 @@
...
+function howManyMovies(movies) {
+  if ( movies.length === 0) {
+    return 0;
+  } 
+  const spielbergDrama = movies.filter(element => { 
+    return element.director === 'Steven Spielberg' && element.genre.includes('Drama');
+  });
+return spielbergDrama.length
+} 
+
```
TA-Remote, line 21: first two are great - nothing to say

src/movies.js
```diff
@@ -1,23 +1,65 @@
...
+function scoresAverage(movies) {
+  if ( movies.length === 0) {
+    return 0; }
+  const average = movies.reduce((acc, value) => {
+    return acc + value.score;
+    },0); 
+
+let scoreResult = average / movies.length
+return scoreResult
+}
```
TA-Remote, line 32: to solve this "should be rounded to 2 decimals places"
you can do this: `let scoreResult = Number((average / movies.length).toFixed(2))`

TA-Remote: to solve this "should return average even if one of the movies does not have score"
you can do this: 
```
const average = movies.reduce((acc, value) => {
    if (value.score){
      return acc + value.score;
    }
    else {
     return acc
    }
```

src/movies.js
```diff
@@ -1,23 +1,65 @@
...
+    return element.genre.includes('Drama') ;
+ });
+ const dramaScore = dramaMovies.reduce((acc, value) => {
+   return acc + value.score;
+ },0);
+
+ let dramaResult = dramaScore / dramaMovies.length
+ return dramaResult
+
+}
```
TA-Remote, line 46: to solve this "should return the average of the array, a floating point number"
you can reuse the last function!
like this:
```
function dramaMoviesScore(movies) {
  const dramaMovies = movies.filter(element => {
    return element.genre.includes('Drama') ;
 });
 return scoresAverage(dramaMovies)
}
```

TA-Remote: and that's it! Now all of the mandatory tasks (iteration 1-4) are correctly. So all in all a good job! Keep up the good work!

PR #2474, opened 2022-01-14
bart torture soul solved lab

Bartocaster:
this was waaaaaay to hard i feel crushed and i pushed myself to hard

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,30 +1,140 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(ListOfMovies) {
+  let ListOfDir = ListOfMovies.map(function (element) {
+    return element['director'];
+  });
+
+  return ListOfDir;
+}
+// console.log('---------');
+// console.log('---------');
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(Datalist) {
+  // we are going to find out how amny drama movies spielberg made
+  // we need to sort out all the movies speilberg made / map
+  // than we need to filter wich one of them are drama / filter
+  let listOfSpielberg = Datalist.filter(function (element) {
+    return element['director'] === 'Steven Spielberg';
+    // This is an element
+
+    // {
+    //   title: 'Indiana Jones and the Last Crusade',
+    //   year: 1989,
+    //   director: 'Steven Spielberg',
+    //   duration: '2h 7min',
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,31 +1,141 @@
...
+  let listOfDrama = listOfSpielberg.filter(function (element) {
+    for (let i = 0; i < element['genre'].length; i++) {
+      let current_genre = element['genre'][i];
+      //return current_genre === "Drama";
+      if (current_genre === 'Drama') {
+        return true;
+      }
+    }
+    return false;
+  });
```
TA-Remote, line 41: to filter only drama movies, you can use exact same syntax as above for director.  Since `filter() `method is already  a loop, you do not need to as additional` for loop`

src/movies.js
```diff
@@ -1,31 +1,141 @@
...
+  // cheak if these movies have drama
+  let listOfDrama = listofmoviesWithGenre.filter(function (movie) {
+    // i need to look through the list of genre and cheak if it containts drama
+    for (i = 0; i < movie['genre'].length; i++) {
+      if (movie['genre'][i] === 'Drama') {
+        return true;
+      }
+    }
+    return false;
+  });
```
TA-Remote, line 81: Again, `filter()` already iterates through your array, so using another loop is not neccesary ;)

Firstly you filter to only movies who have genre property. 
```
  let listofmoviesWithGenre = listOfmovies.filter(function (element) {
    return element['genre'];
  });
```

And you can do exactly same to get only dramas, for example:

```
  let listofmoviesWithGenre = listOfmovies.filter(function (element) {
    return element['genre'].includes('Drama');
  });
```

src/movies.js
```diff
@@ -1,31 +1,141 @@
...
+  let listOfMovieTitles = listOfmovies.map(function (element) {
+    return element['title'];
+    
+  });
+    if (listOfMovieTitles > 20){
+    // if there are more than 20 tiltes return only 20 of them
+    return element['title'].slice(0,20)
+  }
+    //console.log(listOfMovieTitles)
+  return listOfMovieTitles
```
TA-Remote, line 128: line 123-126 does not update your list. Instead you want to 'crop' your final result:
```
 return listOfMovieTitles.slice(0,20)
```

src/movies.js
```diff
@@ -1,31 +1,141 @@
...
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+function orderAlphabetically(listOfmovies) {
+  //sort all the movie titles
+  
+  listOfmovies.sort(function (a, b) {
+    return a['title'].localeCompare(b['title']);
+  });
```
TA-Remote, line 117: Task asks you not to mutate the original array ( which sort() method does! ). So you want to make a copy of `listOfmovies` array, and only then sort it. I shared the link yesterday of several options to clone array

TA-Remote (commented):
Really great effort, well done. I can see you still overcomplicate certain things, but you do understand and make it work. So practice & practice and you 'll be great ;)

PR #2476, opened 2022-01-14
done

throwoutofcoffeeexception:


src/movies.js
```diff
@@ -1,28 +1,76 @@
+//let fs = require("fs");
+const movies = require("./data");
+//let Data = fs.readFileSync("./data.js", "utf-8");
+
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors() {
```
TA-Remote, line 10: Your function is not recognising what `movies `variable is, so tests are not passing. Every function need to receive the argument

src/movies.js
```diff
@@ -1,28 +1,76 @@
...
+  let Directors = []
+  movies.forEach(element => {
+    Directors.push(element.Directors)
+  });
```
TA-Remote, line 14: This solution works! But please try solving this this iteration using `map() `method

src/movies.js
```diff
@@ -1,28 +1,76 @@
...
+function howManyMovies() {
+  let counter = 0;
+  movies.forEach(element => {
+    if (element.Directors == 'Steven Spielberg')
+    {
+      counter++;
+    }
+  });
+  return counter;
+}
```
TA-Remote, line 28: - This iteration should be solved using `filter() `method ;) 
- Task asks to count all 'Drama' movies by Spielberg. So you need to be checking both director and genre

src/movies.js
```diff
@@ -1,28 +1,76 @@
...
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage() {
+  let Average = 0;
+  movies.forEach(element => {
+      Average += element.scoresAverage;
+  });
+  return (Average / movies.length).toFixed(2);
+}
```
TA-Remote, line 37: - Should be solved using `reduce() `method. 
- toFixed() method actually returns a string, so you need to convert this result back to type number. Couple of option is to use `parseInt() `or `Number() `method

src/movies.js
```diff
@@ -1,28 +1,76 @@
...
+  let Counter = 0;
+  movies.forEach(element => {
+      if (element.genre.find("Drama")
+      {
+        Average += element.score;
+        Counter++;
+      }
+  });
+  return Average/Counter;
+}
```
TA-Remote, line 51: - Should be using `filter()`
- Line 42. property genre is an array, so instead you need to check if string 'Drama' is inside that array

src/movies.js
```diff
@@ -1,28 +1,76 @@
...
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear() {
+  return movies.sort((a) => a.score);
```
TA-Remote: Please refer back to student portal how sort method works with numbers :)

src/movies.js
```diff
@@ -1,28 +1,76 @@
...
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+function orderAlphabetically() {
+  return movies.sort((a) => a.title);
```
TA-Remote, line 60: - Before you use sort() method, firstly you want to clone the original array
- Also task asks you to return first 20 titles ;)

PR #2481, opened 2022-01-15
Symvolokova Resolution of the greatest movies lab 

ansymvolokova:


src/movies.js
```diff
@@ -1,25 +1,94 @@
...
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllFilteredDirectors(arr) {
+  const directorsWithDuplicates = movies.map(film => film.director);
+    const filtered = directorsWithDuplicates.filter(function(director, index){
+      return directorsWithDuplicates.indexOf(director) === index
+    })
+    return filtered
+  }
```
Farxa, line 14: Hi Anastasiia,

First of all let me thank you for writing such clean code with good naming that it is easy for me to read and understand. 🙏🏽. I will give you feedback to each iteration one at a time. If I skipped an iteration, then my only feedback would be: "good solution and clean code. 👌🏾"

**Here is my feedback to the first iteration:**
good solution but when we are working with automated tests, we don't change the tests to fit our solution. So please make all the work inside the provided function and do not add new modules to it.

Your approach is good. Here is how with small changes on your code you can make the tests pass without adding another function:

```
function getAllDirectors(movies){
//create a new array of directors
  const directorsArray = movies.map(function(movie) {
    return movie.director;
  })
// create a filtered array with no duplicates of the directors
  const filteredDirectorsArr = directorsArray.filter((value, index) => directorsArray.indexOf(value) === index);
  return filteredDirectorsArr;
}
```

Another way to remove duplicates:
`const filteredDirectorsArr = [...new Set(directorsArray)];`

src/movies.js
```diff
@@ -1,25 +1,94 @@
...
+function howManyMovies(arr) {
+  const moviesDirectedBySteven = arr.filter(function(movie){
+      if (movie.director === 'Steven Spielberg'){
+        return movie;
+      }
+    }).filter(function(movie){
+      return movie.genre.includes('Drama')
+    });
+    return moviesDirectedBySteven.length
+    }
```
Farxa, line 26: **Iteration 2:**
Nice job. Here you can use the logical AND (&&) in your if statement and by that combine two if statements into one and with that you don't have to use the second filter().

src/movies.js
```diff
@@ -1,25 +1,94 @@
...
+  })
+  const listOfTitles = sorted.map(movie => movie.title);
+  return listOfTitles.slice(0, 20)
+}
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
-function turnHoursToMinutes() {}
+function turnHoursToMinutes(arr) {
+  return arr.map(movie => {
```
Farxa, line 85: **Iteration 7:**
Your approach here has good elements you can build on it. Try out the following approach if you'd like and if you have any questions let me know please:

//map over the movies array
//create new variable "convertedDuration" with initial value 0
//if movie duration includes 'h' then add the movie duration to the convertedDuration * 60 for the hours (here we need to first slice the 'h' and turn the duration from string into a number)
//if movie duration includes 'min' the we convertedDuration += movie.duration.slice(movie.duration.indexOf(" ") + 1, movie.duration.indexOf('m')); (as you did on line 162, here also turn the duration from string into a number)
//we return a new array using the spread method and replace the value of the "duration" with the convertedDuration like so: return ({...movie, duration: convertedDuration});

PR #2482, opened 2022-01-15
Solved the lab

YoungTaubi:
Jonas Darley Berlin

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,28 +1,176 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(arr) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,28 +1,176 @@
...
+  })
+  return newArr
+}
+
+
+function cleanDirectors(arr) {
+  const cleanArr = arr.filter(function (ele, index) {
+    return arr.indexOf(ele) === index
+  })
+  return cleanArr
```
Farxa, line 16: Hi Jonas,

First of all let me thank you for writing such clean code with good naming that it is easy for me to read and understand. Generally you wrote good code and I have not much to comment.

the function cleanDirectors() is generally correct, it's not doing anything though, because we are not calling it/ using it anywhere. In the case of our lab this function is redundant, all we need is to filter our newArr of directors. 

In general you can always get back to your labs and refactor your code/ try to solve the bonus. If you want to check different ways to solve the iterations, then please get back to the lab solutions Jan provided on GitHub, and if you have any questions, please don't hesitate to let us know. We can go over it together and try to explain it.


Cheers ✌🏾

PR #2483, opened 2022-01-16
Completed all iterations except 8.

Areeba-Imran:
My code for 7th iteration works on codepen. But it's failing some lab tests which is weird. It says 'movie.duration.indexOf' is not a function. Will try to solve it tomorrow along with 8th iteration.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -168,7 +168,7 @@ function turnHoursToMinutes(movies) {
       return movie;
     }
 
-    let totalMins = hours *;
+    let totalMins = hours * 60;
    
     movie.duration = totalMins;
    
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,28 +1,188 @@
...
+  let avgOfDramaMovies = scoresAverage(dramaMoviesArray);
+
+  return avgOfDramaMovies;
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+
+function orderByYear(movies) {
+
```
Farxa, line 83: **Iteration 5:**
good solution 👌🏾. Here is another approach with less code:

```

  const orderMoviesByYear = movies.slice().sort(function(movie1, movie2){

    if (movie1.year !== movie2.year) {
    } else {
    }
  })

  return orderMoviesByYear
}
```
adding slice() before sort() allows us not to mutate the original array

src/movies.js
```diff
@@ -1,28 +1,188 @@
...
+  const orderedArray = orderMoviesByYear.map(function(movie){
+    return movie;
+  })
+
+  return orderedArray;
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+
```
Farxa, line 106: **Iteration 6:**
Just like on iteration 5 you have done a good job writing logic to reach the correct answer. You can refactor your code to make it shorter/ straight to the point. Here us a shorter solution for you to compare:

```

  let copyOfMoviesArr = movies.map(currentMovie => {
      
    return currentMovie.title;

  })
  .sort((movie1, movie2) => 
    movie1.localeCompare(movie2)
  )
  .slice(0, 20);

  return copyOfMoviesArr;
  
}

```

PR #2489, opened 2022-01-16
gratest movies, "solved lab (somehow)" Saleh Ramadan Berlin FT

SalehR89:


Farxa:
Hi Saleh,

There is much to write regarding the feedback on you lab solution, so let's have a talk in person on Monday instead.

Cheers ✌🏾

SalehR89:
Hi Rahaf,

Alright, see you on Monday then!

Salam :)

PR #2528, opened 2022-01-18
BER-FT-012022 Gianluca Francescato

gifrancescato:


Farxa:
Hi Gianluda,

First of all let me thank you for writing good code with good naming that it is easy for me to read and understand. 🙏🏽 You can maybe start using "format document" on your code (right click in vscode and then click on format document). This will format you code even better. You can also use the Prettier extension on vscode that formats your code upon saving.

-- 
For Iteration 2:
here you can use only one filter and using the logical AND (&&) in an if statement and by that combine two filters into one.

If you want to know why some tests are not working or check the solutions for the bonus iterations, then please get back to the lab solutions Jan provided on GitHub, and if you have any questions, please don't hesitate to let us know. We can go over it together and explain it.

PR #2529, opened 2022-01-18
lab solved

HecMarFen:


Farxa:
Hi Hector,

You use good naming and your code is readable 👌🏾

In your orderAlphabetically() function using console.log doesn't do the job, as it only logs the result on the console and doesn't return anything that the tests can recognise. That's why we use return in such cases.

You can always get back to your labs and refactor your code/ try to solve the bonus. If you want to check different ways to solve the iterations, or if you want to know why some tests are not working, then please get back to the lab solutions Jan provided on GitHub, and if you have any questions, please don't hesitate to let us know. We can go over it together and explain it.

Cheers ✌🏾

PR #2578, opened 2022-03-27
almost solved lab

kapix06:


YoungTaubi:
I just had a look at your code. You had a few syntax errors. That's why your test did not work. Next time your test shows a grey background, try to remove all the syntax errors. This way it is easier to keep track of which iterations still need some work...

PR #2608, opened 2022-03-31
WDFT REMOTE MAR2022 - Mafalda 

mafaldao21:


src/movies.js
```diff
@@ -7,9 +7,9 @@ const movies = require("./data");
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(director) {
-  const allDirectors = movies.map( function(director, index) {
-      movies.indexOf(director)=== index;
+function getAllDirectors(movies) {
+  const allDirectors = movies.map( function(director, index, movies) {
+      return movies.indexOf(director) === index;
   });
```
TA-Remote, line 13: With map method, you say what you want to do in each iteration, In our case, you want to return movie director. 

Couple suggestions here:
- firstly the argument you pass into the function, only pass what you will be using. 
` movies.map( function(director, index, movies)`
in our case we only want to access every movie object of that array

- secondly, you only return the value that you want to see, the element you want to see insdted of previous one. So instead full movie object I only want to see movie director.

See result below:

```
const allDirectors = movies.map( function(movie) {
      return movie.director
  });
```

src/movies.js
```diff
@@ -1,17 +1,44 @@
...
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(movies) {
+  let count = 0
+  const spielbergMovies = movies.filter(function(element) {
+    if(movies.includes('Steven Spielberg') && movies.includes('Drama')) {
+      count += 1
+    } 
+  })
```
TA-Remote, line 24: in iteration2 we need to filter by two things: director and genre. 

- we checkif it's drama  by accessing genre array - > as you done
- we check if director is spielberg by extracting movie director and checking if strings are matching

```
if (element.director === 'Steven Spielberg' && element.genre.includes('Drama'))
```

and if we pass the condition -> we `return true `, no need for counter as this method already does that for us :) 

Also if you have notice in my example, I am actually accessing properties of element not the movies. `movies` is your full array, while `element` is the name you gave -> for every element you are checking

TA-Remote (commented):
Hey Mafalda! I have left some comments for you and suggestions how to improve the code (Karina here :) ) , i strongly suggest to have a look at this with fresh head on the weekend. Or look at the solutions on student portal. Things that still don't make sense - make a not and we can review them! Please ask :)

mafaldao21:
Hey, Karina! Hope you are having a wonderful weekend!

Thanks for the feedback! I really struggled with this LAB, and it helped me a lot!

I still don’t quite understand the solution to Iterations 5 and 6, though… Could we make some time on monday to go over it?

Thanks a lot,

Mafalda

PR #2611, opened 2022-03-31
WDFT 2022 ALEJANDRO BLANCO RT march

abg1995:


TA-Remote:
Really good job, Alejandro! In iteration 4, it was only a small thing, let me comment this directly in the code

src/movies.js
```diff
@@ -1,26 +1,73 @@
...
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore() {}
+function dramaMoviesScore(array) {
+  const dramas = array.reduce(function(accumulator, currentValue){
+    return accumulator + currentValue.genre.includes('Drama')},0);
+    return dramas;
+  const dramasAvg = dramas.reduce(function(accumulator, currentValue){
+    return accumulator + currentValue.score},0);
+    const dramaFilt = array.filter(function(movie){return movie.genre.includes('Drama')})
+}
```
TA-Remote, line 46: you can reuse the scoresAverage function instead of writing all the logic again and then it's just the following code for iteration 4:
```
 const dramaFilt = array.filter(function(movie){return movie.genre.includes('Drama')})

return scoresAverage(dramaFilt)
```

PR #2612, opened 2022-03-31
Solved lab

lukecin:


TA-Remote:
Good job, Dani! 
If you have time this weekend, try to solve the rest! :)

src/movies.js
```diff
@@ -1,20 +1,46 @@
...
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(arr) {
+  if (arr.length === 0) {
+    return 0;
+  } else{
+  const scores = arr.reduce(function(accumulator, currentValue) {
+    return accumulator + currentValue.score
+  }, 0);
```
TA-Remote, line 30: the one test, that is failing here, you could have solved with an if condition like this:
```
 const scores = arr.reduce(function(accumulator, currentValue) {
    if (currentValue.score){
      return accumulator + currentValue.score
    }
    else {
      return accumulator
    }
  }, 0);
```

PR #2616, opened 2022-03-31
WDFT - MARC2022 - RUBEN POELEN

coderKron:
I'm tired and deserve rest. :(

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,77 @@
 // The `movies` array from the file `src/data.js`.
-console.log('movies: ', movies);
+// console.log('movies:', movies);
 
+// const movies = require('./data');
 
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+// const arrayDirector = [];
+function getAllDirectors(arrOfMovies) {
+  return (arrayDirector = arrOfMovies.map(function (element, value) {
+    // console.log(element);
+    return element.director;
+    // console.log(element);
+  }));
+  // return arrayDirector;
+}
+// getAllDirectors(movies);
+// console.log(arrayDirector);
+
+// const directorArray = getAllDirectors(movies);
+// console.log(directorArray);
+// console.log(arrayDirector);
+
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,77 @@
...
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+// const arrayDirector = [];
+function getAllDirectors(arrOfMovies) {
+  return (arrayDirector = arrOfMovies.map(function (element, value) {
+    // console.log(element);
+    return element.director;
+    // console.log(element);
+  }));
```
TA-Remote, line 15: I like that you are returning result straight away! same as in below iterations :)

src/movies.js
```diff
@@ -1,22 +1,77 @@
...
+  }
+  return (spielbergMovies = array.filter(function (element, value) {
+    if (
+      element.genre.includes('Drama') &&
+      element.director.includes('Steven Spielberg')
+    ) {
+      return true;
+    } else {
+      return false;
+    }
```
TA-Remote, line 41: with filter method you do not need else (or return false)

If your condition passess -> filter will save those elements, if not passing it will ignore

src/movies.js
```diff
@@ -1,22 +1,77 @@
...
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(array) {
+  if (array.length === 0) {
+    return 0;
+  }
+  const totalScore = array.reduce(function (accumulator, currentValue) {
+    return accumulator + currentValue.score;
+  }, 0);
```
TA-Remote, line 56: here you need additional check, to see if the movie object has `score`, if it doesn't you skip that object , and only return accumulator

src/movies.js
```diff
@@ -1,22 +1,77 @@
...
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore() {}
+function dramaMoviesScore(array) {
+  const dramaMovies = array.filter(function (element, value) {
+    if (element.genre.includes('Drama')) {
+      return true;
+    }
+  });
+  return scoresAverage(dramaMovies);
```
TA-Remote, line 71: reusing previous function - great!

PR #2617, opened 2022-03-31
WDFT March2022Remote Sepideh.att

Sepidehatt:


src/movies.js
```diff
@@ -1,29 +1,114 @@
...
+
+  if(!moviesArr.length) return 0 ;
+  else{
+    const sumScores = moviesArr.reduce((a,b)=>{
+      if(!b.score){
+        b.score = 0;
+        return a + b.score ;
+      }else{
+        return a + b.score ;
+      }
```
TA-Remote, line 37: if `b.score` is undefined/does not exist - it is enough to` return a`, will be shorter syntax.

```
      if(!b.score){
        return a ;
      }else{
        return a + b.score ;
      }
```

A challenge for you as a **bonus** -> read about **ternary operator,** and try to convert the if statement to new syntax . If not clear let me know!
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

`condition ? exprIfTrue : exprIfFalse`

TA-Remote (commented):
great work Sepid

Karina.

PR #2618, opened 2022-03-31
[Remote-FT-032022] Thibaut LARTIGUE 

thibautDevWorld:
Iteration 5, I'm not sure of my solution to filter the same year and class them.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,23 +1,59 @@
 // The `movies` array from the file `src/data.js`.
-console.log('movies: ', movies);
+//console.log('movies: ', movies);
 
 
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,23 +1,59 @@
...
+function orderByYear(moviesReferences) {
+  const moviesOrdered = moviesReferences.sort(function(a, b){
+    if(a.year === b.year){
+      return a.title.localeCompare(b.title);
+      } else {
+        return a.year - b.year;
+     }
+  })
+  return moviesOrderedCopy = moviesOrdered.map((x) => x);
+}
```
TA-Remote, line 56: In iteration5, better approach is to copy the array in the beginning, and only then work on that copy. So this could be a better approach

```
function orderByYear(moviesReferences) {

  const copiedArray = [...moviesReferences]

  const moviesOrdered = copiedArray.sort(function(a, b){
    if(a.year === b.year){
      } else {
        return a.year - b.year;
     }
  })
  return moviesOrdered 
}
```

TA-Remote (commented):
Great work Thibaut!

PR #2620, opened 2022-04-01
[FTRMT032022] Oscar Gonzalez

oscargonzalezdev:
- [x] solve lab javascript greatest movies

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,26 +1,80 @@
 // The `movies` array from the file `src/data.js`.
-console.log('movies: ', movies);
-
+// console.log('movies: ', movies);
 
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+
+function getAllDirectors(movies) {
+  const directors = movies.map(item => {
+    return item.director;
+  });
+  return directors;
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(movies) {
+  const times = movies.filter(item => item.director === 'Steven Spielberg');
+return times.length;
+}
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,26 +1,80 @@
...
+  });
+  return directors;
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies() {}
+function howManyMovies(movies) {
+  const times = movies.filter(item => item.director === 'Steven Spielberg');
+return times.length;
+}
```
TA-Remote, line 19: Here you are asked to filter by two factors, by director and by genre , so missing one more check ;)

src/movies.js
```diff
@@ -1,26 +1,80 @@
...
+    }, 0);
+    let average = Math.round((sum/movies.length) * 100) / 100;
+    return average;
+  }
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore() {}
+function dramaMoviesScore(movies) {
+  let dramaMovies = movies.filter(item => item.genre[0] === 'Drama');
+  return scoresAverage(dramaMovies);
```
TA-Remote, line 46: Great - you are reusing the function from before

src/movies.js
```diff
@@ -1,26 +1,80 @@
...
+
+  movies.forEach(item => {
+    if (typeof item.score != 'number'){
+      item.score = 0;
+    };
+  });
+
+  if (movies.score){
+    movies.score = 0;
+  }
```
TA-Remote, line 35: in your function `movies` is an array, so lines 33-35 makes no sense, cause array of movies does not have property `score`.. Only objects have properties

You can just remove those lines - makes no difference to the code

TA-Remote (commented):
Great work Oscar. You can review suggested solution for iteration6 in student portal :)

src/movies.js
oscargonzalezdev: I see. Thanks!

src/movies.js
oscargonzalezdev: I see the mistake ;)

PR #2621, opened 2022-04-01
resolved 3 questions 

sina-kaabi:


sina-kaabi:
focused more hours on studying the psuedo code rather than completing iteration 3 to understand better.

PR #2624, opened 2022-04-01
WDFT Mar2022 Pauline Schnoebelen

Paulinettt:


src/movies.js
```diff
@@ -1,26 +1,84 @@
...
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage() {}
+function scoresAverage(array) {
+  if (array.length === 0) return 0;
+  
+
+  const countScores = array.reduce(function(accumulator, currentValue) {
+    return accumulator + currentValue.score;},0);
```
TA-Remote, line 32: Here you are missing one little step. Some movies in data do not have score. They have property name , but not the value. 

so indide reduce method you want to check if `currentValue.score `exists
  - if it does `return accumulator + currentValue.score`
  - if it does not` return accumulator`

src/movies.js
```diff
@@ -1,26 +1,84 @@
...
+
+const result = countScores * 1;
+const roundedResult = result / array.length;
+return Math.round (roundedResult * 100) / 100;
+
+//I was struggling to round to 2 decimals. This is the only way I managed to make it work ^^
```
TA-Remote, line 38: That's definitely one way to do it!

Alternative way could be this:
- toFixed() method rounds the number but converts it to string
- Number() method converts string to number

`return Number(roundedResult.toFixed(2));`

PR #2627, opened 2022-04-01
WDFT MARCH2022 Valentin Paul

Valentin-Paul:


src/movies.js
```diff
@@ -1,23 +1,70 @@
...
+      movie.genre.includes('Drama') === true
+    ) {
+      return true;
+    }
+  });
+  let counter = 0;
+  for (let i = 0; i < spielbergMovies.length; i++) {
+    counter += 1;
+  }
+  return counter;
```
TA-Remote, line 30: This code is actually unnessary. Remember, you already have array with filtered Spielberg movies. So we could just count how many movies are in that array -> return `spielbergMovies.length`

TA-Remote (commented):
Great start Valentin! I suggest either to work on this a bit further on the weekend or look over the suggested solutions for remaining iterations 

Will be uploading on student portal later today.

PR #2634, opened 2022-04-14
[FTLIS042022] Chloé Faurie

chloe4E:
solved all iterations

I tried the #7, seems OK though it does not pass all the tests - not sure why :)

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,34 +1,245 @@
 // The `movies` array from the file `src/data.js`.
-console.log('movies: ', movies);
+//console.log('movies: ', movies);
 
+// {
+//   "title":"The Shawshank Redemption",
+//   "year":1994,
+//   "director":"Frank Darabont",
+//   "duration":"2h 22min",
+//   "genre":["Crime","Drama"],
+//   "score":9.3
+// }
 
 // Iteration 1: All directors? - Get the array of all directors.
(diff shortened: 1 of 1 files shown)


webLisbonTA (commented):
Great job on finishing all the mandatory iterations!

src/movies.js
```diff
@@ -1,35 +1,246 @@
...
+//         score: 9.7
+//       }
+//     ])
+// );
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore() {}
+function dramaMoviesScore(arr) {
+  let dramas = arr.filter((element) => element.genre.includes('Drama'));
+  return scoresAverage(dramas);
```
webLisbonTA, line 86: Amazing job using the previous function to ease your work!

src/movies.js
```diff
@@ -1,35 +1,246 @@
...
+    let adate = a.year;
+    let bdate = b.year;
+    let rv = adate - bdate;
+    if (rv === 0) {
+      rv = a.title.localeCompare(b.title);
+    }
+    return rv;
+  });
+  return sortedArr;
+}
```
webLisbonTA, line 142: This solution is correct but you could do it with the same logic but a bit shorter:
```
  const ordered = [...movies].sort((a, b) => {
    if (a.year !== b.year) {
      return a.year - b.year;
    }
  })
  return ordered;
}
```

chloe4E: smart

PR #2643, opened 2022-04-14
[FTWD042022] Catarina Salvado

catarinapsalvado:


webLisbonTA (commented):
Great job on finishing the mandatory iterations!

src/movies.js
```diff
@@ -1,26 +1,90 @@
...
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear(array) {
+  let arrayNew = [];
+  array.sort((a, b) => {
+    if (a.year < b.year) return -1;
+    if (a.year > b.year) return 1;
+    if (a.year === b.year) return 0
+    if (a.title === b.title) return -1;
```
webLisbonTA, line 67: This function is not working. Try and use localeCompare when ordering things by name. Also, for numeric sorting try and use a.something - b.something

PR #2645, opened 2022-04-14
[FTLIS042022] Mariana Oliveira

miloliveira:
lab solved, without the bonus

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,34 +1,103 @@
 // The `movies` array from the file `src/data.js`.
-console.log('movies: ', movies);
+//console.log('movies: ', movies);
 
+const movies = require('./data');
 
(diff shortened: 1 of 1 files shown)


webLisbonTA (commented):
Great job on finishing all the mandatory iterations!

src/movies.js
```diff
@@ -1,35 +1,104 @@
...
-function scoresAverage() {}
+function scoresAverage(array) {
+  //let sum = 0;
+  // return array.map((movie) => sum += movie.score)/movie.score.length
+  if (array.length === 0) {
+    return 0;
+  }
+  let sum = array.reduce(function (acc, movie) {
+    if (movie.score === 0) {
+      acc + 0;
```
webLisbonTA, line 34: No need for the if statement. If the value is 0 the reduce still calculates it

PR #2648, opened 2022-04-14
[FTLIS042022] Francisco Bento

FranciscoGBento:


webLisbonTA (commented):
Good job on finishing all the mandatory iterations!

src/movies.js
```diff
@@ -1,26 +1,78 @@
...
+
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear(arr) {
+  let newArr = [];
+  const ordered = arr.sort(function(a, b) {
+  if (a.year < b.year) return -1;
+  if (a.year > b.year) return 1;
+  if (a.year === b.year) return 0;
```
webLisbonTA, line 53: this function is not working, try and use localeCompare for ordering alpha and a.something - b.something to sort by number

src/movies.js
```diff
@@ -1,26 +1,78 @@
...
+   // a.title.localeCompare(b.title) ??
+  })
+  return ordered
```
webLisbonTA, line 56: You should return a new array, you are returning the array passed as a parameter of your function

PR #2649, opened 2022-04-14
[FTLIS042022] Alexandre Alves

aletrad:


webLisbonTA (commented):
Great job on finishing the whole lab!

src/movies.js
```diff
@@ -1,35 +1,81 @@
...
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear(arr) {
+  let byYear = arr.map((movie) => {
+    return movie;
+  });
+  byYear.sort((a, b) => {
+    if (a.year < b.year) return -1;
+    if (a.year > b.year) return 1;
+    if (a.year === b.year) return a.title.localeCompare(b.title);
```
webLisbonTA, line 56: The logic here is good but you could refactor it for a cleaner code:
The "localeCompare" already returns -1/0/1 so the first and second if statements are redundant

PR #2651, opened 2022-04-14
[FTLIS042022] anaPinho

AnaGonPinho:


webLisbonTA (commented):
Good job on finishing all mandatory iterations!

src/movies.js
```diff
@@ -1,35 +1,80 @@
...
+}
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+function orderByYear(array) {
+  let arrayNew = [];
+  array.sort((a, b) => {
+    if (a.year < b.year) return -1;
+    if (a.year > b.year) return 1;
+    if (a.year === b.year) return 0;
+  });
```
webLisbonTA, line 56: Here try and use the a.something - b.something for the sort, furthermore, you should use localeCompare instead of returning -1/0/1

PR #2656, opened 2022-04-14
[FTLIS042022] matildeCosta

matildedacosta:
I had a hard time with bonus 7 and 8! Tried some ideas, but none were working. (for bonus 7 I was having a hard time figuring out how to transform the strings into numbers and then the numbers into actual 'hour values'; for bonus 8 I think I got confused with what was actually being asked so my solution ideas weren't very clear)

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,34 +1,135 @@
 // The `movies` array from the file `src/data.js`.
-console.log('movies: ', movies);
-
+//console.log('movies: ', movies);
+const movies = require('./data.js');
 
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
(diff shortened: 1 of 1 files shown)


webLisbonTA (commented):
Great job on finishing all the iterations and giving a shot at the bonus!

src/movies.js
```diff
@@ -1,34 +1,216 @@
...
+/* Go ahead and create a howManyMovies() function that receives an array as a parameter and filter 👀 
+the array so we can have only the drama movies where Steven Spielberg is the director. */
+
+function howManyMovies(array) {
+  const directorMovies = array.filter(
+    (element) => element.director === 'Steven Spielberg'
+  );
+  const dramaMovies = directorMovies.filter((element) =>
+    element.genre.includes('Drama')
+  );
```
webLisbonTA, line 63: You can do this in the same filter next time for a cleaner code.

PR #2661, opened 2022-04-18
Solved lab

nataliabuzila:
please let me know why the following tests failed:
1. Iteration 3 function scoresAverage
    √ should be declared
    × should return a number (3 ms)
    ×  should return the average score of 2 movies with score 8 each (1 ms)        
    × should be rounded to 2 decimals places
    √ should return 0 if an empty array is passed
    × should return average even if one of the movies does not have score (1 ms)
2. Iteration 6 function orderAlphabetically, 2nd test: should return an array
    √ should be declared
    × should return an array (1 ms)
    √ should not mutate the original array
    √ should only return the title of the movies, each value should be a string    
    √ should return all of items when the array passed has fewer than 20 items     
    √ If there are more than 20 elements, return only 20 of them. (1 ms)
    √ should order them alphabetically.
    √ should return the top 20 after ordering them alphabetically.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,26 +1,159 @@
 // The `movies` array from the file `src/data.js`.
-console.log('movies: ', movies);
+//console.log('movies: ', movies);
+
+//const movies = require("./data");
+
 
 
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors() {}
+function getAllDirectors(arrayMovies) {
+let newArrayDirectors = arrayMovies.map(function (array) {
+return array.director;
+})
+return newArrayDirectors;
+}
+//console.log (getAllDirectors(movies));
+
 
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,26 +1,164 @@
...
+  if(array.length===0) return 0;
+    else {
+      let numberOfScores=0;
+      const sumScores = array.reduce (function (accumulator, movie) {
+        if(movie.score) {
+          numberOfScores++;
+          return accumulator + movie.score;
+        } else return accumulator;
+      }, 0);
+      return (sumScores / numberOfScores).toFixed(2);
```
Matteo-Ironhack, line 55: Definition of `toFixed` method:
Returns a string representing a number in fixed-point notation.
you needed to convert it back to a number

src/movies.js
```diff
@@ -1,26 +1,164 @@
...
+  });
+  return arrayCopy;
+}
+//console.log(orderByYear(movies))
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically() {}
+
+function orderAlphabetically(arrayMovies) {
+  if(arrayMovies.length===0) return 0;
```
Matteo-Ironhack, line 131: why would this return 0?
the subject states that the function `receives an array and returns an array`; since this returns a number, it is invalid

PR #2664, opened 2022-04-19
solve lab greatest movies

ihab5520:


src/movies.js
```diff
@@ -1,32 +1,112 @@
...
+    return movie.genre.includes('Drama');
+  });
+  return scoresAverage(dramaAvg);
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear() {}
+//a.title.toLowerCase().localeCompare(b.title.toLowerCase())
+function orderByYear(arr) {
+  const sortMoviesByYear =[arr];
```
Matteo-Ironhack, line 52: this add the original array into a new one, so you basically end up having
```javascript
[
  [
    {
      "title":"Blade Runner"
    },
    // all the other movies here
  ]
]
```
in order to succeed in what you had to do, you either needed it to copy in some way, or use destructuring (we looked at it briefely in class, it's the three dots `...`

PR #2673, opened 2022-05-07
fix: fixing a ReferenceError on movies variable

abernier:
`movies` will be passed as 1st parameter of the different functions by the tests, but the variable does not exist here


ross-u:
@abernier The variable `movies` is available in the global scope and comes from the file `src/data.js` that is loaded in the `index.html`:
> ```js
>   <script src="./src/data.js"></script>
>   <script src="./src/movies.js"></script> 
>```

PR #2742, opened 2022-06-20
[RMFT JUNE 22] Gabriel Gomez

gabriel-vcs:


TA-Remote:
Great Job Gabriel!
just a minor thing

src/movies.js
```diff
@@ -1,30 +1,79 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray = []) {
+  const allDirectors = moviesArray.map((movie) => movie.director);
+  return allDirectors.filter(
```
TA-Remote, line 6: this filter method return the same Array as allDirectors, because the condition that you write is always true!
 you need to make sure if the specific director's name is unique in this array or not ;)

gabriel-vcs: Sorry are you sure? The condition will be only true if the element found matches the first element found by indexOf(), if the element is repeated, indexOf() will alway return the first element found and will not match the position in the array, so the condition will be false.

TA-Remote: you are right Gabriel ! I checked it in a wrong way.
thank you for correcting me

PR #2746, opened 2022-06-20
Begüm Eryıldız WDFTJune2022

begumeryildiz:
I solved five iterations and half of iteration 6. I will continue to solve the rest of them when I feel better.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,63 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+
+const movies = require("./data");
+
+function getAllDirectors(moviesArray) {
+  return moviesArray.map(movie => movie.director);
+};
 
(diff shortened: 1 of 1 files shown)


TA-Remote:
Good Job Begüm!

src/movies.js
```diff
@@ -1,22 +1,63 @@
...
+  });
+  return copyMoviesArray;
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+  const copyMoviesArray = [...moviesArray];
+  copyMoviesArray.sort(function(aMovie, anotherMovie) {
+    if (aMovie.title > anotherMovie.title) {
```
TA-Remote, line 54: in this case, you need to make sure about case sensitive! 
also you need to return firs 20 title as an array, 
you can write it like this :

```
let result = copyMoviesArray.sort(function(aMovie, anotherMovie) {
 if (aMovie.title.toLowerCase() > anotherMovie.title.toLowerCase()) {
 return 1;
    } else  if (aMovie.title.toLowerCase() < anotherMovie.title.toLowerCase()) {
  return -1
  } else {
  return 0
  }
  }) 
  .slice(0, 20);
  
return result;
```

PR #2749, opened 2022-06-20
[RM FT June 22] Alex's Lab solved 

Virgule09:
I finished the lab until lab 4 as the others were optional. I thought I would have more time this evening to do more, but I didn't. I will come back to it this week end or when I have some time.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,46 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
+
+const movies = require("./data");
+
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  const allDirectors = moviesArray.map(function(movies){
+      return movies.director
+  });
+  return allDirectors
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+  const spielbergDramas = moviesArray.filter(function(movies) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,46 @@
...
+    return (movies.director === 'Steven Spielberg' && movies.genre.includes('Drama'))
+  })
+  return spielbergDramas.length
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  const totalScore = moviesArray.reduce(function(accumulator, currentValue) {
+    return accumulator + currentValue.score
```
TA-Remote, line 25: some of movies don't have any score. because of that, you should check if movie has a score or not:

if(currentValue.score) {
      return  accumulator + currentValue.score
    }else {
      currentValue.score = 0
      return  accumulator + currentValue.score
    }

src/movies.js
```diff
@@ -1,16 +1,46 @@
...
+  const totalDramaScore = moviesArray.reduce(function(accumulator,currentValue){
+    if (currentValue.genre.includes(`Drama`)){
+      counter++
+      return accumulator + currentValue.score
+    }
+    else {
+      return 0
+    }
+  },0)
+  return Math.round(Math.round((totalDramaScore / counter) * 100)/100)
```
TA-Remote, line 42: here you need to round number just once! ;)

src/movies.js
```diff
@@ -1,16 +1,46 @@
...
+    return accumulator + currentValue.score
+  },0)
+  return Math.round((totalScore / moviesArray.length) * 100) / 100
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
+  let counter = 0;
+  const totalDramaScore = moviesArray.reduce(function(accumulator,currentValue){
```
TA-Remote, line 33: you can Filter drama movies and after that reduce it ,
make sure you check is score exist or not ;)

PR #2750, opened 2022-06-20
WD WF June 2022 Matthieu DUPERREY

MatthieuDUPERREY:


src/movies.js
```diff
@@ -1,13 +1,34 @@
...
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+
+  function scoresAverage(movies) {
+    if (movies.length === 0) {
+      return 0
+    }
+    const scores = movies.map(movie => movie.score)
+    const sum = scores.reduce((accumalator, currenValue) => {
+      return accumalator + currentValue
```
TA-Remote, line 26: Here you made a type at `currenValue` and `currentValue`

src/movies.js
```diff
@@ -1,13 +1,34 @@
...
+    })
+    const average = sum / scores.length
```
TA-Remote, line 28: If you `return`  the average you'll get more tests passing, just needing to make some adjustments for decimals and some scores that doesn't have a value

src/movies.js
```diff
@@ -1,13 +1,34 @@
...
+function howManyMovies(movies) {
+  const spielbergMovies = movies.filter((element) => element.director === 'Steven Spielberg' && element.genre.indexOf ('Drama') > -1);
+  console.log(howManyMovies);
+  return spielbergMovies.length;
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+
+  function scoresAverage(movies) {
```
TA-Remote, line 20: Just an example:

```
  function scoresAverage(movies) {
    if (movies.length === 0) {
      return 0
    }
    const scores = movies.map(movie => movie.score)
    const sum = scores.reduce((accumalator, currentValue) => {
      if (currentValue && typeof currentValue === 'number'){
        return accumalator + currentValue
      }
      return accumalator
    })
    const average = sum / scores.length
    return parseFloat(average.toFixed(2))
  }
```

Take your time to understand the changes and if needed ask for help

TA-Remote (commented):
- Good job!

- Try pushing the other iterations and the bonus as well, they are really good at understanding data structures and will be useful in the future.

PR #2754, opened 2022-06-21
Khalid918

ghost:
there is something wrong with iteration 5 solution but I can not figure it out, also
line 32 when I try to use .toFixed() it gives me failed test so I searched google and found this way but I do not quite understand it .

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,19 +1,59 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  return moviesArray.map(function(movie){
+    return movie.director 
+  })  
+  //return myDirectors
+} 
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
-
+function howManyMovies(moviesArray) {
+  const result1 = moviesArray.filter(function(movie0){
+    return movie0.director === 'Steven Spielberg' && movie0.genre.includes('Drama')
+  });
+  return result1.length
+}
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,19 +1,59 @@
...
+  }
+  const totalScores = moviesArray.reduce(function(accu, movie){
+    if (movie.score > 0) {
+      return accu + movie.score;
+    } else  {
+      return accu ;
+    }
+  }, 0);
+  const average = totalScores / moviesArray.length;
+  const decimal = (value) => Math.round(value * 100) / 100;
```
TA-Remote, line 32: `Math.round()` will take a value and round it to the closest integer, that is been used just to transform the string into a number. When you  `*100` and after `/100` you get the decimals value.  So If you get `Math.round( 6.96 * 100)` you get 696 and if we get `696 /100` we would get the 6.96 as a number  

```
let score = "6.96"

console.log(Math.round(score)) // 7
console.log(Math.round(score * 100)) // 696
console.log(Math.round(score * 100) / 100) // 6.96
```

src/movies.js
```diff
@@ -1,19 +1,59 @@
...
+  const dramaMovies = moviesArray.filter(function(movie){
+    return movie.genre.includes('Drama');
+  });
+  average = scoresAverage(dramaMovies);
+  return average;
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
```
TA-Remote, line 46: I made some modifications in your code and get the tests passing, but the following it's just and example:

```
  const copy = [...moviesArray];
  copy.sort(function(element1, element2){
    if (element1.year === element2.year) {
      let titleA = element1.title.toUpperCase();
      let titleB = element2.title.toUpperCase();
      if (titleA < titleB) return -1;
      if (titleA > titleB) return 1; // here you can check at MDN why this code is return -1 or 1 
    }
    return element1.year - element2.year
  })
  return copy
}
```
 This is a shorter solution:
 
```
  
  return [...moviesArray].sort((a, b) => {
    
    if (a.year !== b.year) {
      return a.year - b.year;
    }

    return a.title.toLowerCase().localeCompare(b.title.toLowerCase());
  });
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

TA-Remote (commented):
- Good job!

- Try keep practicing as much as you can and code will come more naturally

PR #2755, opened 2022-06-21
Nic Solitom - Solved lab

nicsolitom:


src/movies.js
```diff
@@ -1,30 +1,110 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  let allDirectors = moviesArray.map((movie) => {
+    return movie.director;
+  });
+  return allDirectors;
+}
```
TA-Remote, line 9: For shorter syntax:

1. when use use arrow function and have only one line of code to return (literally) , you can use below syntax:
```
  let allDirectors = moviesArray.map((movie) => movie.director);
```

2. You can return the result directly. If you are not going to use the variable  `allDirectors`, no need to actually declare it
```
return moviesArray.map((movie) => movie.director);
```

src/movies.js
```diff
@@ -1,30 +1,110 @@
...
+  }).length;
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  if (moviesArray.length > 0) {
+    let totalScore = moviesArray.reduce((accumulator, currentValue) => {
+      return accumulator + currentValue.score;
+    }, 0);
```
TA-Remote, line 25: You missed little step. First you want to check if `currentValue.score` exists - before summing it

src/movies.js
```diff
@@ -1,30 +1,110 @@
...
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
-function turnHoursToMinutes(moviesArray) {}
 
-// BONUS - Iteration 8: Best yearly score average - Best yearly score average
-function bestYearAvg(moviesArray) {}
+// Nic: I did not finish this one, but I did manage to turn hours into minutes, 
+// but it ended up manipulating the original array.. :)  
+function turnHoursToMinutes(moviesArray) {
+    let resultArray = [];
+  moviesArray.forEach((movie) => {
```
TA-Remote, line 83: In this iteration, you firstly want to make deep copy of the array, before you iterate over it.

One approach to make deep copy is this:
```
const newMoviesArry = JSON.parse(JSON.stringify(moviesArray))
```

https://www.freecodecamp.org/news/how-to-clone-an-array-in-javascript-1d3183468f6a/

TA-Remote (commented):
Really well done, added little notes to improve the code

PR #2756, opened 2022-06-21
Ivona Pepic Lab greatest movies done

IvonaPep:


TA-Remote:
you wrote some console.log() and variable decleration outside functions, and because of that, some of tests didnt pass ,
when we have automated test in Labs, writing any code outside functions break the tests :)

src/movies.js
```diff
@@ -1,22 +1,78 @@
...
+
+console.log(howManyMovies(movies));
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+if (moviesArray.length === undefined){
+  return 0;
+} else {
+  let score = moviesArray.reduce( function(acumulator, currentValue){
```
TA-Remote, line 33: you need to check if score exist for every movies or not :
inside reduce method you can write:
 if(!currentValue.score){
        currentValue.score = 0;
        return acumulator  + currentValue.score ;
      }else{
        return acumulator  + currentValue.score ;
      }

src/movies.js
```diff
@@ -1,22 +1,78 @@
...
+    return acumulator + currentValue.score;
+  },0);
+  
+ return score/moviesArray.length.toFixed(2);
```
TA-Remote, line 37: toFixed method return string instead of number. so for changing type of result to number , this syntax can be one option :

`Number((score/moviesArray.length).toFixed(2))`

PR #2757, opened 2022-06-21
Lara Rockenstein WD FT June22

LMRBLN:


src/movies.js
```diff
@@ -1,28 +1,110 @@
...
+    element.duration =
+      parseInt(element.duration.slice(0, 1)) * 60 +
+      parseInt(element.duration.slice(3, 5));
+  });
+  return moviesArrayCopy;
+}
 
 // BONUS - Iteration 8: Best yearly score average - Best yearly score average
-function bestYearAvg(moviesArray) {}
+
```
TA-Remote, line 79: I think that in this iteration would be better try to maintain the relationship of years and scores in the same element

TA-Remote (commented):
- Great job!

- Great names for variables 
- Clean code :)  Just try to reuse your code wherever is possible :)

PR #2759, opened 2022-06-21
Stephanie Tovar - Lab done

Stephanie-Tovar:


src/movies.js
```diff
@@ -1,28 +1,70 @@
...
+const movies = require("./data");
+
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  const directors = moviesArray.map(function (element) {
+    return element.director;
+  })
+  return directors;
+}
```
TA-Remote, line 12: For an extra challenge try converting this to an arrow function, this way you can get shorter syntax instead of lines 8-10, you can have it in one ;)

src/movies.js
```diff
@@ -1,28 +1,70 @@
...
+  return dramaMovies.length;
+}
+
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  if (moviesArray.length === 0) {
+    return 0;
+  }
```
TA-Remote, line 31: I love that you distinctly separate edge cases - makes it easier to read code

src/movies.js
```diff
@@ -1,28 +1,70 @@
...
+  const average = (Math.round((sum / moviesArray.length) * 100) / 100);
+  return average;
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
+  if (moviesArray.length === 0) {
+    return 0;
+  }
```
TA-Remote, line 43: For shorter syntax, if you return only one line (literally) , you can use below syntax: 

```
```

TA-Remote (commented):
Great! 

If you'll find time, make some more progress on further interations. The more time you put in - the better you'll get in the long run :)

PR #2760, opened 2022-06-21
all_without_bonus

Christoph86:
1-6; no bonus;


src/movies.js
```diff
@@ -1,28 +1,80 @@
...
+  return uniqueMovArr
+}
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+  if (moviesArray.length === 0) {return 0;}
+
+  return moviesArray.filter( (element) => element.director === 'Steven Spielberg')
```
TA-Remote, line 20: Here you can check 2 condition in one filter method ;)
`moviesArray.filter( (element) => element.director === 'Steven Spielberg' && element.genre.includes("Drama") )`

PR #2782, opened 2022-07-11
[FTLIS072022] Raquel

raqpoletto:
missing #6


src/movies.js
```diff
@@ -1,19 +1,76 @@
...
+  return moviesBySpielberg.length;
+}
+console.log(howManyMovies(movies));
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  const totalScore = moviesArray.reduce((acc, value) => {
+    return acc + value.score;
+  }, 0);
```
webLisbonTA, line 25: Here you were just missing a conditional to check if the score exists or not:
```
const totalScore = moviesArray.reduce((acc, value) => {
    if (value.score) {
         return acc + value.score;
     } else {
          return acc;
     }
  }, 0);
```

webLisbonTA (commented):
Good job in the 4 iterations!

PR #2841, opened 2022-08-03
js-movies-methods-w4

jdlms:


src/movies.js
```diff
@@ -1,25 +1,88 @@
...
-function turnHoursToMinutes(moviesArray) {}
+function turnHoursToMinutes(moviesArray) {
+  let lengthInMin = moviesArray.map((old) => {
+    let newDuration = hoursToMin(old.duration);
+    let x = old.duration.replace(old.duration, newDuration);
+    return x;
+  });
+
+  function hoursToMin(x) {
+    let duration = Number(x.replace(/\D/g, ""));
```
trebeljahr, line 73: this can be much much more elegant. I like your solution but if you use regexes - use them with their full powaaaah! :joy: 

```javascript
return moviesArray.map((movie) => {
    const hours = movie.duration.match(/(?<hours>\d+)h/)?.groups?.hours || "0";
    const mins = movie.duration.match(/(?<mins>\d+)min/)?.groups?.mins || "0";
    return { ...movie, duration: parseInt(hours) * 60 + parseInt(mins) };
  });
```

trebeljahr: now this might look a bit like chinese but let me explain...
first we have these two:
`/(?<hours>\d+)h/`
and 
`/(?<mins>\d+)min/`
they use a regex feature called "named capturing groups", which look like this: `(?<name> ...)`
docs are here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Backreferences

now when we use match from regex, this will create a key under the value of groups for the match of the <name> and assign the matched string to it. 

so these two search for d+ (any amount of digits) followed by h and mins respectively, then put the digits they matched into groups named "hours" or "mins" respectively, which then can be accessed via .groups.hours and .groups.mins

the extra ? marks in there are to make sure, that if at any step along we have an undefined -> i.e. there was no match for whatever reason -> or no groups -> we don't throw an error when we further chain . -> undefined?.mins can be executed and returns undefined. undefined.mins would throw an error. the last part is the || if the thing before the || returns undefined, for whatever reason -> use "0" as the "default" time string for hours/minutes. This could be the case if the string to parse would be say "2h" 

lastly, create a new object, spread over ...movie to copy all of the existing key/value pairs into the object, finally overwrite value with parseInt(hours) * 60 + parseInt(mins) -> because the regex matches / "0" defaults are still strings, we have to do the conversion.

trebeljahr (commented):
This review took me waaay too long to write, but building a proper regex solution that handles all the cases nicely is a bit of a pain ^^ 

Good job on the lab. I have nothing super drastic to add, just a few things here and there, where things could be shorter or more concise and a few corrections where you do things that don't make too much sense.

src/movies.js
trebeljahr: image

trebeljahr: 1 thing that really helps when building and testing regexes is this website - make sure to put it in JS (EcmaScript) mode
https://regex101.com/

trebeljahr: even better is a solution I saw from Patrick. because it doesn't have regexes ;) 
```js 
const arr = duration.split(" ");
return (arr.length === 1) ? 60 * parseInt(arr[0]) : 60 * parseInt(arr[0]) + parseInt(arr[1]);
```
split at the space, then do stuff depending on the number of splits you got. :)

src/movies.js
```diff
@@ -1,25 +1,88 @@
...
+      return Number(split[0] * 60) + Number(split[1]);
+    } else if ((split.length = 3)) {
+      return Number(split[0] * 60) + Number(split[1]) + Number(split[2]);
+    }
+  }
 
+  return lengthInMin;
+}
 // BONUS - Iteration 8: Best yearly score average - Best yearly score average
 function bestYearAvg(moviesArray) {}
```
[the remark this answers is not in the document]
trebeljahr: one solution would be:
```js
function bestYearAvg(moviesArray) {
  if (!moviesArray.length) return null;

  const moviesByYear = moviesArray.reduce((groups, movie) => {
    if (!groups[movie.year]) {
      return { ...groups, [movie.year]: [movie] };
    }
    groups[movie.year].push(movie);
    return groups;
  }, {});

  const bestYear = Object.entries(moviesByYear).reduce(
    (best, [year, moviesForThatYear]) => {
      const score = scoresAverage(moviesForThatYear);
      if (score >= best.score) {
        const tiedAndCurrentIsOlder = best.score === score && best.year < year;
        return tiedAndCurrentIsOlder ? best : { year, score: score };
      }

      return best;
    },
    { score: -Infinity, year: Infinity }
  );

  return `The best year was ${bestYear.year} with an average score of ${bestYear.score}`;
}
```

src/movies.js
```diff
@@ -1,25 +1,88 @@
...
+      movie.director === "Steven Spielberg" && movie.genre.includes("Drama")
+    );
+  });
+  return spielberg.length;
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+
+//is there a way to do this with only 1 method? I couldn't figure it out.
```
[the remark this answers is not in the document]
jdlms: I have to look into what { } around params does because I've forgotten.

src/movies.js
jdlms: Thanks for this! I saw Patrick's solution too. Much better!

PR #2845, opened 2022-08-06
Done iteration 2, still can't solve iteration 3 and working on iterat…

JoaoSerafim3001:
…ion 5

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,25 +1,48 @@
+
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+    const allDirectors = moviesArray.map(movie => movie.director)
+    return allDirectors
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+    const dramaMovies = moviesArray.filter(movie => movie.director === "Steven Spielberg" && movie.genre.includes("Drama"))
+    // const dramaMovies = spielbergMovies.filter(spielberg => spielberg.genre.includes("Drama"))
+    return dramaMovies.length
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+    const averageMovieScore = moviesArray.reduce((acc, current) => acc + current.score, 0) / moviesArray.length
+    return averageMovieScore
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) { 
+    const dramaMovies = moviesArray.filter(movie => movie.genre.includes("Drama"))
+    return dramaMovies.length
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
+    const sortByYear = [...moviesArray].sort((a, b) => a.year - b.year)
+    return sortByYear
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    const movieTitle = [...moviesArray].filter(movie => movie.title)
+    const orderedTitle = movieTitle.sort((a, b) => a - b)
+    return orderedTitle
+}
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
-function turnHoursToMinutes(moviesArray) {}
+function turnHoursToMinutes(moviesArray) {
+    
+}
 
 // BONUS - Iteration 8: Best yearly score average - Best yearly score average
-function bestYearAvg(moviesArray) {}
+function bestYearAvg() { }


src/movies.js
```diff
@@ -1,25 +1,48 @@
+
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+    const allDirectors = moviesArray.map(movie => movie.director)
```
trebeljahr, line 6: could be shorter :) 
```javascript
return movies.map(({director})=>director) 
```

src/movies.js
```diff
@@ -1,25 +1,48 @@
...
+    return allDirectors
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+    const dramaMovies = moviesArray.filter(movie => movie.director === "Steven Spielberg" && movie.genre.includes("Drama"))
```
trebeljahr, line 12: nice. you could get away without the new variable. just return and put .length after the filter.
```js
return moviesArray.filter(({ director, drama }) => movie.director === "Steven Spielberg" && genre.includes("Drama")).length
```

src/movies.js
```diff
@@ -1,25 +1,48 @@
...
+    // const dramaMovies = spielbergMovies.filter(spielberg => spielberg.genre.includes("Drama"))
+    return dramaMovies.length
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+    const averageMovieScore = moviesArray.reduce((acc, current) => acc + current.score, 0) / moviesArray.length
```
trebeljahr, line 19: nice.

src/movies.js
```diff
@@ -1,25 +1,48 @@
...
+    return averageMovieScore
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) { 
+    const dramaMovies = moviesArray.filter(movie => movie.genre.includes("Drama"))
+    return dramaMovies.length
```
trebeljahr, line 26: This should return averageMovieScores for the drama movies. so probably put the dramaMovies variable, so compute scoresAverage(dramaMovies) and return that! ;)

src/movies.js
```diff
@@ -1,25 +1,48 @@
...
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
+    const sortByYear = [...moviesArray].sort((a, b) => a.year - b.year)
```
trebeljahr, line 31: nice. again the intermediate step with the variable name is not necessarily needed. Only have it there if you the extra line. 
otherwise just return.
```js
return [...moviesArray].sort((a, b) => a.year - b.year)
```

trebeljahr: however I see, that this is not completed yet right? 
it missed the case where a.year === b.year and we want to compare a.title to b.title instead. 
A little hint -> don't forget to boolean convert the string comparison, maybe like so?
```js
a.title < b.title ? -1 : 1)
```
so - spoken in algorithms -> 
IF a.year and b.year are equal, sort by titles, i.e. return -1 if a.title comes first and 1 if b.title comes first, ELSE sort by years, like you already do.

src/movies.js
```diff
@@ -1,25 +1,48 @@
...
+function orderAlphabetically(moviesArray) {
+    const movieTitle = [...moviesArray].filter(movie => movie.title)
+    const orderedTitle = movieTitle.sort((a, b) => a - b)
+    return orderedTitle
+}
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
-function turnHoursToMinutes(moviesArray) {}
+function turnHoursToMinutes(moviesArray) {
+
```
trebeljahr, line 44: maybe have a look at the solutions from some of the other students - like Bernd :+1:

trebeljahr: I also wrote my own solution for 7 when reviewing Josh's work, so one way would look like this:

```js
return moviesArray.map((movie) => {
    const hours = movie.duration.match(/(?<hours>\d+)h/)?.groups?.hours || "0";
    const mins = movie.duration.match(/(?<mins>\d+)min/)?.groups?.mins || "0";
    return { ...movie, duration: parseInt(hours) * 60 + parseInt(mins) };
});
``` 

now this might look a bit like chinese but let me explain...
first we have these two:
/(?<hours>\d+)h/
and
/(?<mins>\d+)min/
they use a regex feature called "named capturing groups", which look like this: (?<name> ...)
docs are here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Groups_and_Backreferences

now when we use match from regex, this will create a key under the value of groups for the match of the regex and assign the matched string to it.

so these two search for d+ (any amount of digits) followed by h and mins respectively, then put the digits they matched into groups named "hours" or "mins" respectively, which then can be accessed via .groups.hours and .groups.mins

the extra ? marks in there are to make sure, that if at any step along we have an undefined -> i.e. there was no match for whatever reason -> or no groups -> we don't throw an error when we further chain . -> undefined?.mins can be executed and returns undefined. undefined.mins would throw an error. the last part is the || if the thing before the || returns undefined, for whatever reason -> use "0" as the "default" time string for hours/minutes. This could be the case if the string to parse would be say "2h"

lastly, create a new object, spread over ...movie to copy all of the existing key/value pairs into the object, finally overwrite value with parseInt(hours) * 60 + parseInt(mins) -> because the regex matches / "0" defaults are still strings, we have to do the conversion.

src/movies.js
```diff
@@ -1,25 +1,48 @@
...
+    const sortByYear = [...moviesArray].sort((a, b) => a.year - b.year)
+    return sortByYear
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    const movieTitle = [...moviesArray].filter(movie => movie.title)
+    const orderedTitle = movieTitle.sort((a, b) => a - b)
+    return orderedTitle
```
trebeljahr, line 39: this misses the constraint of the 20, one way would be to use slice to do so :) 

just `return orderedTitles.slice(0, 20)` and it should pass all the tests.

trebeljahr (commented):
good work :) little things here and there are missing but overall this is good. Keep it up like this :)

PR #3077, opened 2022-11-28
WDFTRMTNOV2002-Dave Mullally

David-Mullally:


Dengo777:
Managed iteration 7
** not sure why the original Array is being modifed along with the duplicate array

David-Mullally:
managed to denug iterations 5,6 and 7 by applying feedback info

PR #3133, opened 2023-01-16
WDFT 2023 Tobias Krauß

tobkrauss:


matigobbi (commented):
Nice use of the array methods, 
keep going! 😁

PR #3151, opened 2023-01-19
[LIS-FT-01-23] Duarte Franca

Duartefranca:


src/movies.js
```diff
@@ -1,23 +1,75 @@
...
+function getAllDirectors(moviesArray) {
+    const movie = moviesArray.map((movie) => movie.director)
+    return movie
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+    const moviesStevenSpielberg = [];
+     if (moviesArray.length === 0) {
```
webLisbonTA, line 13: The `if (moviesArray.length === 0) { return 0;}` block is redundant in this function. The filter function will not execute if the moviesArray is empty, so you do not need to check whether the array is empty or not. The function will work fine without this block.

webLisbonTA (commented):
Great job on the lab Duarte!

PR #3153, opened 2023-01-19
[LISFT0123] João Raposo

JRaposo22:


src/movies.js
```diff
@@ -1,22 +1,122 @@
...
+    return directors;
+    
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+
+    let spielberg = moviesArray.filter((director) => director.director === 'Steven Spielberg')
+    let howMany = spielberg.filter((drama) => drama.genre.includes('Drama'))
```
lzaquine, line 18: Here, you don't need to create two variables, you can use the && operator to combine the two filter conditions into a single line of code, which will filter the array for movies directed by Steven Spielberg and that are also of the genre 'Drama'.
```
let howMany = moviesArray.filter((director) => director.director === 'Steven Spielberg' && director.genre.includes('Drama'));
```

src/movies.js
```diff
@@ -1,22 +1,122 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+
+    let directors = moviesArray.map(function(movies) {
```
lzaquine, line 6: Yesterday we learned a bit more about the arrow function. Yours here is correct and if you wanted to use it, it would look like this:

```
let directors = moviesArray.map(movies => movies.director);
return directors;
}
```

Regarding behaviour, both the function and the arrow function will have the same behaviour, but the arrow function is shorter, and easier to read.

src/movies.js
```diff
@@ -1,22 +1,122 @@
...
+    let average = +(sumScores / scores.length).toFixed(2);
+    return  average;
+
+    
+    
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
```
lzaquine, line 46: Instead of copying the logic of the scoresAverage function, you could've returned the scoresAverage function itself here.
Like: `return scoresAverage(numDramas);`

lzaquine (commented):
Good job on the lab, Joao! Try doing the bonuses next time

PR #3154, opened 2023-01-19
[LISFT0123] Alexandre Álvaro

AleAlvo:
- Solved lab - no bonus yet
- completed bonus 1.1, attempted 7

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,83 @@
+
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+    const allDirectors = moviesArray.map((movie) => {
+        return movie.director;
+    })
+    
+    return allDirectors;
+};
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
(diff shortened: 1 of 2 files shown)


src/movies.js
```diff
@@ -1,25 +1,100 @@
...
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+    const allDirectors = moviesArray.map((movie) => {
+        return movie.director;
+    })
+    
+    return allDirectors;
```
webLisbonTA, line 10: It's all good here, but keep in mind that since it is a direct return, you didn't need to create the allDirectors variable and could just return the map

src/movies.js
```diff
@@ -1,25 +1,100 @@
...
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+    const justSpielberg = moviesArray.filter((movie) => {
+        return movie.director === 'Steven Spielberg';
+    })
+    const justSpielbergDrama = justSpielberg.filter((movie) => {
+        return movie.genre.includes('Drama');
+    })
+
+    return justSpielbergDrama.length;
```
webLisbonTA, line 33: Once again, everything is correct, but you could return everything at once. Using the methods you used and the `&&` operator, like this:
```
return moviesArray.filter(
    (movies) => movies.director === 'Steven Spielberg' && movies.genre.includes('Drama')
  ).length;
```

webLisbonTA (commented):
Great job on the lab Alexandre, just check the comments for a better understanding.

PR #3155, opened 2023-01-19
[LISBFT0123] Tomás B

TomBelmar:


src/movies.js
```diff
@@ -1,22 +1,64 @@
...
+            return 0
+        }
+        const moviesSpeilberg = moviesArray.filter(allSpielberg => {
+            return allSpielberg.director === "Steven Spielberg" && allSpielberg.genre.includes("Drama")
+        })
+    return moviesSpeilberg.length    
+};
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
```
joaopdg, line 20: This is correct and it's working, but on the tests they ask you to also `return 0` if the `array` is empty, and that the function should be able to work even if one of the movies doesn't have the `key` `score`.

So for this you need to add two `conditions`, one to check the `length` on the array and `return 0` if it is empty:
```
  if (!moviesArray.length) return 0;
```

And another one to check if the `elements` of the `array` have the `key` `score` if it doesn't continue counting the `score` of the other `elements`:
```
    if (!vall.score) {
      return acc;
    }
```

So at the end this function should look something like this:
```
  if (!moviesArray.length) return 0;

  const average = moviesArray.reduce((acc, vall) => {
    if (!vall.score) {
      return acc;
    }

    return acc + vall.score;
  }, 0);
  const averageTotal = average / moviesArray.length;
  return +averageTotal.toFixed(2);
}
```

src/movies.js
```diff
@@ -1,22 +1,64 @@
...
+    })
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    const movieTitles=moviesArray.map (function (movie){
+        return movie.title;
+    })
+    const alphabeticalOrder = movieTitle.sort().slice(0,20);
```
joaopdg, line 59: Here you are just missing one `s` in `movieTitle`. You declared the `variable` has `movieTitles` but then when you're doing the `.sort()` you missed a `s` and called `movieTitle` so if you go back to your code and just add the `s` your code will pass all the tests

joaopdg (commented):
Great job on the lab Tomás!

PR #3156, opened 2023-01-19
[FT0123] Miguel Leite

AakaiLeite:


src/movies.js
```diff
@@ -1,22 +1,56 @@
...
+        if (!val.score) return acc + 0;
+        return acc + val.score;
+    }, 0);
+    const averageScore = summedScores / moviesArray.length;
+    return Number(averageScore.toFixed(2)); 
+};
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
```
joaopdg, line 28: In the previous iteration you already created a function that calculates the average, so you can use that function inside this new function, and in this way spare a couple of lines of code:

```
  const dramaMovies = moviesArray.filter((movie) =>
    movie.genre.includes("Drama")
  );

  return scoresAverage(dramaMovies);
}
```

src/movies.js
```diff
@@ -50,7 +50,13 @@ function orderByYear(moviesArray) {
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
 function orderAlphabetically(moviesArray) {
```
joaopdg, line 52: There is method called localeCompare that already checks the alphabetical order, so you can use this method instead of the conditions:

```
  const moviesAlphabetically = moviesArray.map((movie) => movie.title);
  moviesAlphabetically.sort((a, b) => {

    return a.localeCompare(b);

  });
  return moviesAlphabetically.splice(0, 20);
}
```

joaopdg (commented):
Great job on the lab Miguel!

PR #3157, opened 2023-01-19
[LISFT0123] + Assebe Kleinert

Assebe:


src/movies.js
```diff
@@ -1,22 +1,2081 @@
...
+
+}
+ console.log(orderByYear(movies)) 
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    const orderedMovies = moviesArray.slice()
+
+    function compare(a, b){
```
webLisbonTA, line 2066: To clone the original array and keep just the titles, you could do this:
```
const onlyTitles = moviesArray.map((movies) => movies.title);
```
After you clone the original array and get just the titles, you could use the `.sort()` method: 
```
onlyTitles.sort(function (a, b) {
    return a.localeCompare(b);
  });
  return onlyTitles.slice(0, 20);
```

src/movies.js
```diff
@@ -1,22 +1,2081 @@
...
+    return (averageDramaScore / dramaMovies.length).toFixed(2)
+}
+//console.log(dramaMoviesScore(movies))
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
+const orderedMovies = moviesArray.slice()
+
+function compare(a, b){
```
webLisbonTA, line 2047: Instead of this function, you could just use the `.sort()` method and check the conditionals:
 ```
orderedMovies.sort((a, b) => {
    if (a.year > b.year) {
      return 1;
    } else if (b.year > a.year) {
      return -1;
    } else {
      if (a.title > b.title) {
        return 1;
        return -1;
      }
      return 0;
    }
  });
  return orderedMovies;
```

webLisbonTA (commented):
Assebe, please check the comments I left you for better understanding of this subject.

PR #3158, opened 2023-01-19
[LISFT0123] Vanessa Vieira

vanessav22:


src/movies.js
```diff
@@ -1,22 +1,78 @@
...
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+
+   
+    if (moviesArray.length === 0) return 0;
+
+   const genreMovie = moviesArray.filter(function(movie) {
+    if (movie.genre.includes("Drama") & movie.director === "Steven Spielberg") {
+     return true;
```
joaopdg, line 18: `.filter()` is an `array` `method` that already `returns` you `true` if the `element` passes the `condition`.
So here you can just do:

```

  const genreMovie = moviesArray.filter(
    (movie) => movie.genre.includes("Drama") && movie.director === "Steven Spielberg");
  return genreMovie.length;
}
```

There's no need for the `if` neither the `return true`
Also you're missing one `&` in the `and (&&)` of the condition

joaopdg (commented):
Great job on the lab Vanessa!

PR #3161, opened 2023-01-19
[LISFT0123] Caio Mazur

caiomazur:


src/movies.js
```diff
@@ -1,25 +1,140 @@
...
+    //console.log(spielbergMovies);
+    return spielbergMovies.length;
+}
+//console.log(howManyMovies(movies))
+
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+    if (!moviesArray.length) return 0;
```
webLisbonTA, line 22: The reduce function won't execute if the moviesArray is empty. 
The code block `if (!moviesArray.length) return 0;` is not needed and the function will work fine without it, it will return 0 as default value for empty array.

src/movies.js
```diff
@@ -1,25 +1,140 @@
...
+    const movArrCopy = moviesArray.map((x) => x)
+    const yearSort = movArrCopy.sort((a, b) => a.year - b.year);
+    return yearSort;
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    const movieTitles = moviesArray.map(movie => movie.title);
+    const alphaSort = movieTitles.sort();
```
webLisbonTA, line 57: After you clone the original array and grab the titles, you could just do the `.sort()` and `.slice()` methods at once `const alphaSort = movieTitles.sort().slice(0, 20);` 
That saves lines of code and prevents errors in the `.filter()`

webLisbonTA (commented):
Great job on the lab Caio. Just check the comments for a better understanding.

webLisbonTA (commented):
Just left a solution for the iteration 7

PR #3162, opened 2023-01-19
[LISFT0123] João Carneiro

joao621119:


src/movies.js
```diff
@@ -1,22 +1,55 @@
...
+        if(movie.score) return acc + movie.score
+        else return acc
+    }, 0)
+    averageScore = sumScores / moviesArray.length
+    return Math.round(averageScore * 100) / 100
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
```
lzaquine, line 28: Instead of copying the logic of the scoresAverage function, you could've returned the scoresAverage function itself here. 
Like: `return scoresAverage(filterMovies);`

lzaquine (commented):
Good job on the lab, Joao! Try doing the bonuses next time

PR #3164, opened 2023-01-19
[LISFT0123] Mariana Franco

marianamedeirosfranco:


src/movies.js
```diff
@@ -1,22 +1,69 @@
...
+    } else return acc
+}, 0)
+
+let averageDrama = sumDramaScore / dramaMovies.length;
+return Math.round(averageDrama *100) / 100
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
```
joaopdg, line 50: This iteration is not passing this test `"should order movies with the same year by their title, alphabetically"`
Your function is working but you need to add a few conditions that will order the movies by title in case there are movies with the same year:

```
  const movArrCopy = moviesArray.map((x) => x);
  const yearSort = movArrCopy.sort((a, b) => {

    if (a.year > b.year) return 1;
    if (a.year === b.year) return a.title.localeCompare(b.title);
  });

  return yearSort;
}
```

src/movies.js
```diff
@@ -1,22 +1,69 @@
...
+
+         const movArrCopy = moviesArray.map((x) => x)
+         const yearSort = movArrCopy.sort((a, b) => a.year - b.year);
+         return yearSort;
+     }
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
```
joaopdg, line 58: Here you could have used `.slice()` instead of `.filter()`:

```
  const movieTitles = moviesArray.map((movie) => movie.title);
  const alphaSort = movieTitles.sort();

  return alphaSort.slice(0, 20);
}
```

joaopdg (commented):
Great job on the lab Mariana!

PR #3165, opened 2023-01-19
[LISFT0123] Erika

erikasoaress:
be gentle with my brain please

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,71 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+
+function getAllDirectors(moviesArray) {
+    const directors = moviesArray.map((movies) => movies.director);
+    return directors;
+}
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(array) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,71 @@
...
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(array) {
+    const movieTypes = array.filter((movies) => movies.director === "Steven Spielberg" && movies.genre.includes("Drama"));
+    
+    return movieTypes.length;
+ }// Iteration 3: All scores average - Get the average of all scores with 2 decimals
+function scoresAverage(scoresArray) {
+    if ( scoresArray.length === 0) {
```
webLisbonTA, line 18: There's nothing wrong here, however, the reduce function won't execute if the moviesArray is empty. So, the code block `if (scoresArray.length === 0) return 0;` is not needed as the function will work fine without it, it will return 0 as default value for empty array.

src/movies.js
```diff
@@ -1,22 +1,71 @@
...
+    return yearSort;
+}
+
+
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    const movieTitles = moviesArray.map(movie => movie.title);
+    const alphaSort = movieTitles.sort();
```
webLisbonTA, line 59: After you clone the original array and grab the titles, you could just do the `.sort()` and `.slice()` methods at once `const alphaSort = movieTitles.sort().slice(0, 20);`
That saves lines of code and prevents errors in the `.filter()`

webLisbonTA (commented):
Good job on the lab Erika, just check the comments for a better understanding.

PR #3167, opened 2023-01-19
[LISFT0123] Tiago Rato

TiagoRato13:


src/movies.js
```diff
@@ -1,25 +1,128 @@
...
+function orderAlphabetically(moviesArray) {
+    let arr = [...moviesArray];
+    
+    const orderMovies = arr.map((name) => name.title);
+    return orderMovies.sort().splice(0, 20);
+  }
+  
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
-function turnHoursToMinutes(moviesArray) {}
+function turnHoursToMinutes(moviesArray) {
```
joaopdg, line 102: The logic is more or less what you've done here but you were missing some steps:

```
function turnHoursToMinutes(moviesArray) {

  function convertDuration(movie) {
    let total = movie.duration.split(" ");
    let hour = 0
    let min = 0
    if (total[0].includes("h")) {
      hour = total[0].split("h");
      hour = Number(hour[0]) * 60;
    }
    if (total[1] && total[1].includes("min")) {
      min = total[1].split("min");
      min = Number(min[0]);
    }
    total = min + hour;
    return Number(total);
  }

  let newArray = moviesArray.map((oneMovie) => {
    let newMovie = {};
    newMovie.title = oneMovie.title;
    newMovie.year = oneMovie.year;
    newMovie.director = oneMovie.director;
    newMovie.duration = convertDuration(oneMovie);
    newMovie.genre = oneMovie.genre;
    newMovie.score = oneMovie.score;
    return newMovie;
  });
  return newArray;
}
```
First you create a function (convertDuration) inside this function (turnHoursToMinutes) that will convert the duration of only one movie into minutes. 
Then you need to map over the given array and make a copy of every single element of the array but, instead of copying the duration you call the previous function that will convert it into minutes.

joaopdg (commented):
Great job on the lab Tiago!

PR #3168, opened 2023-01-19
[LISFT0123] Rafaela Umbelino

RafaelaUmbelino:


src/movies.js
```diff
@@ -1,25 +1,85 @@
...
+  }, 0);  
+
+   return Math.round ((scores / moviesArray.length) * 100) / 100;
+   
+}
+   
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
```
joaopdg, line 34: This iteration is not passing this test `"should only calculate the average for drama movies"`
In the previous iteration you already created a function that calculates the average (scoresAverage). So you can use it inside this new function, and in this way the only thing left to do is check if the movie includes the genre drama:

```
  const dramaScores = moviesArray.filter((movie) =>
    movie.genre.includes("Drama")
  );
  return scoresAverage(dramaScores);
}
```

src/movies.js
```diff
@@ -1,25 +1,85 @@
...
+        }
+  }, 0);  
+
+   return Math.round ((dramaScores / moviesArray.length) * 100) / 100;
+   
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
```
joaopdg, line 48: This iteration is not passing this test `"should order movies with the same year by their title, alphabetically"`
In your last condition you have (a.title < b.title) it is correct but, first you need to check if a.year is equal to b.year, and then if this is true you need to compare the titles to decide which one comes first:

```
  const orderedByYear = moviesArray.slice();
  function compare(a, b) {
    if (a.year > b.year) {
      return 1;
    } else if (a.year < b.year) {
      return -1;

    } else if (a.year === b.year);
  }

  return orderedByYear.sort(compare);
}
```

joaopdg (commented):
Great job on the lab Rafaela!

PR #3170, opened 2023-01-19
[LISFT0123] Pedro Leal

PeterLoyal:


src/movies.js
```diff
@@ -1,22 +1,61 @@
...
+        const avg = moviesArray.reduce((acc ,val) => {
+    if(!val.score) {return acc}
+        return acc + val.score}, 0);
+    const decimal = avg / moviesArray.length;
+    return Math.round(decimal * 100) / 100;
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
```
joaopdg, line 27: In the previous iteration you already created a function that calculates the average, so you can use that function inside this new function, and in this way spare a couple of lines of code:

```
  const avgDram = moviesArray.filter((drama) => drama.genre.includes("Drama"));

  return scoresAverage(avgDram);
}
```

joaopdg (commented):
Great job on the lab Pedro!

PR #3171, opened 2023-01-19
[LISFT0123] Gabriel

gbrlcsgn:


src/movies.js
```diff
@@ -1,22 +1,64 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+    return movies.map((value) => value.director)
```
lzaquine, line 5: The function is trying to access the 'movies' property, but it should be accessing the 'moviesArray' parameter that is passed into the function.

`return moviesArray.map((value) => value.director)`

src/movies.js
```diff
@@ -1,22 +1,64 @@
...
+            }
+        }, 0)
+        return Number((moviesScore / movies.length).toFixed(2));
+      }
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(movies) {
+    const dramaMovies = movies.filter((films) => films.genre.includes("Drama"))
+    return scoresAverage(dramaMovies);
```
lzaquine, line 32: Great job using the scoresAverage() here!

src/movies.js
```diff
@@ -1,22 +1,64 @@
...
+} 
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(movies) {
+    const order = movies.splice(0, moviesArray.length)
```
lzaquine, line 37: The function is trying to access the length property of moviesArray, but it doesn't exist. It should be `movies.length`

src/movies.js
```diff
@@ -1,22 +1,64 @@
...
+                return 0;
+            }
+        } return a.year - b.year
+    })
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(movies) {
+    function orderAlphabetically(movies) {
```
lzaquine, line 53: This function is defining another function with the same name 'orderAlphabetically' inside the first one. This is not needed and will cause an error.

lzaquine (commented):
Good job on the lab, Gabriel!

PR #3172, opened 2023-01-20
[LISFT0123] André Lisboa

andrelisboapt:
solved lab

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,63 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,63 @@
...
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
+
+    sortedArray = []
+    for(let i = 0; i < moviesArray.length; i++){
+        sortedArray[i] = moviesArray[i]
+    }
+    sortedArray.sort((a, b) => a.year - b.year)
```
webLisbonTA, line 47: You missed the conditionals in the sort method.
 ```
 sortedArray.sort((a, b) => {
    if (a.year > b.year) {
      return 1;
    } else if (b.year > a.year) {
      return -1;
    } else {
      if (a.title > b.title) {
        return 1;
        return -1;
      }
      return 0;
    }
  });
  return sortedArray;
```

webLisbonTA (commented):
Good job on the lab André

PR #3173, opened 2023-01-20
[LISFT0123] Francisco Pereira

FranciscoManuelPereira:
-Lab done with my colleagues but many doubts.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,70 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  let container = [];
+  moviesArray.map((movie) => {
+    container.push(movie.director);
+  });
+  return container;
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+  let genreDirector = moviesArray.filter(
+    (movie) =>
+      movie.director === "Steven Spielberg" && movie.genre.includes("Drama")
+  );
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,70 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  let container = [];
```
lzaquine, line 5: Just so you know – the container variable is not necessary here. The map method returns a new array, and the push method is not needed.
You can directly use the map method to return the array of directors, like this:

```
  return moviesArray.map((movie) => movie.director);
}
```
This way, the function will return an array containing the 'director' property of each object in the input array without creating an additional container variable.

src/movies.js
```diff
@@ -1,22 +1,70 @@
...
+  return Math.round((totalScores * 100) / moviesArray.length) / 100;
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
+  const dramaMovies = moviesArray.filter((movie) =>
+    movie.genre.includes("Drama")
+  );
+  return scoresAverage(dramaMovies);
```
lzaquine, line 40: Great job on using the scoresAverage function here!

src/movies.js
```diff
@@ -1,22 +1,70 @@
...
+    (movie) =>
+      movie.director === "Steven Spielberg" && movie.genre.includes("Drama")
+  );
+  return genreDirector.length;
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  if (moviesArray.lenght === 0) return 0;
```
lzaquine, line 23: small typo here:
`length` instead of lenght

lzaquine (commented):
Great job on the lab, Francisco!

PR #3175, opened 2023-01-20
(LISFT0123)Farid Cunha

Farid-97:


src/movies.js
```diff
@@ -1,22 +1,61 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
-
+function getAllDirectors(moviesArray) {
+    let film=moviesArray.map(function(cinema){
```
lzaquine, line 5: Yesterday we learned a bit more about the arrow function. Yours here is correct, and if you wanted to use it, it would look like this:

```
let film = moviesArray.map(cinema => cinema.director);
return film;
}
```
Regarding behaviour, both the function and the arrow function will have the same behaviour, but the arrow function is shorter and easier to read.

src/movies.js
```diff
@@ -1,22 +1,61 @@
...
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
-
+function howManyMovies(moviesArray) {
+    let countDown=moviesArray.filter((cinema2) => cinema2.director==="Steven Spielberg"&& cinema2.genre.includes("Drama"))
+    return countDown.length
+}
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+if (moviesArray.length === 0) {
```
lzaquine, line 16: Everything here is ok but you deleted this line:

`function scoresAverage(moviesArray) {`

src/movies.js
```diff
@@ -1,22 +1,61 @@
...
+}, 0)
+return Number((movieScore / moviesArray.length).toFixed(2));
+
+
+   
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) { const dramaMovies = moviesArray.filter( movie => movie.genre.includes('Drama'))
+return scoresAverage(dramaMovies)
```
lzaquine, line 33: Nice job using the scoresAverage function!

lzaquine (commented):
Good job on the lab, Farid!

PR #3176, opened 2023-01-20
[LISFT0123] Déborah

deborahdelimaa:


src/movies.js
```diff
@@ -1,25 +1,84 @@
...
+    // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
+     function howManyMovies(moviesArray) {
+     const newArray= moviesArray.filter((movie) => 
+     movie.director === "Steven Spielberg" && movie.genre.includes("Drama"))
+     return newArray.length;
+    };
+    
+    // Iteration 3: All scores average - Get the average of all scores with 2 decimals
+    function scoresAverage(moviesArray) {
+        if (moviesArray.length === 0) return 0;
```
webLisbonTA, line 27: There's nothing wrong here, however, the reduce function won't execute if the moviesArray is empty. So, the code block `if (!moviesArray.length) return 0;` is not needed as the function will work fine without it, it will return 0 as default value for empty array.

webLisbonTA (commented):
Great job on the lab Deborah.

PR #3231, opened 2023-02-22
Develop

serge-svg:


serge-svg:
Recovered the original movies.spec.js file

PR #3250, opened 2023-02-27
[RMTFTFEB23] valentin

denefi:
First commit that completes 4 iterations, I hope to finish some more in the evening.


uebu12:
Great job!
Really smart to use the previous functions when possible
Very clean code
You should get into the habit of commenting your code, something like: "This loop is doing this, this conditions is checking that, etc", it's really good practice and will help you a lot in the long run

PR #3251, opened 2023-02-27
[FTRMT022023] - Marin Valenta

valyy151:
Iteration 1-7  |  Green

Iteration 8  |  Not attempted ( red )


ykztenoks:
Hey Marin! 
Great stuff on your development, well structured and clever code.
Small improvement would be:
In the `scoresAverage()` you don't need that `filter()` as the reduce will already return the value we want as long as you use the `movie.score`. And a quick tip would be to reutilize this function for the `dramaMoviesScores()`.
Other than that, awesome work and keep it up!

PR #3252, opened 2023-02-27
[RMT WEBDEV FEB2023] Cassandre Lavallard

Canacepa:
Song of the day 😃 :
https://youtu.be/yE-qYZAwiHs


ykztenoks:
Hi Cassandre! Awesome work on your lab, really well structured following the best practices!
Only thing I could notice is when you orderByYear(), you could use localeCompare() for the movies that have the same year, because sort() is case sensitive as it's based on the ASCII Table, so if you just compare with < or > it might not always give you the wanted answer, consequently that's applied for the orderAlphabetically() function.
Other than that, keep up the great and clean work!! 🥳

PR #3255, opened 2023-02-27
[WD-RTM-FT-Feb2023] Chris

Webster312:
can't get it all


ykztenoks:
Hey Cristoph! Great work on your lab! Good structure and best practices being followed, keep it up.
One small thing would be for ordering alphabetically, the sort() method doesn't always give us the wanted result, so instead we should use localeCompare(). That is because if you just use sort for sorting alphabetically, there'll be a problem if you have both capital and lowercased initials for the word, as sort() is based on the ASCII table. Other than that, great work again!

PR #3256, opened 2023-02-27
[FTRMT022023] Pete

PDXIII:
LAB solved!
Iterations 1 - 6 are done. Bonus comes maybe later.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,7 +1,12 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey peter! 
Congrats on your development, very clever work and props to the `dramaMoviesScore()`.
Small improvements would be:
A even cleaner function for `howManyMovies()` would be to use only one `filter()` and the && (and) operator, like such:
`moviesArray.filter((movie) => movie.genre.includes("Drama") && movie.director === "Steven Spielberg")`
The other thing is a different approach on the `scoresAverage()` as you should rarely use `switch`. Would be the following:
`moviesArray.reduce((acc, movie) => acc += (movie.score || 0))`
You didn't need to check if it was a string but still a clever solution you had, this is just another approach with less code written.
By the way : `const result = +(sum / moviesArray.length).toFixed(2)` is completely legitimate 😅 the unary operator+ in front of the string will turn it into a floating number.

Other than those small details, awesome work, keep it up!

PR #3259, opened 2023-02-27
[FTRMT022023] amina mohamed khedr

amina247:


ykztenoks:
Hey Alexandre! Good work on the iterations, clean way of using array methods, I'm sure if you pushed it you'd rock all of the iterations.
Small thing is that in the scores avg iteration you were referencing the wrong array. 
Any questions don't hesitate to ask.

PR #3260, opened 2023-02-27
[RMT-WD.FT-FEB.2023] Ismael Freitas

IsmaelFreitas98:
Almost had a brain meltdown in iteration 8, but Jasmine is all green! :)


ykztenoks:
Hey Ismael!
Good stuff on completing all iterations! Got some clever solutions in there.
Small improvements would be:
In `orderByYear()` you could use the localeCompare() to compare movie titles in case they had the same year, as sort() is not going to check for lower/upper cases as it's based on the [ASCII Table](https://www.asciitable.com/).
As well as for the `orderAlphabetically()`, at the end to grab only 20 elements in the array, we could use the `slice()` method. `newArray.slice(0, 20)` 

Other than that, great work and keep it up!

PR #3262, opened 2023-02-27
[rmt ft 20.feb.2023]Peng Chen

Mr-Chennn:
from 1 to bonus 7


ykztenoks:
Hey Peng!
Congrats on this lab, awesome work with clever solutions and clean code.
One small note that I could give would be:
In the `orderAlphabetically()` to get the first 20 items in the array you could use the `slice()` method.
`newArray.slice(0, 20)`

Other than that, keep up the great work!

PR #3263, opened 2023-02-27
[WD-RMT-FT-Feb2023] Tuna Erkmen 

TunaErkmen:
This LAB is very complicated and has long iterations. Please add more explanation to Jasmine's iterations.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -47,9 +47,10 @@ return Math.round(averageScore * 100) / 100;
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 function dramaMoviesScore(moviesArray) {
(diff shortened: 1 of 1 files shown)


uebu12:
You had a small mistake on the first iteration where you didn't map the moviesArray argument.

The first 2 conditions in you second iteration can be removed without affecting the end result, your filter already cover those cases.

The third iteration will never fully work because you are trying to calculate the average score only of the movies that have an 8 score (that average will always be 8). The test that says "should return the average score of 2 movies with score 8 each" is just there to check if your function works on integer numbers. You can try that again, calculating the average of all scores and using the `toFixed()` method

Your logic on the fourth iteration is perfect, you just need to get the `scoresAverage()` working to pass all the tests.

PR #3264, opened 2023-02-27
[RMT FT WD 022023] Mattie Barton

maitlandbarton:
solved all the mandatory iterations, they all seemed to be working when I tried them in stackblitz but a lot of the tests in jasmine still came up red :/

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2001,3 +2001,5 @@ const movies = [
     score: 8
   }
 ];
+
+export{movies};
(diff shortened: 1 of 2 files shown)


ykztenoks:
Hey Mattie! 
Great stuff on your lab, clever solutions and clean code.
One small improvement to make it even cleaner is in the `dramaMoviesScore()` for the `filter`, we could do instead:
`let dramaMovies = moviesArray.filter((current) => current.genre.includes("Drama"))`

That way we have a cleaner function were we can still understand what we are receiving from the `filter`.
Really small thing really, you did a great job anyways, keep it up!

PR #3265, opened 2023-02-27
[rmt-wb-ft-0223]Pim

8lom:
#1, it says it should return same lengt array. I used map. so that should leave it as the same length?!
#2, I can't access the array of the genre
#3, works in stackblitz, but only 2/5 in jasmine
#4, I can't access the array of the genre

Needs more research,  explanation and practice

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,55 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
-
+function getAllDirectors(moviesArray) {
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hi Pim! 
Great work on your development on this lab!
#1 - Yes, [map ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) will always return a new array with the same amount of elements from the original array.
#2 - As genre is an array of genres, we could use the method [includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) to check if a genre is in the array, like such: 
`moviesArray.filter((movie) => movie.genre.includes("Drama"))`
With this code we would have a new array only with movies that have 'Drama' in their genres array.
#3 - Jasmine has very specific tests, it's very challenging to get it all green but it's doable.
#4 - Same as #2 😅 

Again, great job on your development and if you have any questions, please let me know!

PR #3266, opened 2023-02-27
[FTRMT022023]DANIAL RAJA

danialraja1600:
Completed as much as I could.


TA-Remote:
Really well done, clean code and very smart and readable solutions!🙌
I like you search for different way to go, (like doing a copy using .slice() ) and using parsInt to get a number from a string.
In iter 6, don't need to make a copy of the result. Starting with the .map already returns a new array(with the titles) you are then sorting (and modify) that new array. 
next step could be trying to chain the methods 😉
Keep coding!!!!

PR #3269, opened 2023-02-28
[RMT FT 02202023] Paulius Mockevicius

pmockevicius:
Got stuck on iteration 6. Will try to continue on my free time. Fun lab

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -79,14 +79,35 @@ function orderAlphabetically(moviesArray) {
     })
     const first20 = []
     sortedAbc.forEach(element => {
-        first20.push(element.title)  
+        if (first20.length < 20){
+            first20.push(element.title)
+        } else 
+        return
+          
     });
-    console.log(first20)
+    
     return first20
 }
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
-function turnHoursToMinutes(moviesArray) { }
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Paulius!
Great job on this lab, clever solutions and good structure.
Few improvements would be:
For the array methods you have, we could simplify it and make it cleaner for our code. if you didn't have any previous experience I see that writing that way is easier to understand, but I'll give you these suggestions just to keep in mind 😁
For example in the `dramaMoviesScores()` instead of:
const dram = moviesArray.filter(function (film) {
` const dramaArray = moviesArray.filter(function (movie) {`
        `return movie.genre.includes('Drama');`
      `});`

We could do:

 `const dramaArray = moviesArray.filter((movie) => movie.genre.includes('Drama'))`

That way we have a cleaner way of getting the movies that have "Drama" in their genres array. This is a good practice for almost all array methods when we are not doing any other complex operation inside the method. As soon as you get comfortable using these methods I would suggest you to try this syntax in case you find that interesting.

One other small thing would be for limiting to 20 elements in the `orderAlphabetically()`, to make it simpler we could use the `slice()` method. 
`newArray.slice(0, 20)`

I know it's a lot of info, but don't worry as those are only details, you did an awesome job, if you have any questions don't hesitate to ask and keep it up!

PR #3270, opened 2023-02-28
[FTRMT022023] Hugo

hugowannabedev:
I only managed to complete the first 2 iterations. It was a bit confusing to understand and apply the knowledge from the classes.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,13 +1,40 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Hugo! 
Congrats on your development, nice structure and clever solutions, props for reutilizing the avg function!
Small improvements would be:
on the `scoresAverage()` you don't need the `filter()` as the `reduce()` will already return the value you need, and just to make sure that all the movies have scores, otherwise we say 0 we could do:
`let totalScore = moviesArray.reduce((sum, movie) => {
      return sum + (movie.score || 0);
    }, 0);`

Other than that, great work and I'm sure if you had enough time you'd rock all iterations. Keep it up!

hugowannabedev:
Hey Daniel,

Thanks for the feedback, it definitely helps!
I'll work on it and try to improve eheh.

Best regards,

Sent with [Proton Mail](https://proton.me/) secure email.

------- Original Message -------

PR #3272, opened 2023-02-28
[RMT FT 20-FEB-2023]Hristos Pappas

hristosppp:
this lab was really challenging, but solved the iteration 1-7 with the methods we recently learnt. The Bonus is to much for my head for now :)


ykztenoks:
Hey Hristos! 
Good stuff on completing all mandatory iterations! Got some clever solutions in there.
Small improvements would be:
In orderByYear() you could use the `localeCompare()` to compare movie titles in case they had the same year, as `sort()` is not going to check for lower/upper cases as it's based on the [ASCII Table](https://www.asciitable.com/). 
Also on the `orderAlphabetically()`, instead of a `forEach()` to limit the elements in the array, you could use the `slice()` method:
`newArray.slice(0, 20)`
Other than that, great work and if you have any questions don't hesitate to ask!

PR #3274, opened 2023-02-28
[RMT FT 20-FEB-2023] Nihat Algan

nihatalgan:
That was hard

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,89 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Nihat! 
Congrats on your development, some great solutions in there.
Small improvements would be:
For the `filter` methods you have, we could simplify it and make it cleaner for our code. if you didn't have any previous experience I see that writing that way is easier to understand, but I'll give you these suggestions just to keep in mind 😁
For example in the `dramaMoviesScores()` instead of:
` const dram = moviesArray.filter(function (film) {
    if (film.genre.includes("Drama")) {
      return true;
    }
  });`

We could do:

` const dram = moviesArray.filter((film) => film.genre.includes("Drama"))`

That way we have a cleaner way of getting the movies that have "Drama" in their genres array. This is a good practice for almost all array methods when we are not doing any other complex operation inside the method. As soon as you get comfortable using these methods I would suggest you to try this syntax in case you find that interesting.
Other than that, congrats again and if you have any questions don't hesitate to ask!

PR #3276, opened 2023-02-28
[rmt- ft- 2023] Senem N. Osmanova

SenemDEV:
I honestly could not really complete much of the lab, the consept is now a little more clearer but combining different methods or functions is still confusuing to me.


uebu12:
You don't need the if statements in the filter method, you could just do
 `const totalMovies = moviesArray.filter(function (currentMovie) {
     return  currentMovie.director === "Steven Spielberg" && currentMovie.genre.includes("Drama")    
       });`

In the fourth iteration, you tried to do everything inside the `reduce()` method. Always try to take one step at a time, in this case, first get the array of movies that are drama and then you use the `reduce()` method to calculate the score average

PR #3346, opened 2023-04-13
LIS-WD-FT-APR-23-TOMÁS-MARIA

SinTomas:


src/movies.js
```diff
@@ -1,25 +1,129 @@
...
+
+    let dramaMovies = moviesArray.filter((movie) =>
+        movie.genre.includes("Drama"));
+
+    if (dramaMovies.length===0) {
+        return 0;
+    }
+
+
+    let dramaMoviesTotalScores = dramaMovies.reduce((accumulator, currentValue) => {
```
brunorocha20, line 66: You previously wrote a function that calculates the average `scoresAverage`, so, instead of writing again the whole logic to do the average inside this `dramaMoviesScore` function, you should reuse `scoresAverage`.
After you `filter` the movies, you should just `return scoresAverage(dramaMovies)`.

brunorocha20 (commented):
Great job on the lab Tomás.

PR #3349, opened 2023-04-13
[Lis-wdft-2023] - Lucas Telles

lltelles:


src/movies.js
```diff
@@ -1,22 +1,92 @@
...
+    }
+
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    titles = moviesArray.map((i)=>{
+        return i.title; 
+    })
+    return titles;
```
brunorocha20, line 85: Here, you just returned all the titles of the `moviesArray`, but since you now have that, you just need to order them alphabetically and then get only the first 20.
So, straight in the return, you could use the `.sort()` method that by default is going to order it ascendingly and the `.slice()` method that allows you to get just a piece (or a slice) of the sorted array.
Here's a possible solution:
```
  titles = moviesArray.map((i)=>{
      return i.title; 
  })
  return titles.sort().slice(0, 20);
}
```

brunorocha20 (commented):
Good job on the lab Lucas.

PR #3355, opened 2023-04-13
Lis-WDFT-2023-MarisaPinheiro

Marisa-Pinheiro:
Couldn't finish number 7, wrote down the logical steps I tough off.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,25 +1,114 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,25 +1,114 @@
...
+
+  scoresDramaArray = moviesArray.filter((movie) => {
+    return movie.genre.includes("Drama");
+  });
+
+  scoresDramaArray = scoresDramaArray.map((movie) => {
+    return movie.score;
+  });
+
+  let total = scoresDramaArray.reduce((acc, film) => {
```
brunorocha20, line 50: Previously, you wrote a function that calculates the average scores, right? So, here, you should reuse it. Instead of writing again the logic to calculate the average, you can call the `scoresAverage` function.
`return scoresAverage(scoresDramaArray)`.

brunorocha20 (commented):
Great job on the lab Marisa. Check lab solution for iteration 7.

src/movies.js
Marisa-Pinheiro: thank you!

PR #3356, opened 2023-04-13
Lis-WDFT-APR-2023-Nuno-Durão

NunoDurao:


src/movies.js
```diff
@@ -1,25 +1,111 @@
...
+  
+    scoresDramaArray = moviesArray.filter((movie) => {
+      return movie.genre.includes("Drama");
+    });
+  
+    scoresDramaArray = scoresDramaArray.map((movie) => {
+      return movie.score;
+    });
+  
+    let total = scoresDramaArray.reduce((acc, film) => {
```
brunorocha20, line 53: Previously, you wrote a function that calculates the average scores, right? So, here, you should reuse it. Instead of writing again the logic to calculate the average, you can call the `scoresAverage` function.
`return scoresAverage(scoresDramaArray)`

brunorocha20 (commented):
Good job on the lab Nuno.

PR #3361, opened 2023-04-14
[LIS-WDFT-APR2023] Elnaz

elnazfe:


src/movies.js
```diff
@@ -1,25 +1,117 @@
...
+    const scores = moviesArray.filter(movie => typeof movie.score === 'number');
+        
+    if (scores.length === 0) {
+       return 0;
+    }
+        
+    const total = scores.reduce((sum, movie) => sum + movie.score, 0);
+    const average = total / scores.length;
+        
+    return Math.round(average * 100) / 100;
```
joaopdg, line 44: In this iteration you're not passing the last test because you're excluding the movies that don't have score, and what they want is for the movies that don't have score you count it as 0:

```
    return 0;
  }

  const total = moviesArray.reduce((sum, movie) => {
    if (!movie.score) {
      return sum;
    }
    return sum + movie.score;
  }, 0);

  const average = total / moviesArray.length;
  return Math.round(average * 100) / 100;
}
```

joaopdg (commented):
Great job on the lab Elnaz!

PR #3362, opened 2023-04-14
[LIS-WD-FT-APR-2023] Margarida Pereira

mapmargaridapereira:


brunorocha20 (commented):
Great job on the lab Margarida.

src/movies.js
```diff
@@ -1,22 +1,63 @@
...
+function orderByYear(moviesArray) {
+    const movieYear = [...moviesArray]
+    return movieYear.sort((a, b) => {
+      if (a.year > b.year) return 1
+      else if (a.year < b.year) return -1
+      else if (a.title > b.title) return 1
+      else if (a.title < b.title) return -1
+      return 0
+    })
+    return movieYear;
```
brunorocha20, line 50: This `return` here is an unreachable code, since you are returning directly the `movieYear.sort(...)` instead of a variable with that result, this second `return` is not being read.

brunorocha20 (commented):
Great job on the lab Margarida

PR #3364, opened 2023-04-14
Solved lab

Gui-Dantas:


src/movies.js
```diff
@@ -1,22 +1,77 @@
...
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    let orderTitles = [...moviesArray];
+    orderTitles.sort((a, b)=>{
+        return a.title.localeCompare(b.title);
+    });
+    return orderTitles.slice(0,20);
```
joaopdg, line 73: In this iteration you're failing 3 tests because:
- you are sorting the array but you are not storing it any where
- they want you to return only the titles and you're returning the whole object

```
  //making a copy of the original array and storing it in copyArray
  let copyArray = [...moviesArray];

  //getting only the titles from the movies and storing it in titlesArray
  let titlesArray = copyArray.map((movie) => movie.title);

  //sorting the array and storing it in sortedTitlesArray
  let sortedTitlesArray = titlesArray.sort((a, b) => {
    return a.localeCompare(b);
  });

  //returning just the first 20 movies
  return sortedTitlesArray.slice(0, 20);
}
```

joaopdg (commented):
Good job on the lab Guilherme, check the correction below and try to apply it

PR #3367, opened 2023-04-14
[LIS-WDFT-APR23] Pedro Nogueira

Pedro-No:


src/movies.js
```diff
@@ -1,22 +1,65 @@
...
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
+    const dramaMovies = moviesArray.filter(movies => movies.genre.includes("Drama"));
+    const dramaMoviesScore = dramaMovies.map(movies => movies.score);
+    if (dramaMovies.length ===0){
+        return 0;
+    } else {
+        
+        const averageScore = (dramaMoviesScore.reduce((p,p2) => p + p2)/dramaMovies.length);
```
brunorocha20, line 41: Previously, you wrote a function that calculates the average scores, right? So, here, you should reuse it. Instead of writing again the logic to calculate the average, you can call the `scoresAverage` function.

brunorocha20 (commented):
Great job on the lab.

PR #3368, opened 2023-04-14
[LIS WDFT APR23] CRISTIAN

Cris-k1:


src/movies.js
```diff
@@ -1,19 +1,64 @@
...
+        if (allDrama.genre.includes('Drama'))
+        return allDrama
+        
+    })
+    return scoresAverage(dramaM)
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
```
joaopdg, line 51: This is one of the possible solutions for this iteration:

```
  //making a copy of the original array
  const copyArray = [...moviesArray];

  //sort the copy of the original array
  const sortedArray = copyArray.sort((a, b) => {
    //if there are movies with the same year sort alphabetically
    if (a.year === b.year) {
    }
    //if not sort by year
    return a.year - b.year;
  });

  //return the copy of the array sorted
  return sortedArray;
}
```

joaopdg (commented):
Good job on the lab Cris, check the corrections that I left you below and check the lab's solution for the rest of the iterations

Cris-k1:
Nossa! Então em partes eu tava seguindo a lógica do sort, só não sabia
executar de forma correta

PR #3392, opened 2023-04-25
[WDPT-APR23] Andres Franco

Andresfrla:


src/movies.js
```diff
@@ -60,10 +60,22 @@ function dramaMoviesScore(moviesArray) {
...
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    const titles = moviesArray.map(movie => movie.title).sort()
+    return titles.slice(0,20)
+}
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 function turnHoursToMinutes(moviesArray) {}
```
ta-web-mexpt2, line 81: I share with you my solution to this 

```
 if (typeof movie.duration === "number") {
      return movie;
    }
    
    movieDuration = movie.duration;

    const timeString = movieDuration.split(" ");
    // Example:
    // "2h 22min"   ["2h", "22min"]
    // "2h"   ["2h"]
    // "22min"  ["22min"]

    const minutesDuration = timeString.reduce(function(total, string) {
      if (string.includes("h")) {
        const numOfHours = parseInt(string);
        return total + numOfHours * 60;
      } else {
        const numOfMinutes = parseInt(string);
        return total + numOfMinutes;
      }
    }, 0);

    movie.duration = minutesDuration;
    return movie;
```

PR #3394, opened 2023-04-29
[PTMEX042023]Victor-Tovar

TovCrowe:


src/movies.js
```diff
@@ -1,25 +1,86 @@
...
+},0)
+
+const avg = sumAllScores / moviesArray.length
+
+return Number(avg.toFixed(2))
+}
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+const moviesArray = [
```
ta-web-mexpt2, line 39: why this array ? if the array comes from data ?

src/movies.js
```diff
@@ -1,25 +1,86 @@
...
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    const moviesArrayDC = JSON.parse(JSON.stringify(moviesArray))
+    return moviesArrayDC
+    .map(movie => movie.title)
+    .sort((a,b) => a.localeCompare(b))
+    .slice(0, 20)
```
ta-web-mexpt2, line 75: nice move there

src/movies.js
```diff
@@ -1,25 +1,86 @@
...
+    
+}
 
 // BONUS - Iteration 7: Time Format - Turn duration of the movies from hours to minutes
 function turnHoursToMinutes(moviesArray) {}
```
ta-web-mexpt2, line 80: const movies = JSON.parse(JSON.stringify(moviesArray))
  movies.map(movie => {
    const mIndex = movie.duration.indexOf('min') - 2
    const hToMin = movie.duration[0] * 60
    let minutes = 0

    movie.duration.indexOf("min") === - 1
      ? minutes = 0
      : minutes = Number(movie.duration.slice(mIndex, mIndex + 2))

    movie.duration = hToMin + minutes
  })

  return movies```
```

File: src/movies.js [hunk 4]
```diff
@@ -1,25 +1,86 @@
...
 
 // BONUS - Iteration 8: Best yearly score average - Best yearly score average
 function bestYearAvg(moviesArray) {}
```
ta-web-mexpt2, line 83:
```
 const movies = JSON.parse(JSON.stringify(moviesArray))
  let bestYear = 0, bestScore = 0

  // What if the array is empty? Return null
  if(movies.length < 1) return null

  movies.map(movie => movie.year)
    .filter((year, index, self) => self.indexOf(year) === index) // saves only the years
    .sort((a, b) => a - b) // sorts incrementally
    .forEach(year => { // let's evaluate one year at a time

      // Filters the movies with the year in turn
      const moviesPerYear = movies.filter((movie) => movie.year === year)

      // creates a variable for the Acc to register the sum of scores
      let yearScore = moviesPerYear.reduce((acc, movie) => {
        return acc + (movie.score / moviesPerYear.length) // Score average
      }, 0)

      // Adjusts the best score
      bestScore < yearScore
        ? ((bestScore = yearScore), (bestYear = year))
        : void 0
    })

  return `The best year was ${bestYear} with an average score of ${bestScore}`
```

PR #3395, opened 2023-04-29
[MEXWDPTAPR2023]-Juan Pablo Maurer

churrundo:


src/movies.js
```diff
@@ -1,25 +1,158 @@
...
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+    return moviesArray.map(movies=>movies.director)
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {    
+    let spielbergMovies = moviesArray.filter(function(movie){
```
ta-web-mexpt2, line 10:
```suggestion
     const stevenDramaMovies = moviesArray
    .filter(movie => movie.director === "Steven Spielberg" && movie.genre.includes("Drama"))

  return stevenDramaMovies.length
```

src/movies.js
```diff
@@ -1,25 +1,158 @@
...
+    return Math.floor(avg*100)/100
+}
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
+    let sortedMovies = JSON.parse(JSON.stringify(moviesArray))
+    
+    sortedMovies.sort(function(a,b){
+        if (a.year!==b.year){
```
ta-web-mexpt2, line 62: if statetment can be a switch when you have more than 3 if nested

src/movies.js
```diff
@@ -1,25 +1,158 @@
...
+        return {...movie,
+                duration:durationInMinutes}
+    })    
+}
+
 
 // BONUS - Iteration 8: Best yearly score average - Best yearly score average
-function bestYearAvg(moviesArray) {}
+function bestYearAvg(moviesArray) {
+    let moviesByYear
```
ta-web-mexpt2, line 108:
```suggestion
   const movies = JSON.parse(JSON.stringify(moviesArray))
  let bestYear = 0, bestScore = 0

  // What if the array is empty? Return null
  if(movies.length < 1) return null

  movies.map(movie => movie.year)
    .filter((year, index, self) => self.indexOf(year) === index) // saves only the years
    .sort((a, b) => a - b) // sorts incrementally
    .forEach(year => { // let's evaluate one year at a time

      // Filters the movies with the year in turn
      const moviesPerYear = movies.filter((movie) => movie.year === year)

      // creates a variable for the Acc to register the sum of scores
      let yearScore = moviesPerYear.reduce((acc, movie) => {
        return acc + (movie.score / moviesPerYear.length) // Score average
      }, 0)

      // Adjusts the best score
      bestScore < yearScore
        ? ((bestScore = yearScore), (bestYear = year))
        : void 0
    })

  return `The best year was ${bestYear} with an average score of ${bestScore}`
```

another solution you make a great job solving this on your way :rocket:

ta-web-mexpt2 (commented):
great job in general just take a time to read my suggestions and keep in that way

PR #3404, opened 2023-05-15
[FTRMT052023] Pierre Docquin

Dokryptos:


src/movies.js
```diff
@@ -1,22 +1,103 @@
...
+}
+    
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    const copyMovies = [...moviesArray];
+    const twentyFirstMovie =[];
+ 
+    copyMovies.sort((a, b) => (a.title > b.title ? 1 : -1))
+    copyMovies.length = 20
```
TA-Remote, line 94: Here another option would be to `.slice(0, 20)` the result so you return the needed number of movies

src/movies.js
```diff
@@ -1,22 +1,103 @@
...
+    return allDirectors;
+}
+
+//  Bonus 1
+
+function clearAllDirector(arr){
+    let arrClearDirector = [];
+
+    arr.forEach(function(value){
+        if(!arrClearDirector.includes(value)){
```
TA-Remote, line 18: After if you have time also check `Set` for making unique arrays, if you have the free time and want to look on that one

TA-Remote (commented):
- Good work, keep it up :)

PR #3405, opened 2023-05-15
[FTRMT052023] Carolin Klose

Chiara888:
Initial link created

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,19 +1,106 @@
 // Iteration 1: All directors? - Get the array of all directors.
+
+function getAllDirectors(moviesArray) {
+
+const findDirector = moviesArray.map(function (name) {
+    return name.director
+})
+
+return findDirector
+}
+
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+
+// function getAllDirectors(moviesArray) {
+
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey Carolin!
Well done with your iterations, and using the array methods correctly🙌
- As we share in class this morning a possible solution for iter 4 (where you commented you couldn't find a way to divide just by dramaMovies), we need to moviesArray.filter() and get only the 'drama' movies. As you produce a perfect scoresAverage function🤩, you can also reused passing the filteredDramaArray as argument.
- About the Bonus 1 solution you asked me (sorry I realised later there was no solution🤦‍♀️) if it can help you, I'd share how I solved:
```
//Bonus 
function getUniquifyDirectors(moviesArray) {
  const directors = getAllDirectors(moviesArray);  --> reuse the function to get the array just with all directors

// POSSIBLE SOLUTTION USING .filter()
// --> accept 3 arguments: 'director' = placeholder for the director name, index = index of the "director element" , 
//  array = a copy of the 'directors' array itself
const uniqueDirectors = directors.filter((director, index, array) => { 
    return array.indexOf(director) === index; //--> using the indexOf to clean repetitions 
   });

// SOLUTION USING .forEach()
  const uniqueDirectors = []; // initialise empty array
  directors.forEach((director) => {
    if (!uniqueDirectors.includes(director)) { // if 'director' is not already inside --> add it 
      uniqueDirectors.push(director);
    }
  });
  return uniqueDirectors;  
}
```
Keep it up💪
Alessandra

PR #3406, opened 2023-05-15
 [FTRMT052023] Philippe Miossec

pmiossec:


src/movies.js
```diff
@@ -1,25 +1,120 @@
...
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    let count = 0;
+    return moviesArray
+        .map(m => m)
+        .sort((m1, m2) => m1.title.localeCompare(m2.title))
+        .map(m => m.title)
+        .filter(m => {
```
TA-Remote, line 59: Here it would be a little easier to use the `.slice(0, 20)` so you don't need to create a variable too :)

TA-Remote (commented):
- Impressive solution, congratulations

PR #3409, opened 2023-05-15
[FTRMT052023] Filipe Ferreira 

filipemgf:
Iteration 1-6 completed

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -56,10 +56,35 @@ function dramaMoviesScore(moviesArray) {
 }
 
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
+	let moviesArrayCopy = [...moviesArray];
+	let sortedByYear = moviesArrayCopy.sort((a, b) => {
+		if (a.year !== b.year) {
+			return a.year - b.year;
+		} else {
+			if (a.title > b.title) return 1;
+			if (a.title < b.title) return -1;
+		}
+	});
+
+	return sortedByYear;
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+
+function orderAlphabetically(moviesArray) {
+	let moviesArrayCopy = [...moviesArray];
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,25 +1,119 @@
...
+function scoresAverage(moviesArray) {
+	if (moviesArray.length === 0) {
+		return 0;
+	} else {
+		let scoreSum =
+			moviesArray.reduce((acc, curr) => {
+				if (typeof curr.score !== "number") {
+					curr.score = 0;
+					return acc + curr.score;
+				} else {
```
TA-Remote, line 32: You could remove the `else` statement because you're doing the same thing inside the `if `and the `else `with the `return`, so you could only change the score inside the `if `and outside the `if `you call the `return`
```
	moviesArray.reduce((acc, curr) => {
				if (typeof curr.score !== 'number') {
					curr.score = 0
				}
				return acc + curr.score
			}, 0) / moviesArray.length
```

src/movies.js
```diff
@@ -1,25 +1,119 @@
...
+function dramaMoviesScore(moviesArray) {
+	let dramasArray = moviesArray.filter((movieObj) => {
+		return movieObj.genre.includes("Drama");
+	});
+
+	if (dramasArray.length === 0) {
+		return 0;
+	} else {
+		let scoreSum =
+			dramasArray.reduce((acc, curr) => {
```
TA-Remote, line 50: Reuse your functions :) Maybe not in the first go, but in a refactor try to reuse then

src/movies.js
```diff
@@ -1,25 +1,119 @@
...
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
+	let moviesArrayCopy = [...moviesArray];
+	let sortedByYear = moviesArrayCopy.sort((a, b) => {
+		if (a.year !== b.year) {
+			return a.year - b.year;
+		} else {
+			if (a.title > b.title) return 1;
+			if (a.title < b.title) return -1;
```
TA-Remote, line 66: Use the power of built-in methods when you know/remember then, also can be applied in a refactor

TA-Remote (commented):
- Good work
- Good quality on your code, congratulations

PR #3412, opened 2023-05-15
[FTRMT052023] JuanLozano

Jualolo:
I couldnt sort alphabetically inside the yearly sorted :( but the rest I finished everything :)

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,78 @@
+
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Juan!
Awesome job in your lab! I'm sure if you had enough time you'd even rock the bonuses.

Few improvements that I could point out:
For itearation 2, where you should find all drama movies directed by Steven Spielberg, you could have only one filter with a if statement that looks like this:
`if(movie.director === 'Steven Spielberg && movie.genre.includes('Drama')`
A little shorter so it looks cleaner and more readable.

As for iteration 6, the best way to order alphabetically is to use[ localeCompare()](https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare). That's because sort() is not case sensitive as it's based on the [ASCII Table](https://www.asciitable.com/). 

Other than that, keep up the great work!

PR #3413, opened 2023-05-15
[FTRMT052023] cédricRocher

hymced:
all done with bonus
struggled with debugging...
1) why vs code does not alert when using a single = instead of == in an if condition?
2) and also I was calling the functions multiple times with the same input to make some checks, but even if the array is a (shallow) copy, the objects in it are not! so the first call to function was mutating the content of the array!

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,25 +1,2171 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  return moviesArray.map(movie => movie.director)
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey Cédric!
I'm always amazed by your comments, the way you describe the logic, the methods you used and what you want to achieve, and that you’re going further and try different solutions...  just my personal opinion.
Some notes based on your comments:
* Bonus 1: the instructions weren’t so clear, and it was really a plus.  you could build another function, reusing the getAllDirectors(), to clean up the repetitions. I know… jasmine doesn’t check this one, so console.log is our only way to test😜
*  = , == , === : as we choose how to compare, especially in if statement, I guess VS code can’t really tell us what to do, as someone suggested, there are some extensions to autocomplete/give suggestions. I personally so some student work with them and I found them annoying and not so precise. 
* When you want to compare values keep in mind the differences between the different [equality comparisons](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness)  the === is always the safest, and I suggest you to used it.
* I hope I could go through all your code…  
Anyway, great job🤩
Alessandra

hymced:
Thank you Alessandra :) and I will try to keep in mind to use "===", you are right, it's safer as Luis told us ;) More chances to catch something unexpected with it!

PR #3414, opened 2023-05-15
RM WD may23 Emilio Moreno

E-m-i-l-i-o:
I tried and advanced with the wonderful help of Vale. Later, I helped out other colleagues who were stuck with the same  issues Ale helped me with.  I continued further and now I feel exhausted so have to leave it here.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,13 +1,109 @@
+const data= [  {
+    title: 'The Shawshank Redemption',
+    year: 1994,
+    director: 'Frank Darabont',
+    duration: '2h 22min',
+    genre: ['Crime', 'Drama'],
+    score: 9.3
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey Emilio!
Great that you were able to go further and you even helped your classmates🙌!!!
Even for the tricky task, keep going with the same approach you used working on iteration 3:
* Recognise what the function has to return 
* Type of data you are working with
* Value we need , and where it stored
* Break it in step
* Console.log(a lot😉) :  to see your progress, identify the data you are returning before going further with another task, or whenever you need

That’s the right way to go!!
And if you have any other questions about those methods, please let us know.
 Keep it up💪
 Alessandra

PR #3415, opened 2023-05-15
[FTRMT052023] janiOjala

jani500:
Very challenging lab and not enough time to complete all iterations.
Need to revisit this, especially to practice sort & reverse, which I didn't have time for.


ykztenoks:
use length for iteration 2
Hi Jani!
Great work on your lab, I'm sure if you had enough time you'd nail all iterations.
Only thing I could point out for improvement is, instead of having a variable `count` for the second iteration, you just return the `dramaMoviesSpielberg.length`.
Other than that, keep it up!

PR #3416, opened 2023-05-15
[FTRMT052023] Lina Zerkly

Lina-z:
Sorry, tried to solve but everything is confused I don't understand nothing is working. Jasmine is also very confusing.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,19 +1,62 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+
+function getAllDirectors(moviesArray) {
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Lina!
It's a very tricky lab and there's small details to pay attention to, but don't worry, you did awesome!

For iteration 3 you might have an error because you needed to round it to 2 decimals using the [toFixed()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed) method like such:
`return Number(average.toFixed(2))`

the [Number()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number) is because the toFixed() will turn the value into a string. You could also use `+average.toFixed(2)` as the + would have the same result.

As for iteration 5, you just forgot to access the property year from the current element, like this:
`moviesArr.sort((currentElement, prevElement) => {currentElement.year - prevElement.year}`

If you have any questions, don't hesitate to ask.

Keep it up! ✨

PR #3417, opened 2023-05-15
Solved Lab

Anpal1986:
At the iteration 4 I need some help please.

What I would do: with the function filter, I would filter the elements that have as genre:"Drama". then I would take: from the new array with the drama movies: .score (But I cant! Is always giving me as undefined :s)


TA-Remote (commented):
- Hopefully with the code along things in iteration 4 were a bit clearer :) 
- I actually had a hard time to understand what is working and what maybe isn't working:
  - When we have automated tests is a good practice to use them and pass them, and it seems you just ignored it.
  - Some of the indentation of your code still doesn't help to read it 
- You should have created the functionality inside the functions already provided

PR #3419, opened 2023-05-15
[FTRMT052023] Mary Sue Meissner

MarySueMe:


src/movies.js
```diff
@@ -1,25 +1,128 @@
...
+//       return true;
+//     } else {
+//       return false;
+//     }
+//   });
+
+//   return dramaBySpielberg.length;
+// };
+
+function howManyMovies(moviesArray) {
```
TA-Remote, line 56: For that one, because you're using a count, probably it would be easier to use the `.forEach()` method.

Another approach would be to `return movieObj.director === 'Steven Spielberg' && movieObj.genre.includes('Drama')` inside the filter, so that way it would include in the final array all the movies that have the Steven Spielberg as Director and is a Drama movie.

and because we need a number as the return we can return the length of that filter for or function `return dramaBySpielberg.length`

Putting everything together:
```
	})

}
```

TA-Remote (commented):
- Keep working, check the solutions and see if you can get more iterations done / understood

PR #3421, opened 2023-05-15
movies oLA bONATI

olabonati:
not so great,
no idea what I am doing wrong thta even the basic entries dont lit up green in Jasmine
got stuck on assignment 3 :(

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,59 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey Ola!
 Good job identifying the right array methods to use in each iteration, and  retrieving the needed data 🙌
  The main reasons for your Jasmin to be completely red was that there were few syntax errors in your code. 
- When you VS is red —> find where is complaining, most of the time are parenthesis that are not matching, or wrong punctuation. VS code underline them in red.
* Tips to avoid it : keep your code indented (this is a good practice anyway😉) so that’s more readable , open and close parenthesis, then start typing you code inside, let VS code help you.

Some general suggestions on how to approach the iterations:
* Recognise what the function has to RETURN
* Type of data you are working with
* Value we need , and where it stored
* Break it in steps
* Console.log :  to see your progress, identify the data you are having before going further with another task, or whenever you need😉

if you have any other questions, please let us know🙂
keep it up with your hard work!!
Alessandra

PR #3422, opened 2023-05-15
[FTRMT052023] Lucas Gavião

Lucas-Gaviao:


src/movies.js
```diff
@@ -1,16 +1,63 @@
...
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+    const onlyDirectors = moviesArray.map((element) => element.director);
+    return onlyDirectors;
+};
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
```
TA-Remote, line 11: For that one, because you're using a count, probably it would be easier to use the .forEach() method.

Another approach would be to return movieObj.director === 'Steven Spielberg' && movieObj.genre.includes('Drama') inside the filter, so that way it would include in the final array all the movies that have the Steven Spielberg as Director and is a Drama movie.

and because we need a number as the return we can return the length of that filter for or function return dramaBySpielberg.length

Putting everything together:
```
	})

}
```

TA-Remote (commented):
- Good work
- Keep the practice

PR #3425, opened 2023-05-15
[FTRMT052023] Nuno 

nunojsmonteiro:


src/movies.js
```diff
@@ -1,16 +1,66 @@
...
+    let dramaScore =
+      dramaMovies.reduce((acc, cValue) => {
+        if (typeof cValue.score !== "number") {
+          cValue.score = 0;
+          return acc + cValue.score;
+        } else {
+          return acc + cValue.score;
+        }
+      }, 0) / dramaMovies.length;
+    return Number(dramaScore.toFixed(2));
```
TA-Remote, line 61: In that one you could have reused the previous function to calculate the average

TA-Remote (commented):
- Good work, if you have some spare / free time try to go for a few more iterations and bonus because those can be really good practice for the general things we need to do with data

PR #3426, opened 2023-05-15
[FTRMT052023] eirikKlebo

eiriktobias:


TA-Remote:
Hey Eirik, 
Quick note on iter 3 and the small part wasn't passing the test:
It happens very often that in our functions we are checking if the data we have to exist before we start to manipulate/ work with. It’s a way to prevent our code to break, avoid errors, and our app to crash. In this lab is not a big deal, but a simple if statement can save us😜
It happens very often that in our functions we are checking if the data we have to exist before we start to manipulate/ work with. It’s a way to prevent our code to break, avoid errors, and our app to crash. In this lab is not a big deal, but a simple if statement can save us😜
It happens very often that in our functions we are checking if the data we have to exist before we start to manipulate/ work with. It’s a way to prevent our code to break, avoid errors, and our app to crash. In this lab is not a big deal, but a simple if statement can save us😜
```
    let sum = 0;
    let count = 0;
  
    moviesArray.forEach((movie) => {
    if(movie.score){    //  only if it exist add it to sum, prevents to make the sum value to become NaN or Undefine
               sum += movie.score;
    }
      count++;
    });
  
    const averageScore = count > 0 ? sum / count : 0;
    return Number(averageScore.toFixed(2));
  }
```
... and if you have time maybe you can try to salve it using .reduce()
keep it up 💪
Alessandra

PR #3427, opened 2023-05-15
[FTRMT052023] Fernando

fndoRamos:


TA-Remote:
Hey Fernando, 
Very well done🙌

Few tips if you wanna clean up a bit your code😉:

```
// your solution…

    let reducedScores = [];

        return 0;
    } else {
        reducedScores = moviesArray.reduce(function(acumulator, currentValue){
            if (currentValue.score === undefined){
                return acumulator;
            } else {
                return acumulator + currentValue.score;
            }
            
        }, 0);
    }

    return +((reducedScores / moviesArray.length).toFixed(2));
}

// your solution a bit cleaner
    return 0; // don’t need ‘else’, if it doesn’t pass this condition the code won’t run further
  }
// CONVENTION: store the method result in a variable using const
  const reducedScores = moviesArray.reduce((acumulator, currentValue) => {
    if(currentValue.score === undefined){ // other way should check if falsy (!currentValue.score)
        return acumulator
    }
    return acumulator + currentValue.score;
  }, 0);
  return +((reducedScores / moviesArray.length).toFixed(2));
}
```

Take it as suggestions on good practices and conventions to follow, when we are getting more familiar.  
keep it up with your great work💪
Alessandra

PR #3429, opened 2023-05-15
[FTRMT052023] Viktor N.

ViktorNstr:
Solved Lab.


TA-Remote (commented):
- Very good solution, I don't have any to say aside from good work and try to do the bonus if you have the free time

PR #3430, opened 2023-05-16
[FTRMT052023] Diogo Martins

disaacjm:


TA-Remote:
Hey Diogo!
small feedback about the parts of the test you couldn't pass:

- iter 4 (drama movies average) 2 ways to go: 
* add if statement to check if dramaMovies is not empty, before proceeding with the reduce method and find the average.
* Reuse scoresAverage() function, and passing dramaMovies as a parameter

- iter 5 (sort by year): you perfectly use the sort method, the only thing is that we need to compare a and b , year property (as a and b are elements of the moviesArray —> in this case object. 

keep it up with you great job!!💪
Alessandra

PR #3431, opened 2023-05-16
[FTRMT052023] Gokce Berndt

Gokceber:
I could only solve 4 of them and I couldn't manage to turn everything green on Jasmine but my answers seemed to work on codepen🤷🏻‍♀️

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,16 +1,79 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+    return directorsArray
+}
+
+const directorsArray = movies.map ((element) => {
+    return element.director;
+});
+
+getAllDirectors(movies) 
+
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+  let count = 0;
+
+const moviesSpielberg = moviesArray.filter ((element) => {
+  if (element.director === 'Steven Spielberg') {
+    count ++;
+    return true;
+    }
+});
+    
+}
+
+howManyMovies(movies)
+
+
+
+    
+
+
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  const scores = moviesArray.reduce(function(accumulator, currentValue) {
+    return accumulator + currentValue.score;
+  },0);
+
+  return scores;
+}
+
+const averageResult = scoresAverage(movies) / movies.length;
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,16 +1,79 @@
...
+    return element.director;
+});
+
+getAllDirectors(movies) 
+
+
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
```
TA-Remote, line 17: For that one, because you're using a count, probably it would be easier to use the `.forEach()` method.

Another approach would be to `return movieObj.director === 'Steven Spielberg' && movieObj.genre.includes('Drama')` inside the filter, so that way it would include in the final array all the movies that have the Steven Spielberg as Director and is a Drama movie.

and because we need a number as the return we can return the length of that filter for or function `return dramaBySpielberg.length`

Putting everything together:
```
	})

}
```

src/movies.js
```diff
@@ -1,16 +1,79 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+    return directorsArray
+}
+
+const directorsArray = movies.map ((element) => {
```
TA-Remote, line 8: You needed to put that piece of code inside your functions and change which array you're targeting
```suggestion
const directorsArray = moviesArray.map ((element) => {
```

src/movies.js
```diff
@@ -1,16 +1,79 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
```
TA-Remote, line 4:
```
  const directorsArray = moviesArray.map ((element) => {
    return element.director;
  });
  return directorsArray
}
```

src/movies.js
```diff
@@ -1,16 +1,79 @@
...
+
+
+
+    
+
+
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
```
TA-Remote, line 38: Here it would be something similar to the first one, you probably should have written your code inside the function and you would need do some checking, like for and empty array and if the object has the `currentValue.score`

A possible solution:

```
		return 0
	}

	const scores = moviesArray.reduce(function (accumulator, currentValue) {
		if (typeof currentValue.score !== 'number') {
			currentValue.score = 0
		}

		return accumulator + currentValue.score
	}, 0)

	const averageResult = scores / moviesArray.length
	const decimals = +averageResult.toFixed(2)
	console.log(decimals)
	return decimals
}
```

TA-Remote (commented):
- Good work, keep up the practice
- Try to check the suggestions, solutions and go over again the code along and see if that helps to understand everything better :)

PR #3432, opened 2023-05-16
[FTRMT052023] Maria Friedemann

softcake1988:
Really hard, could barely solve any problems. Had to google a lot for what I was able to solve.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,19 +1,90 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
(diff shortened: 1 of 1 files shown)


TA-Remote:
Hey Maria!
good job with your iterations🙌
Google is the best tool to use...
just a quick note on iter 4: 
when you need to retrive only the drama movie is rigth to check for the property genre inside the 'movie' object,
taking a look at the dat structure, you can see that genre is an array. So, as you did in iter 2, is better to use indexOf() or, even better, .includes('Drama')
```
 {
    title: 'The Godfather: Part II',
    year: 1974,
    director: 'Francis Ford Coppola',
    duration: '3h 22min',
    genre: ['Crime', 'Drama'],
    score: 9
  },
```
Please, if you have any other questions let us know😉
And keep it up💪
Alessandra

PR #3433, opened 2023-05-16
Iteration 1-2, started 3

maandysi:
I felt like I had to go through the lessons again because I struggled with the lab. So read the lessons and did little exercises instead of being frustrated with the lab.

### DIFF
--- a/src/data.js
+++ b/src/data.js
@@ -2001,3 +2001,4 @@ const movies = [
     score: 8
   }
 ];
+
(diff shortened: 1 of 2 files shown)


TA-Remote (commented):
- Good work and code on the iteration that you did
- I would advise you to try go over that lab because those array methods are really useful in JS and to manage the data that we got, usually the collection of data comes in an array of objects for example :)

PR #3434, opened 2023-05-16
[FTRMT052023] joaoFaria

jfaria23:
Still struggling with functions and with the introduction of the arrow functions, just made it all worse.
Most of these was research, it's taking me a lot of time to write code by memory and I keep making syntax errors.
Let's keep the good fight


ykztenoks:
Hey João!
Good work on your iterations, I'm sure if you had enough time you'd rock more iterations.

Don't struggle trying to write code by memory, as you're just starting there's no problem at looking at the teachers code, documentation or even articles as reference. Don't stress it 😉 

Keep it up!

PR #3436, opened 2023-05-16
lab done

JCouderc87:
Only managed to do the first. Could not figure out how to filter Spielberg and drama for the second. My brain was absolutely fried after the day of lectures yesterday.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,10 +1,36 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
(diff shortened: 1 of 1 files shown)


ykztenoks:
Hey Jeremy!
For the `howManyMovies` function/iteration you'd need a couple of conditions. You'd have to check if in the `genre` array there's Drama included for the current movie AND also if the current movie director is Steven Spielberg. You can achieve that with:

```
moviesArray.filter((current) => {
  if(current.director === 'Steven Spielberg' && current.genre.includes('Drama')
} 
```

That way you can filter the movies that are directed by Steven Spielberg and have Drama in the genre array.

You can learn more about the [includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes). It's a very powerful method to check for elements in an array.

Keep it up!

PR #3437, opened 2023-05-16
Nishtha

nishag-26:


src/movies.js
```diff
@@ -1,13 +1,34 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(movies) {
+    const myArr = movies.filter(function (director) {
+      });
+      return myArr
+      console.log(myArr);
```
TA-Remote, line 8: The piece of code below this line will never be executed, because we have a return statement on line 7

src/movies.js
```diff
@@ -1,13 +1,34 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(movies) {
+    const myArr = movies.filter(function (director) {
```
TA-Remote, line 5: Possibly because I want to get the directors from the movies array, that would be a transformation, so you possibly would use a `.map()` for that one.

One possible piece of code to make it work would be:

```
  const justDirectors = movies.map(function(movie){
    return movie.director
  })
  
  return justDirectors
```

If you paste the previous piece of code inside the `getAllDirectors` function, now I'm creating a new array with only the directors from each movie

src/movies.js
```diff
@@ -1,13 +1,34 @@
...
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+    const arr = movies.filter(function(director,genre) {
+     
+        return director.name == "Steven Spielberg"
+        return genre.name == "Drama"
+        return arr
```
TA-Remote, line 22: Remember that a function usually can have just one `return` the exception would be if you have different triggers for thar return to be reached

```
	function biggerThanOne(number) {
		if (number > 1) {
			return 'The number is bigger than one'
		} else {
			return 'The number is smaller than one'
		}
	}
```

In the previous case you only have one return being activate anytime you call the function 
biggerThanOne(2) => return 'The number is bigger than one'
biggerThanOne(0) => return 'The number is smaller than one'

TA-Remote (commented):
- Good work, keep it up and keep practicing
- Try in your free time to go over the solution and understand what could have been done in that code

PR #3438, opened 2023-05-16
 [FTRMT052023]  Fabio

FabioSM46:


TA-Remote:
Hey Fabio, 
Lab very well done🙌 Just a note on the iteration 6:

```
  const copy = [...moviesArray] // make a copy, .sort() modify the original, you have also other way tomato a copy…
    const sorted = copy.sort(function (a, b) {
       if(a.year === b.year){ // this if will check green only the last test
            return a.title.localeCompare(b.title) 
          }
      return a.year - b.year; // will order all the movie objects by year
    });
    return sorted;
  }
```


Keep it up with your great solutions an clean code💪
Alessandra

PR #3477, opened 2023-07-13
[LIS-FT-JULY23] Agnes

AgnesMorawska:


src/movies.js
```diff
@@ -1,22 +1,74 @@
...
+  });
+}
+
 
+//GO THROUGH THIS ONE
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+function orderAlphabetically(moviesArray) {
+    let titleArr = moviesArray.slice(0)
+    let  ordTitleArr = titleArr.map(el => {
```
brunorocha20, line 66: Agnes, good way of approaching this exercise! The only thing I want to review is that in this specific case, you didn't need to do that first step of cloning the `moviesArray` since you are going to return a new array when you apply the `.slice(0,20)` later on to return the first 20 titles.

brunorocha20 (commented):
Great work on the lab Agnes.

PR #3484, opened 2023-07-13
[LISFTJULY23] FILIPA FLORA LAB4

filipaflora:


src/movies.js
```diff
@@ -1,25 +1,97 @@
...
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+
+function dramaMoviesScore(moviesArray) {
+
+    let dramaMovies = moviesArray.filter((films) =>{
+        if(films.genre.includes('Drama')){
+            return films;
+        }
+        else if (!films.genre.includes('Drama')){
```
brunorocha20, line 47: Filipa, this step here is not that necessary since you are doing the `.filter()` method to check which `films` genre includes 'Drama' and storing that values in your `dramaMovies` variable. That `if` condition acts the way you want so there's no need to double-check if the movies don't include 'Drama' and return 0 if so.

src/movies.js
```diff
@@ -1,25 +1,97 @@
...
+    });
+    return newArray
+}
 
 // Iteration 6: Alphabetic Order - Order by title and print the first 20 titles
-function orderAlphabetically(moviesArray) {}
+
+
+function orderAlphabetically(moviesArray) {
+    let newArray = moviesArray.slice(0)
```
brunorocha20, line 71: Filipa, good way of approaching this exercise! The only thing I want to review is that in this specific case, you didn't need to do that first step of cloning the `moviesArray` since you are going to return a new array when you apply the `.slice(0,20)` later on to return the first 20 titles.

brunorocha20 (commented):
Great job on the lab Filipa.

PR #3514, opened 2023-08-07
[FTRMT072023] Johanna

johges:
I was able to solve the lab until iteration 4. Unfortunately, I could not solve in iteration 3 the task "should return average even if one of the movies does not have score".  I'm looking forward to seeing how others solved the problem and the iterations coming next.


TA-Remote:
hi @johges, 
hope this morning you've got the answer about the only small part of the test wasn't passing, and for any other questions... you know where to find us😉
Ale

PR #3515, opened 2023-08-07
[FTRMT072023] Alex

alexahijado:


TA-Remote:
hi @alexahijado, 

I’d just give you few hints on how to break the tasks of a function in little steps:  
* When using arrays methods as .map, .filter , etc. store them in a variable and name it trying to describe what it contains. 
` const directorsArr = moviesArray.map(....)`
       In this way you can always console.log your variable inside your function and check if the result is what you were expecting before going forward with the next task. And have a more readable code if you still need to work on it.
   `const numberOfDirectors = directorsArr.length`
* Getting properties value in array of objects: 
- Take a look at the data structure of the array you’re iterating
```
 const dramaMoviesArr = moviesArray.filter(function(movie){
      console.log(movie) --> print all the movie objects one by one
      console.log(movie.genre) --> print only the genre property of all movies
    })
```
- Recognise the value data type stored in a property and how to access it:
```
 {
    title: 'Dead Poets Society',
    year: 1989,
    director: 'Peter Weir',
    duration: '2h 8min',
    genre: ['Comedy', 'Drama'], --> movie.genre is an array of strings
    score: 8
  }
```
- To know if an array contains a specific value we can use 2 methods: [.includes()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) and [.indexOf()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf):
   ```
   movie.genre.includes('Drama') --> return true or false
   movie.genre.indexOf('Drama') --> return 1 if Drama is in or -1 if it doesn't
```
Console.logs are the best way to check what we are doing, test if the code is giving the result we want, and are free😜!!!

Please let us know if you have any question, and don't hesitate to ask whenever you get stuck!!!

Keep it up💪
Ale

PR #3574, opened 2023-10-16
[LIS FT OCT 23 ] André Gomes

Andre8285:


webLisbonTA:
Hi André,
Good job on your lab!
On Interation 3, you didn't need to use map, only reduce. Please see the code below:
```
    let mScore = moviesArray.reduce((acc, val) => {
      if (!val.score) return acc;
  
      return acc + val.score;
    }, 0);
    let mScoreAvg = mScore / moviesArray.length;
    return Math.round(mScoreAvg * 100) / 100;
  }
```
Overall, good job!

PR #3580, opened 2023-10-16
[LIS-FT-OCT23] João_Cravidão

joaocravidao:


jacobviv (commented):
Great job with your lab João!

src/movies.js
```diff
@@ -1,25 +1,155 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+        
+        const newMoviesArray = [...moviesArray];
```
jacobviv, line 6: You don't need to make a copy when using map, filter or reduce

PR #3600, opened 2023-10-16
solving BONUS8 felt like catching mew for the first time!

VascoGaitinha:


jacobviv (commented):
Great work Vasco!

src/movies.js
```diff
@@ -1,25 +1,165 @@
...
+    })
+    return newArray}; 
+
 
 // BONUS - Iteration 8: Best yearly score average - Best yearly score average
-function bestYearAvg(moviesArray) {}
+function bestYearAvg(moviesArray){
+    if (moviesArray.length === 0){return null}
+    // Step 1 -Sort by Year
+        arrayToWork = [...moviesArray]
```
jacobviv, line 117: Another way, a shorter one
:)

```javascript
function bestYearAvg(moviesArray) {
        return null
    }
    const moviesYears = {};

    moviesArray.forEach(function (movie) {
        if (!moviesYears[movie.year]) {
            moviesYears[movie.year] = [];
            moviesYears[movie.year].push(movie);
        } else {
            moviesYears[movie.year].push(movie);
        }
    });

    let highest = 0;
    let bestYear;

    for (const year in moviesYears) {
        const currentYearAverage = scoresAverage(moviesYears[year]);

        if (currentYearAverage > highest) {
            highest = currentYearAverage;
            bestYear = year;
        } else if (currentYearAverage === highest) {
            // Check which year is the lowest and save it to the `oldestYear`
            const oldestYear = year < bestYear ? year : bestYear;
            bestYear = oldestYear;
        }
    }

    return `The best year was ${bestYear} with an average score of ${highest}`;
}

```

PR #3630, opened 2023-10-23
 [FTWDOCT23] Ana Ribeiro

AnaRibeiroPT:


src/movies.js
```diff
@@ -1,25 +1,85 @@
...
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  if (moviesArray.length === 0) {
+    return 0;
+  }
+  const calcAvg = moviesArray.reduce(function (acc, movies) {
+    return acc + movies.score;
```
iPoolito, line 25: In order to check if the movie has a score we can create an if statement that check if the value score exist inside the movie, if it exist add it to acc if not only return the acc 
```suggestion
        if (movies.score) {
            return acc + movies.score;
        }
        return acc;
```

src/movies.js
```diff
@@ -1,25 +1,85 @@
...
+    }
+    if (movie1.year > movie2.year) {
+      return 1;
+    }
+    if (movie1.title < movie2.title) {
+      return -1;
+    }
+    if (movie1.title > movie2.title) {
+      return 1;
+    }
```
iPoolito, line 59: Another way to do it is using localeCompare when you want to compare 2 strings alphabetically in this case the title of the movie only if the year of both iare equal 
```suggestion
      if (movie1.year === movie2.year) {
            return movie1.title.localeCompare(movie2.title);
        }
        return movie1.year - movie2.year;
```

PR #3643, opened 2023-10-23
[RMT-WDFT-OCT23] Vasco Martins

LM-Vasco:
On exercise 6, jasmine is giving me the following errors:

- should only return the title of the movies, each value should be a string;
- should order them alphabetically.

I tested the code in codepen and it seems to work just fine

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,25 +1,103 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  const directors = moviesArray.map((movie) => movie.director);
+
+  return directors;
+}
+
+//Bonus: Iteration 1.1
+
+function cleanAllDirectorsArray(moviesArray) {
+  const directors = getAllDirectors(moviesArray);
+
+  let count = 0;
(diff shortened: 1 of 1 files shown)


LM-Vasco:
Also, I mentioned to Luís that I would be doing this lab with Camila, but still individually. Each did it's own and then we compared to see how each one did it.

TA-Remote:
hey @LM-Vasco 
great job with your solution🙌

```
  const moviesArrayCopy = JSON.parse(JSON.stringify(moviesArray));

  // orderByTitle --> it's an array of movies object sorted alphabetically
  const orderByTitle = moviesArrayCopy.sort((a, b) => {
    if (a.title < b.title) return -1;
    if (a.title > b.title) return 1;
    return 0;
  });

  // Here you are saying: if there are less then 20 elements in moviesArray,
  // return orderByTitle --> taht is an arr of movies OBJ
  // but the test it's expecting an array of strings (only with titles)

  if (moviesArray.length < 20) {
      console.log(orderByTitle)
      return orderByTitle;
  }

  // you can just remove the 👆 if statement,
  // as you still want to map and get the array with all titles first
  // and then check the legth

  const titlesOnly = orderByTitle.map((titles) => titles.title);

  if (moviesArray.length > 20) {
    titlesOnly.splice(20);
  }
  return titlesOnly;

  // alternative without if statement using .slice()
  // will keep only the first 20 elements
  // but it'll also keep all elements if less then 20
  // return titlesOnly.slice(0, 20)
}
```
Hope it makes sense 😉, if not please let me know!

Ale

PR #3668, opened 2023-12-05
[WD-PT] Eduardo Botelho

UPSxACE:
The first bonus could not be done in the same function as the first iteration, because it would make it fail the tests, so I created a new "getAllDirectorsBonus" function!

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,25 +1,168 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  const directors = moviesArray.map((movie) => movie.director);
+  return directors;
+}
+
+// BONUS ITERATION (with the filter for duplicates):
+function getAllDirectorsBonus(moviesArray) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,25 +1,168 @@
...
+// BONUS ITERATION (with the filter for duplicates):
+function getAllDirectorsBonus(moviesArray) {
+  const directors = moviesArray.reduce((directors, movie) => {
+    if (!directors.includes(movie.director)) {
+      directors.push(movie.director);
+    }
+    return directors;
+  }, []);
+  return directors;
+}
```
iPoolito, line 18: Also another way to do it its with this method 

```suggestion
// BONUS ITERATION (with the filter for duplicates):
function getAllDirectorsBonus(moviesArray) {
 return [...new Set(moviesArray)]
}
```

src/movies.js
```diff
@@ -1,25 +1,168 @@
...
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
-function dramaMoviesScore(moviesArray) {}
+function dramaMoviesScore(moviesArray) {
+  const dramaMovies = moviesArray.filter((movie) =>
+    movie.genre.includes("Drama")
+  );
+  const scoresSum = dramaMovies.reduce(
+    (accumulatedScore, movie) =>
+      movie.score ? (accumulatedScore += movie.score) : accumulatedScore,
```
iPoolito, line 54: 🔥 nice!!

iPoolito:
Well done with the code 🚀 a really good job congratz 🔥

PR #3672, opened 2023-12-08
Lab Greatest Movies - Irene

PurrfectNene:


src/movies.js
```diff
@@ -1,22 +1,72 @@
...
+  return moviesArray.map(function (element) {
+    return element.director;
+  });
+}
+
+const directors = getAllDirectors(movies);
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
```
iPoolito, line 14: First whe need to check the lenght of the movies array 
```suggestion
function howManyMovies(moviesArray) {
  if (!moviesArray.length) {
        return 0
    }
```

src/movies.js
```diff
@@ -1,22 +1,72 @@
...
+  } else {
+    return stephenMovies.length;
+  }
+}
+
+const dramaMovies = howManyMovies(movies);
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
+
 function scoresAverage(moviesArray) {}
```
iPoolito, line 32: Here probably you can use a reduce function and inside of the reduce make sure that the property array from the currentElement exists in order tu add it to the total, if not just return the total

src/movies.js
```diff
@@ -1,22 +1,72 @@
...
 
 // Iteration 4: Drama movies - Get the average of Drama Movies
 function dramaMoviesScore(moviesArray) {}
```
iPoolito, line 35: Here prob you can re use the function that is up, just first filter the array

PR #3674, opened 2023-12-08
lab done except bonuses

moreira8125:


iPoolito:
Nice job working with the Arrays 🚀

src/movies.js
```diff
@@ -1,22 +1,94 @@
...
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+   const newMovieArr = moviesArray.map(function(element){
+        return element.director
+    })
+
+//    const filteredArr = newMovieArr.filter(function(director, index, array) {
+//     return array.indexOf(director) === index;
+// })
+// i checked with AI but dont completely understand the parameters and how is automatically plugged what it needs
```
iPoolito, line 12: were you can do something like this, this cleans the array 

`  return [...new Set(newMovieArr)]`

PR #3675, opened 2023-12-08
lab done

MonicaAgrela:
i have some difficults with:
scoresAverage;
dramaMoviesScore.

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -1,22 +1,78 @@
 // Iteration 1: All directors? - Get the array of all directors.
 // _Bonus_: It seems some of the directors had directed multiple movies so they will pop up multiple times in the array of directors.
 // How could you "clean" a bit this array and make it unified (without duplicates)?
-function getAllDirectors(moviesArray) {}
+function getAllDirectors(moviesArray) {
+  return moviesArray.map((movie) => {});
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+  if (moviesArray.length === 0) {
+    return 0;
+  }
+  const spielbergDramas = moviesArray.filter(
+    (movie) => movie.director === "Steven Spielberg" && movie.genre === "Drama"
+  );
+  return spielbergDramas.length;
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  if (moviesArray.length === 0) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,22 +1,78 @@
...
+}
 
 // Iteration 2: Steven Spielberg. The best? - How many drama movies did STEVEN SPIELBERG direct?
-function howManyMovies(moviesArray) {}
+function howManyMovies(moviesArray) {
+  if (moviesArray.length === 0) {
+    return 0;
+  }
+  const spielbergDramas = moviesArray.filter(
+    (movie) => movie.director === "Steven Spielberg" && movie.genre === "Drama"
```
iPoolito, line 14: the property genre in the movies data is an Array of genres something like this 
`    genre: ['Crime', 'Drama'],` so what we need to use is the include methods for the array would be something like this 

```suggestion
    (movie) => movie.director === "Steven Spielberg" && movie.genre.includes('Drama')
```

src/movies.js
```diff
@@ -1,22 +1,78 @@
...
+}
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+  if (moviesArray.length === 0) {
+    return 0;
+  }
+
+  const validScores = moviesArray.filter((movie) => typeof movie === "number");
```
iPoolito, line 25: here you need to specify which property 

```suggestion
  const validScores = moviesArray.filter((movie) => typeof movie.score === "number");
```

src/movies.js
```diff
@@ -1,22 +1,78 @@
...
+  const average =
+    validScores.reduce((sum, movie) => sum + movie, 0) / validScores.length;
```
iPoolito, line 27: and here te property also that you want to add and the average its refering to all the elements of the array so should be something like this 

```suggestion
    validScores.reduce((sum, movie) => sum + movie.score, 0) /  moviesArray.length;
```

src/movies.js
```diff
@@ -1,22 +1,78 @@
...
+  const dramaMovies = movies.filter((movie) => movie.genre === "Drama");
+
+  if (dramaMovies.length === 0) {
+    return 0;
+  }
+
+  const totalRating = dramaMovies.reduce((sum, movie) => sum + movie.rating, 0);
+
+  const averageRating = totalRating / dramaMovies.length;
+  return averageRating;
```
iPoolito, line 47: So once we fix the function up we can use it here in order to get the average
```suggestion

  const dramaMovies = moviesArray.filter((movie) => movie.genre.includes("Drama"));

  const averageRating = scoresAverage(dramaMovies)
  return averageRating;
```

PR #3677, opened 2023-12-09
WEB DEV PT NOV Natalia Pinto

Natpinper:
Iterations 1 to 6 complete. 
Iteration 7: Trying to split a string into numbers and delete the letters to get hours and minutes as integer

### DIFF
--- a/src/movies.js
+++ b/src/movies.js
@@ -51,13 +51,40 @@ const dramaScore= (moviesArray.filter(function (element) {
 }
 console.log(dramaMoviesScore(movies))
 // Iteration 5: Ordering by year - Order by year, ascending (in growing order)
-function orderByYear(moviesArray) {}
+function orderByYear(moviesArray) {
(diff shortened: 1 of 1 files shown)


src/movies.js
```diff
@@ -1,25 +1,90 @@
...
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+    if(moviesArray.length===0){
+        return 0
+    }
+   
+    const scoresAv=(moviesArray.reduce(function(sum, currentValue){
+        return sum+currentValue.score
```
iPoolito, line 30: Here we need to check the type of score, sometimes they dont pass it so wee need to make sure its a number

```suggestion
       if(typeof currentValue.score==="number"){
return currentValue.score +sum
    }
      return sum
```

PR #3678, opened 2023-12-09
ELEANOR PITCHER WD PT 

eleanorpitcher:


src/movies.js
```diff
@@ -1,25 +1,96 @@
...
+}
+
+
+function cleanAllDirectors(moviesArray){
+    let newArr = [];
+    for (let i = 0; i<moviesArray.length; i++){
+        if (newArr !== moviesArray[i].director){
+            newArr.push(moviesArray[i])
+        }
+    }
```
iPoolito, line 17: Nice jooob alose There is another way to do it, like this 🚀 
```suggestion
 return [...new Set(moviesArray)]
    }
```

src/movies.js
```diff
@@ -1,25 +1,96 @@
...
+}
+
 
 // Iteration 3: All scores average - Get the average of all scores with 2 decimals
-function scoresAverage(moviesArray) {}
+function scoresAverage(moviesArray) {
+    if (moviesArray.length === 0){
+        return 0;
+    }
+    const totalScores = moviesArray.reduce((sum, element) => sum + element.score, 0);
```
iPoolito, line 36: Here in order to get the last iteration correct you need to create an if statement to check  if the score property its coming 

```suggestion
    const totalScores =moviesArray.reduce((sum, element) => {
        if (element.score) {
            return sum + element.score
        } else {
            return sum
        }
    }, 0)
```

iPoolito:
Nice jooob 🚀

PR #3956, opened 2025-06-19
Solved lab

FelixRodriguezG:


FelixRodriguezG:
In the first exercise, a Set structure was added to eliminate duplicates.

PR #3992, opened 2026-03-24
Lab JavaScript movies done

noahramoss:


Imran-imtiaz48 (commented):
The implementation covers most of the required functions and demonstrates a good understanding of common JavaScript array operations such as map, filter, reduce, and sort. The code is generally readable and the use of guard clauses, such as returning 0 when arrays are empty, helps prevent runtime errors. The structure of the solutions also shows a logical approach to solving each iteration of the lab. However, there are several issues that should be addressed to improve correctness and maintainability. Some sections of the file appear duplicated or partially copied, which makes the code harder to read and suggests that parts of the implementation may not have been cleaned before committing. In getAllDirectors, the function currently returns all directors including duplicates, while the exercise specifically asks for a unified list without duplicates. In the scoresAverage and drama average calculations, the logic works but could be simplified by reusing helper functions instead of repeating similar reduce logic. The bestYearAvg function also contains a logic problem: the average for each year is not actually calculated because the total score is not divided by the number of movies, and the line yearScores[year].length; has no effect. Additionally, the returned string does not format the average value to two decimal places as expected. Cleaning duplicate code, fixing the average calculation, and improving reuse of existing helper functions would significantly improve the overall quality and correctness of the solution.

